PR #20, opened 2026-02-28
Implement config system + state dir + CLI (Fixes #10)

### LINKED ISSUE #10: [Day 1][Samet] Config system + secrets source precedence + config init/print UX
Implement the full configuration subsystem and CLI setup flow.

Scope:
- Precedence: flags > env > `.dir2mcp.yaml` > defaults
- `secrets` + `security` config blocks
- `config init` interactive wizard (masked secrets, TTY aware)
- `config print` effective resolved config
- Non-interactive missing-config contract (exit code 2)

Acceptance criteria:
- Precedence behavior verified with tests
- Snapshot never stores plaintext secrets
- Missing required config yields actionable output

sametd04:
- Config: precedence flags > env > file > defaults, init/print, validation, snapshot (no plaintext secrets)
- State: .dir2mcp/ layout, secret.token, connection.json, lock file
- CLI: up, status, ask, reindex, config init/print, version; interactive wizard (TTY, masked input)
- Stubs: MCP server (501), retrieval, ingest
- Tests: precedence, snapshot redaction, missing-config actionable output
- Makefile, README build/run, Go 1.22 in CI

## Summary by CodeRabbit

* **New Features**
  * Added build and run support with updated setup instructions.
  * Introduced retrieval engine framework for enhanced functionality.

* **Documentation**
  * Updated build and run instructions in README.
  * Added detailed documentation for project configuration and CLI improvements.

* **Chores**
  * Updated Go language requirement to version 1.22.
  * Added required framework and system dependencies.

Closes #10


yidakra:
@copilot+gpt-5.3-codex 

Verify each finding against the current code and only fix it if needed.

Inline comments:
In `@internal/cli/ask_cmd.go`:
- Line 48: Replace the hardcoded retrieval depth K: 10 in the engine.Ask call
with the configured RAG retrieval depth; read the configured value (the project
config key rag.k_default / your config struct field, e.g., cfg.RAG.KDefault or
similar) and pass it into retrieval.AskOptions instead of 10 so
engine.Ask(question, retrieval.AskOptions{K: ...}) uses the configured
precedence; update the call site in ask_cmd.go where engine.Ask and
retrieval.AskOptions are used to source the value from the existing config
object.
- Line 28: The code currently ignores the error from filepath.Abs for stateDir
(stateDir, _ = filepath.Abs(stateDir)), which can let a bad path propagate;
change this to capture and check the error (e.g., stateDir, err :=
filepath.Abs(stateDir) and handle err) inside the function that uses stateDir
(the code around the filepath.Abs call in ask_cmd.go), and on error return it up
the call stack or log and exit with a clear message so execution does not
continue with an invalid stateDir.

In `@internal/cli/config_cmd.go`:
- Line 78: The call to filepath.Abs(stateDir) ignores its error; update the code
around stateDir, filepath.Abs(...) to capture (abs, err) :=
filepath.Abs(stateDir) and only assign stateDir = abs on err == nil, and on err
handle it the same way as the earlier error check (log the error and
exit/return) so the failing Abs call doesn't silently leave an invalid path;
reference variables/functions: stateDir and filepath.Abs.

In `@internal/cli/root.go`:
- Around line 60-65: The helper function setExitCode is a no-op and unused;
remove the setExitCode function and its comment, and replace any intended
references with the existing exitWith(...) routine (which prints to stderr and
calls os.Exit), or if you intend to keep it implement it properly; update any
documentation/comments that referenced setExitCode to mention exitWith instead.
Ensure no remaining calls or tests reference setExitCode before committing.

In `@internal/cli/status.go`:
- Line 27: The call to filepath.Abs in internal/cli/status.go currently discards
its error (stateDir, _ = filepath.Abs(stateDir)); update the code that calls
filepath.Abs to capture and handle the error (e.g., stateDir, err :=
filepath.Abs(stateDir)), then propagate or handle err appropriately in the
surrounding function (return the error, call processLogger/error logger and
exit, or surface a user-friendly error) so that path resolution failures for
stateDir are not silently ignored; reference the stateDir variable and the
filepath.Abs call when making the change.

In `@internal/config/template.go`:
- Around line 29-31: Template contains an unsupported key ingest.audio.cache
which is not defined in the config struct (see internal/config/config.go where
IngestMode only exposes Mode), so remove the unsupported "cache" entry from
internal/config/template.go's audio block; update the audio section to only
include supported fields (e.g., "mode: auto") and ensure no other template keys
refer to "cache" or other non-existent fields.

In `@internal/mcp/server.go`:
- Around line 26-27: NewServer currently constructs a Server without validating
opts.McpPath, which can later panic when passed to http.ServeMux.HandleFunc;
update NewServer to validate ServerOptions.McpPath (e.g., ensure it's non-empty
and begins with "/") and return a descriptive error instead of creating the
Server when validation fails (reference NewServer, ServerOptions, Server, and
McpPath), and consider mirroring the same validation in the config validation
code (internal/config/validate.go) to catch bad config earlier.

In `@Makefile`:
- Around line 12-14: There are duplicate Makefile targets named "test" (and
duplicate .PHONY entries) which causes the latter to override the former; remove
the redundant definition and consolidate so there's a single .PHONY: test
declaration and one "test:" target (the existing recipe "go test ./..." should
be kept) — update the Makefile to keep only one "test" target and its .PHONY
entry and delete the other duplicate to avoid silent override.

---

Nitpick comments:
In `@cmd/dir2mcp/main.go`:
- Around line 10-14: Cobra may already print errors before Execute() returns,
causing duplicate stderr output from main's cli.Execute() handling; to avoid
this, set the root command's SilenceErrors flag (e.g., modify the rootCmd in
internal/cli/root.go by initializing rootCmd.SilenceErrors = true, e.g., in an
init() that references rootCmd) so that only your explicit error printing in
main (the main function calling cli.Execute()) emits the error to stderr.

In `@go.mod`:
- Around line 3-15: Update the module declarations to use current Go and
dependency versions: bump the "go" directive from 1.22 to a newer stable release
(e.g., 1.21+ or the project's target), update golang.org/x/term from v0.28.0 to
the latest (e.g., v0.41.0), and bump github.com/spf13/cobra from v1.8.1 to the
latest (e.g., v1.10.2) while keeping gopkg.in/yaml.v3 at the current secure
version; then run go get (or go get <module>@latest) and go mod tidy to resolve
transitive updates and verify no build/regressions in functions that import
these packages (search for imports of golang.org/x/term and
github.com/spf13/cobra to validate compatibility).

In `@internal/cli/reindex.go`:
- Around line 13-16: runReindex currently returns nil silently; update the
function (runReindex) to print a clear user-facing message such as "Reindexing
not yet implemented — feature pending" (or similar) before returning so users
get feedback, and add "fmt" to the imports to support printing; keep the
existing comment about calling ingest.NewIndexer(...).FullRebuild() when the
ingestion pipeline is implemented.

In `@internal/cli/up.go`:
- Around line 102-105: The conditional block checking upPublic and comparing
cfg.Server.Listen to "0.0.0.0:0" and reassigning listenAddr is redundant because
listenOverride already sets cfg.Server.Listen when upPublic is true; remove the
entire if block that contains the no-op reassignment (the variables involved are
upPublic, listenOverride, cfg.Server.Listen and local listenAddr) so listenAddr
is simply initialized from cfg.Server.Listen without the unnecessary
conditional.
- Around line 175-176: The call to enc.Encode(out) ignores its returned error;
update the json.NewEncoder(os.Stdout) usage so you capture err :=
enc.Encode(out) and handle it (e.g., write a clear error to stderr with
fmt.Fprintln or processLogger/error logger and exit non-zero) so NDJSON
user-facing failures (closed stdout, encoding issues) are surfaced; the change
should be made where json.NewEncoder(os.Stdout) and enc.Encode(out) are used
(refer to enc and Encode).

In `@internal/config/load_test.go`:
- Around line 23-24: The test currently calls os.Setenv("MISTRAL_API_KEY",
"test-key") and defers os.Unsetenv, which can clobber an existing environment
value; update the test to preserve and restore the original value (read via
os.LookupEnv before setting, and in defer restore either the original value with
os.Setenv or call os.Unsetenv if it was absent) or simply replace the call with
t.Setenv("MISTRAL_API_KEY", "test-key") in the test function (Go 1.17+) to
guarantee automatic cleanup; modify the test in load_test.go around the env
setup (the os.Setenv/os.Unsetenv usage) accordingly.

In `@internal/config/load.go`:
- Around line 41-46: The current os.ReadFile call in load.go silently ignores
all read errors; change the logic in the function that reads configPath (the
os.ReadFile call and following error handling around cfg) to only treat "file
not found" as optional (use os.IsNotExist(err) or similar) and return a wrapped
error for any other I/O or permission error so callers see real failures; keep
the existing yaml.Unmarshal error behavior for malformed YAML into cfg.

In `@internal/config/validate.go`:
- Around line 12-16: The two branches in validate.go both return the exact same
error for missing MISTRAL_API_KEY; simplify by removing the duplicate branch and
returning the error once (eliminate the conditional around nonInteractive).
Locate the block using the nonInteractive variable in the function that checks
MISTRAL_API_KEY and replace the conditional returns with a single return of the
formatted error string so the logic is identical but not duplicated.

In `@internal/retrieval/engine.go`:
- Around line 31-35: The Citation.Span field is declared as interface{} but
should use the concrete type for clarity and safety—change the Citation struct's
Span field to *model.Span (update the Citation definition in
internal/retrieval/engine.go) and add or adjust the model import if needed; then
update any callers or assignments (constructors, unmarshalling, comparisons)
that currently treat Span as interface{} to pass or handle *model.Span values
instead so types compile cleanly.
- Around line 12-15: The NewEngine constructor currently ignores its parameters
(stateDir, rootDir, cfg) and returns a bare stub Engine; add a short TODO
comment inside the NewEngine function noting that these parameters will be used
when wiring up the real implementation (e.g., "TODO: wire stateDir, rootDir, cfg
into Engine initialization"), and optionally reference them in a noop way
(assign to underscore or set fields on Engine) to avoid linter warnings; update
the NewEngine function and the Engine type initialization accordingly so the
intent is clear to future implementers.

In `@internal/state/state.go`:
- Around line 31-40: The advisory lock creation using os.OpenFile(lockPath,
os.O_CREATE|os.O_EXCL|os.O_WRONLY, 0600) can leave a stale lock file on unclean
exit; update the os.IsExist(err) branch to return a more actionable error that
includes lockPath and a note about manual removal (e.g., "another dir2mcp up is
already running (lock: %s); if this is stale, remove the file and retry").
Locate the lock logic around variables lockPath and stateDir in
internal/state/state.go and replace the existing fmt.Errorf message accordingly
so callers see both the path and remediation steps.

PR #136, opened 2026-04-07
[#99] bug: replace Python ElevenLabs bridge with Go bridge

### LINKED ISSUE #99: bug: elevenlabs bridge.py pollutes indexed corpus and has wrong token path; needs Go rewrite
## Branch

`feat/elevenlabs-voice-demo-v1`

## What the intern was trying to build

ElevenLabs Conversational AI supports server-side webhook tools: you register HTTP endpoints and the ElevenLabs agent calls them during a conversation. The bridge is a thin Flask HTTP layer that translates those calls into MCP JSON-RPC tool calls against a running `dir2mcp` instance, enabling a voice agent to query the indexed corpus and speak answers back to the user.

The concept is sound. The implementation has several bugs.

---

## Bug 1: `bridge.py` is committed inside indexed corpus directories

There are two copies of the bridge:

- `demo-corpus/bridge.py`
- `rezeptbuch/bridge.py`

Both of these directories are passed to `dir2mcp up --dir`. This means the Python source file, its module docstring, Flask route comments, and all its boilerplate prose get chunked, embedded, and stored in the vector index. The script will appear in search results for queries like "how does search work" or "what is a session", polluting retrieval with code that has nothing to do with the knowledge base.

The two files are identical — the intern copy-pasted it rather than extracting it.

**Fix:** Move the bridge to `scripts/elevenlabs-bridge/` or `cmd/elevenlabs-bridge/` (see Bug 4). Never commit demo tooling inside a corpus directory.

---

## Bug 2: Token path resolves relative to the script, not the corpus root

```python
token_path = os.path.join(os.path.dirname(__file__), ".dir2mcp", "secret.token")
```

`__file__` is `demo-corpus/bridge.py`. So the bridge looks for:

```
demo-corpus/.dir2mcp/secret.token
```

But `dir2mcp up` writes the token to `<state-dir>/secret.token`, which by default is `<corpus-root>/.dir2mcp/secret.token` — i.e., relative to the directory passed to `--dir`, not relative to wherever the bridge script happens to live.

When the bridge is moved out of the corpus directory (Bug 1 fix), this path will break entirely. The fallback token discovery needs to derive the path from `MCP_URL` or accept an explicit `--state-dir` flag.

**Fix:** Accept `STATE_DIR` as an environment variable (defaulting to `.dir2mcp` relative to CWD), or require `MCP_TOKEN` to be set explicitly and remove the fragile path fallback.

---

## Bug 3: `/ask` calls `dir2mcp.search` instead of `dir2mcp.ask`

```python
@app.route("/ask", methods=["POST"])
def ask():
    result = _call_tool("dir2mcp.search", {"query": question, "k": 3})
    ...
    return jsonify({
        "result": f"Question: {question}\n\nRelevant document content:\n\n{context}"
    })
```

The bridge calls `dir2mcp.search`, stitches the raw document chunks together with the question, and returns the blob. It then relies on the ElevenLabs agent LLM to synthesize a coherent answer from that raw context — which is unreliable and produces poor voice responses.

`dir2mcp.ask` already performs retrieval + generation and returns a synthesized answer with citations. The bridge should be calling that instead. The intern appears to have used `search` as a workaround because `Engine.Ask()` was still a stub when this was written (tracked in #70). However, that assumption is not documented and the workaround is not flagged, so it will silently remain broken even after #70 is resolved.

**Fix:** Call `dir2mcp.ask` (and update when #70 lands). Add a comment if a temporary workaround is needed.

---

## Bug 4: Python runtime dependency in a Go repo

The bridge requires Python 3, `flask`, and `requests` at runtime. This creates an undocumented external dependency that is:

- Not pinned (`requirements.txt` doesn't exist on the branch)
- Not installed by `make` or any build step
- Out of place in a Go-only repo that otherwise has zero Python

**Fix:** Rewrite the bridge as a Go binary in `cmd/elevenlabs-bridge/` using `net/http`. It can import `dir2mcp`'s own MCP client types directly, share config loading, and be compiled and distributed alongside the main binary. No external Python dependency needed.

---

## Summary

| # | Bug | Severity |
|---|-----|----------|
| 1 | `bridge.py` inside corpus dirs — gets indexed and embedded | High |
| 2 | Token path resolved relative to script, not state dir | Medium |
| 3 | `/ask` calls `dir2mcp.search` not `dir2mcp.ask` | Medium |
| 4 | Python runtime dependency, no `requirements.txt`, wrong language for the project | Medium |

The fix for all four is essentially the same PR: move and rewrite the bridge as a Go binary in `cmd/elevenlabs-bridge/`, wire it to call `dir2mcp.ask` once #70 lands, and document the required environment variables (`MCP_URL`, `MCP_TOKEN`, `PORT`).

yidakra:
## Summary
- add new Go bridge binary at `cmd/elevenlabs-bridge` (net/http)
- route `/ask` through `dir2mcp.ask` via MCP JSON-RPC client code
- add robust bridge config resolution with explicit `MCP_TOKEN` and `STATE_DIR` fallback behavior
- add tests for bridge config and ask-flow behavior
- document bridge env vars/usage in README and `.env.example`
- wire build to include the bridge binary

## Notes
- In this checkout, `demo-corpus/bridge.py` and `rezeptbuch/bridge.py` were not present, so there were no corpus bridge files to remove in this PR.

## Validation
- `go test ./tests/elevenlabsbridge -count=1`
- `go build ./cmd/elevenlabs-bridge`
- `GOCACHE=/tmp/dir2mcp-gocache make check`

Closes #99

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -40,3 +40,12 @@ DIR2MCP_ALLOWED_ORIGINS=
 # docs/x402-payment-adapter-spec.md for details on x402 modes and
 # configuration.
 DIR2MCP_X402_FACILITATOR_TOKEN=
+
+# ElevenLabs bridge (optional helper binary)
+# The bridge forwards ElevenLabs webhook calls to a running dir2mcp MCP
+# endpoint.  It can read the dir2mcp bearer token explicitly or from the
+# state directory token file when MCP_TOKEN is unset.
+MCP_URL=http://127.0.0.1:8087/mcp
+MCP_TOKEN=
+STATE_DIR=.dir2mcp
+PORT=8088
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,15 @@
 # Build binaries. Requires Go 1.24+.
-.PHONY: build build-dir2mcp
-build: build-dir2mcp
+.PHONY: build build-dir2mcp build-elevenlabs-bridge
+build: build-dir2mcp build-elevenlabs-bridge
 
 DIR2MCP_VERSION ?= 0.0.0-dev
 DIR2MCP_LDFLAGS ?= -X dir2mcp/internal/buildinfo.Version=$(DIR2MCP_VERSION)
 
 build-dir2mcp:
 	go build -ldflags "$(DIR2MCP_LDFLAGS)" -o dir2mcp ./cmd/dir2mcp/
+
+build-elevenlabs-bridge:
+	go build -o elevenlabs-bridge ./cmd/elevenlabs-bridge/
 
 # Run dir2mcp up (set MISTRAL_API_KEY first)
 .PHONY: up
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 # dir2mcp
 
-Deploy any local directory as an MCP knowledge server with indexing, retrieval, and citations, built to maximize use of Mistral models across embedding, OCR, transcription, and generation. Optional layers include ElevenLabs voice output and x402 request gating (payment/request‑gating protocol).
+Deploy any local directory as an MCP knowledge server with indexing, retrieval, and citations, built to maximize use of Mistral models across embedding, OCR, transcription, and generation. Optional layers include ElevenLabs voice output via a dedicated bridge binary and x402 request gating (payment/request-gating protocol).
 
 ## Why dir2mcp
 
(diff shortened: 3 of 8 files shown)


yidakra:
Addressed the open review comments on this branch:

- Serialized MCP session initialization to avoid concurrent `initialize` races in `Client.ensureSession`.
- Added session-expiry detection for `404`/session-not-found cases and clear cached session similarly to auth failures.
- Switched bridge RPC/tool strings to shared protocol constants (`protocol.RPCMethod*`, `protocol.ToolName*`).
- Added request body limits + max-size handling (`413`) for `/ask` and `/search`.
- Added `Allow` header emission on 405 in method guard.
- Added hardened server timeouts (`ReadHeaderTimeout`, `ReadTimeout`, `WriteTimeout`, `IdleTimeout`).
- Updated tests to avoid `t.Fatalf` inside handler goroutines; errors are now reported back to the main test goroutine.
- Added regression tests for `Allow` header and oversized ask-body rejection.

Validation rerun:
- `go test ./tests/elevenlabsbridge -count=1`
- `make check`

Pushed in commit: `188f0b3`.

Note on README table comment: the current branch already uses valid single-pipe GFM table separators in the ElevenLabs section.

PR #214, opened 2026-05-28
feat(cli): service command to auto-start daemon at login (launchd)

yidakra:
## Summary

A client's MCP server went `Server disconnected` after a laptop reboot. Root cause is two converging gaps that surface together on reboot:

1. **Daemon doesn't persist.** `dir2mcp up` double-forks a detached `setsid` daemon (`up_daemon.go`) reparented to launchd, but there's no LaunchAgent/login integration — nothing re-spawns it at boot. After reboot the daemon is simply gone (`dir2mcp down` → "no daemon registered").
2. **Credential doesn't persist** unless it's in `.env.local`. A `MISTRAL_API_KEY` that only lived as a shell `export` dies with the login session, so `dir2mcp up` then fails with `CONFIG_INVALID: no embedding provider configured`.

This PR adds `dir2mcp service install|uninstall|status`, which registers a **per-corpus macOS LaunchAgent** that supervises the daemon across reboots and crashes.

### How it works
- Runs `dir2mcp up --foreground` under launchd with `RunAtLoad` + `KeepAlive`. `--foreground` (`up.go`) runs the server in-process (no double-fork), so launchd can supervise it directly. `connection.json` is still published, so the existing MCP client entry reconnects; `dir2mcp status` (sqlite/corpus-snapshot based) is unaffected.
- `WorkingDirectory` is set to the corpus root. This is load-bearing twice over: it indexes the right directory **and** makes the booted daemon resolve credentials from `.env.local` (read cwd-relative by `loadDotEnvFiles`), which matters because launchd does **not** inherit shell `export`s.
- Label is `com.dirstral.<server-name>`, reusing the stable per-corpus `identity.AutoServerName` (slug + path hash).
- `install` runs a **credential preflight** and warns (does not block) when no persisted credential is found in `.env.local`/`.env`, pointing the operator at `dir2mcp config init` — closing the exact trap the client hit.
- Idempotent install (`bootout` → `bootstrap` → `kickstart -k`); `uninstall` unloads and removes the plist; `status` reports installed/running.

### Scope
- macOS/launchd only this PR. Non-darwin returns a clear unsupported error (`service_other.go`); a **Linux systemd user-unit** backend is a planned follow-up behind the same `serviceManager` interface.
- CLI/operator concern only — no MCP tool/spec surface change. Added to the canonical command map + repo-split boundary allowlist accordingly.

## Test plan
- [x] `make check` green (build, vet, gocyclo ≤ 15, full suite, python, release build)
- [x] Pure helpers unit-tested cross-platform: label derivation, plist rendering incl. XML escaping, `.env.local`/`.env` credential scanner, dispatch guard rails (`service_test.go`)
- [x] launchd backend tested via injected `runCmd` (darwin-tagged): install writes plist + issues `bootout`/`bootstrap`/`kickstart`; bootstrap failure surfaces; uninstall removes plist; status parses `state = running` (`service_darwin_test.go`)
- [x] Smoke: usage lists `service`; `service`/`service bogus` exit 2; `service status` reports not-installed via read-only `launchctl print`
- [ ] **Live launchd install/uninstall not exercised in CI** — the unit tests mock `launchctl` (a real run mutates the login session). Recommend a manual install→reboot→`status` pass before relying on it in the field.

## Summary by CodeRabbit

* **New Features**
  * Added a `service` subcommand with `install`, `uninstall`, and `status` to manage an auto-start agent on macOS.

* **Documentation**
  * README updated with service command docs and an “Auto-start at login (macOS)” guide, including note about persisted credentials required for the agent.

* **Platform**
  * Service management is available on macOS; other platforms report unsupported.

* **Tests**
  * Added tests covering macOS agent install/uninstall/status and related validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dirstral/dir2mcp/pull/214?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,10 +164,17 @@ DIR2MCP_DEMO_TOKEN="$(cat .dir2mcp/secret.token)" \
 | `uninstall <client>` | Remove dir2mcp from a supported MCP client |
 | `doctor <client>` | Run client-integration diagnostics |
 | `print-config <client>` | Print the MCP-server JSON snippet a client expects |
+| `service install\|uninstall\|status` | Auto-start the daemon at login so the corpus survives a reboot (macOS launchd) |
 | `version` | Print version |
 
 Running `dir2mcp` with no arguments prints usage, which you can consult anytime to see available commands.
 `ask`, `search`, `open-file`, and `list-files` are legacy compatibility shims; new client/orchestrator UX belongs in `dirstral-cli`.
+
+### Auto-start at login (macOS)
+
+`dir2mcp up` runs a background daemon, but it does not come back on its own after a reboot. `dir2mcp service install` registers a per-corpus launchd agent that restarts `dir2mcp up --foreground` at every login (and on crash), so the MCP server stays connected across reboots. Use `service status` to inspect it and `service uninstall` to remove it.
+
+The launchd job starts from a clean environment and will **not** inherit a `MISTRAL_API_KEY` you only `export`ed in a shell. Persist the credential first with `dir2mcp config init` (writes `.env.local` in the corpus directory) so the booted daemon can find it; `service install` warns when no persisted credential is present.
 
 ## MCP Tools
 
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -74,6 +74,7 @@ var commands = map[string]struct{}{
 	"doctor":         {},
 	"print-config":   {},
 	"support-bundle": {},
+	"service":        {},
 	"version":        {},
 }
 
@@ -453,8 +454,9 @@ func (a *App) runCommand(ctx context.Context, globalOpts globalOptions, remainin
 }
 
 // runSimpleCommand dispatches the non-up/down commands (status, reindex,
-// config, bridge, install, uninstall, doctor, print-config) plus the legacy
-// shims, returning handled=false when command is none of these.
+// config, bridge, install, uninstall, doctor, print-config, support-bundle,
+// service) plus the legacy shims, returning handled=false when command is
+// none of these.
 func (a *App) runSimpleCommand(ctx context.Context, globalOpts globalOptions, command string, args []string) (int, bool) {
 	if code, handled := a.runLegacyShimCommand(ctx, globalOpts, command, args); handled {
 		return code, true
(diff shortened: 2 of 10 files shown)


yidakra:
### Live launchd verification ✅

Ran a full cycle on a throwaway corpus (`/tmp/dir2mcp-live-*` with a persisted dummy `MISTRAL_API_KEY`), since the unit tests mock `launchctl`:

- **install** → wrote `~/Library/LaunchAgents/com.dirstral.dir2mcp-...plist` (correct `ProgramArguments` = `dir2mcp up --foreground`, `WorkingDirectory` = corpus, `RunAtLoad`/`KeepAlive` true), bootstrapped, and started.
- **status** → `installed: true, running: true`; `launchctl print` confirmed `state = running`, `keepalive | runatload`, live pid. Daemon actually bound — `connection.json`, `meta.sqlite`, and HNSW vector files appeared in the corpus state dir.
- **KeepAlive (reboot proxy)** → `kill -9` the daemon pid `56624`; launchd respawned it as `56907` within ~1s. This is the same mechanism (`RunAtLoad`) that restores it after a reboot.
- **uninstall** → booted out + removed the plist; `status` → `not installed`; no lingering process.

Temp corpus removed afterward; no changes left on the machine.

yidakra:
### All review findings addressed (db3bccc)

| Finding | Reviewer | Fix |
|---|---|---|
| Path traversal via `--name` | Copilot | `validateServiceName()` rejects `/`, `\`, absolute paths before they reach `serviceLabel()`/`plistPath()` |
| `plistPath` defense-in-depth | Copilot | `filepath.Base(label)` clamps the label so any bypass can't escape `~/Library/LaunchAgents` |
| Global flags not forwarded to service | Copilot + CodeRabbit | `--config`/`--state-dir` from global options are appended to `spec.Args` at install time |
| Status swallows non-"not found" errors | CodeRabbit major | `status()` now inspects output for known-absent phrases ("could not find service" / "not found in domain"); any other failure is returned as an error |
| State dir `0o755` → `0o700` | CodeRabbit major | Changed |
| `scanner.Err()` not checked | CodeRabbit minor | Documented conservative false-on-error behaviour; consistent with `os.Open` error handling |
| Test: "not found" vs command failure | CodeRabbit nitpick | `TestLaunchdManager_StatusReportsNotInstalled` now uses explicit "Could not find service" output; `TestLaunchdManager_StatusErrorsOnCommandFailure` added for non-absent failures; `TestLaunchdManager_PlistPathClampsTraversal` added for the `filepath.Base` defense; `TestValidateServiceName` added |
| Docstring coverage 22.73% < 80% | CodeRabbit | Doc comments added to all unexported functions in `service.go` and `service_darwin.go` |

`make check` green locally.

Issue #269, opened 2026-06-12
index: pgvector vector-store backend (optional, behind pluggable model.Index)


yidakra:
Implement Postgres + pgvector as an OPTIONAL external vector-store backend behind the pluggable `model.Index` (#247). The "boring infrastructure" Tier-C alternative to Qdrant (#268). In-memory HNSW stays the default; this is opt-in.

## Why pgvector (vs Qdrant)
- **Single service for vectors AND payload**: if metadata ever moves off sqlite to Postgres, one store holds vectors + document/chunk metadata with TRANSACTIONAL consistency and full SQL filtering — no two-store sync.
- **Ubiquitous / already-run**: many deployments already operate Postgres; no new specialized service to learn, secure, back up.
- **Pure-Go client**: `jackc/pgx` is pure Go, so adopting pgvector keeps `CGO_ENABLED=0` and the static single-binary build intact (the client side; the server is external). Contrast with sqlite-vec, which is a C extension incompatible with dir2mcps `modernc.org/sqlite` (see declined list in #247).

## Trade-offs (vs Qdrant)
- pgvector HNSW/IVFFlat is solid but its quantization + filtered-vector-search story is weaker than Qdrant at media scale; Qdrant remains preferred for the large-archive/quantization case.
- Still an external service (network + ops), so heavier than the Tier-B pure-Go on-disk option (#246) for single-node use.

## Scope
- pgvector `model.Index` adapter via `jackc/pgx`: schema (vector column + payload columns / jsonb), upsert, delete, KNN search with SQL `WHERE` payload filters (path prefix, glob, doc_type, modality, time span, language, speaker), index creation (HNSW).
- Map dir2mcp filtering onto SQL predicates applied DURING search (not Go-side overfetch).
- Map embed-identity to a table/collection identity; recreate on mismatch.
- Config: `index.backend: hnsw|qdrant|pgvector` + connection (dsn, schema, table). Default hnsw.
- Tests against an ephemeral Postgres+pgvector (container), gated like the integration suites.

## Decision rule
Choose pgvector over Qdrant when you already run Postgres / want one store for vectors+metadata with SQL filtering, and corpus scale does not demand Qdrants quantization/on-disk profile.

## Depends on / relates
- Depends on #247 (formalized `model.Index`) and the spec gate (#239/#251).
- Tier-C sibling of #268 (Qdrant); alternative to the Tier-B pure-Go on-disk store (#246). Part of dual-machine epic #250.

## Acceptance
[...]

Issue #268, opened 2026-06-12
index: Qdrant vector-store backend (optional, behind pluggable model.Index)


yidakra:
Implement Qdrant as an OPTIONAL external vector-store backend for dir2mcp, behind the pluggable `model.Index` interface (#247). In-memory HNSW stays the DEFAULT; Qdrant is opt-in for scale/distributed deployments. Must not regress the local-first single-binary experience ([[keep general]] — no service dependency forced on small/local users).

## Why (upside, vs in-memory HNSW)
- **Memory ceiling**: HNSW loads ALL vectors into RAM at startup (`map[uint64][]float32` + gob). A years-deep media archive is millions of chunks; Qdrant on-disk/memmapped vectors + quantization (scalar/binary/product, 4-32x smaller) serve far larger corpora on modest hardware.
- **Shared store for distributed workers**: collapses #246 (vector portability) and #248 (distributed embedding) — workers `upsert`, server queries, no custom vector-in-SQLite + HNSW-rebuild plumbing. **Adopting Qdrant may OBVIATE #246.**
- **Filtered search**: payload filters applied DURING search (filtered HNSW) instead of dir2mcps current overfetch-then-filter-in-Go. Compounds with diarization (#266), language metadata (#267), and time-range queries.
- **Incremental upsert/delete + instant restart** (no full index reload).

## Scope
- Qdrant client + a `model.Index` adapter: create/configure collection, upsert vectors with payload (chunk_id, rel_path, doc_type, modality, time span, language, speaker), search, delete.
- Map dir2mcp filtering (path prefix, glob, doc types, secret exclusion) onto Qdrant payload filters (the real win — do NOT just replicate Go-side overfetch).
- Map embed-identity to a collection identity; reindex/recreate on mismatch.
- Config: `index.backend: hnsw|qdrant` + connection (url, api key, collection, quantization, on-disk). Default hnsw.
- Lifecycle: how Qdrant is provisioned (external service; document run/secure/backup; do NOT bundle into the local binary).
- Tests against an ephemeral Qdrant (container) gated like the integration suites.

## Decision rule (when Qdrant is worth it)
Stay on HNSW while the corpus fits comfortably in box RAM and dir2mcp is single-process. Reach for Qdrant when ANY of: corpus exceeds single-box RAM, deployment goes distributed (workers + server sharing vectors), or metadata-filtered search becomes a primary query pattern.

## Depends on / relates
[...]

Issue #250, opened 2026-06-11
[epic] distributed dual-machine ingest/embedding: corpus on S3/NFS + GPU VPS


yidakra:
Tracking epic for running dir2mcp across two machines: the archive (PDFs, video, audio) on **S3 or NFS**, and a **GPU VPS** doing the heavy embedding/OCR/transcription, separate from orchestration and serving.

## Framing
Two orthogonal axes, each with an existing seam:
1. **Where the bytes live** — corpus on S3/NFS instead of local disk. Touches discovery (`internal/ingest/discover.go`) and media reads (`internal/index/embedding_worker.go loadMediaInput`).
2. **Where GPU compute runs** — dir2mcp does NO local GPU work today (embed/OCR/STT all go to a provider `base_url`). So the GPU VPS is simply a **self-hosted provider endpoint**, not new in-process compute.

The load-bearing assumption to watch: vectors currently live only in the in-memory HNSW + local gob files (`.dir2mcp/{text,code}.idx`) — not even in SQLite. Distributing requires making vectors portable.

## Governance
Behavior changes here need a `dirstral-spec` PR merged first, then a gated submodule re-pin — see #239.

## Phase 1 — self-hosted provider + mounted corpus (config/ops, near-zero code)
- [ ] #239 spec: remote-corpus + distributed/self-hosted embedding contract
- [ ] #240 provider: self-hosted OpenAI-compatible embed/OCR/STT endpoints
- [ ] #241 docs: dual-machine deployment guide (daemon on GPU VPS, corpus via NFS/S3-FUSE)

## Phase 2 — native CorpusFS abstraction (when FUSE is not enough)
- [ ] #242 ingest: CorpusFS abstraction (local + S3)
- [ ] #243 ingest/index: S3 range-read media access for PDF/video windows
- [ ] #244 config: corpus source configuration (S3/NFS)
- [ ] #245 store: incremental change detection for remote corpora (ETag/last-modified)

## Phase 3 — distributed worker pool (only if a single GPU node is the bottleneck)
- [ ] #248 ingest/index: distributed embedding workers via job queue
- [ ] #249 cli: standalone embed-worker mode

## Vector-store tiering (pluggable `model.Index`)
HNSW (pure-Go, in-RAM) stays the DEFAULT; tiers below are opt-in via `index.backend`. Full tier map in #247.
- [ ] #247 index: pluggable / externalizable vector store (interface + tier map + declined list)
[...]

Issue #246, opened 2026-06-11
index: pure-Go on-disk/memmapped vectors (Tier B: local, beyond-RAM, no CGO)


yidakra:
Today vectors live ONLY in the in-memory HNSW + gob files (`.dir2mcp/{text,code}.idx`), fully loaded into a Go map at startup. That caps corpus size at available RAM and makes vectors non-portable. This issue is **Tier B** of the vector-store tiering (see #247): keep the deployment LOCAL and dependency-free, but remove the all-vectors-in-RAM ceiling.

## Hard constraint: stay pure-Go
dir2mcp is `modernc.org/sqlite` + `CGO_ENABLED=0`, cross-compiled to static single binaries — the basis of the brew-install/local-first UX. The Tier-B solution MUST preserve that. (This is exactly why sqlite-vec is NOT the answer — it is a C extension incompatible with modernc; see #247 declined list.)

## Approach (pure-Go, no native deps)
- Keep the HNSW graph in RAM but move the vector PAYLOADS to disk: a memmapped vector file, or BLOBs in the existing modernc sqlite (handles BLOBs fine).
- Search loads/maps vectors on demand instead of holding every vector in a Go map.
- Result: corpus can exceed RAM on a single node, vectors are durable + portable, startup no longer pays a full-load cost — all with `CGO_ENABLED=0` intact.

## Relationship to the other tiers
- Tier A (default): in-memory HNSW — unchanged.
- Tier C (external/distributed): Qdrant #268 / pgvector #269 — for multi-node, quantization, or SQL/payload-filtered search at scale.
- This Tier-B store also gives the multi-node story a portable on-disk artifact, but the true distributed case (workers + server sharing one store) is better served by Tier C.

## Seams
- `internal/index/hnsw_index.go` (Save/Load/Add, vector storage), `internal/store/sqlite_store.go` (optional BLOB path), `internal/index/embedding_worker.go` (MarkEmbedded path).

## Acceptance
- A corpus larger than available RAM is searchable on a single node with `CGO_ENABLED=0` preserved; vectors survive restart without a full in-memory reload.
- Depends on spec gate (#239). Pluggable interface from #247.

PR #278, opened 2026-06-13
feat(index): formalize pluggable model.Index interface (#247)

### LINKED ISSUE #247: index: make vector store pluggable / externalizable (Qdrant/Milvus/pgvector)
Phase 3 (endgame). Replace the implicit in-memory HNSW with a pluggable `model.Index` so dir2mcp can serve different deployment tiers from one interface, without forcing any dependency on the local-first default.

## Tier map (pick ONE backend per tier — resist redundant adapters)
| Tier | Profile | Backend | Issue |
|---|---|---|---|
| A — local, in-RAM | fits in memory (laptop/Stas) | in-memory HNSW (pure Go) | default, exists |
| B — local, on-disk | single node, corpus > RAM, NO service | pure-Go on-disk/memmapped vectors | #246 |
| C — networked/distributed | multi-node, shared store, filtered search at scale | Qdrant (preferred) / pgvector (alt) | #268 / #269 |

HNSW stays the DEFAULT. Tiers B/C are opt-in via `index.backend`.

## Interface requirements
The `model.Index` interface must support what external/on-disk backends need to be worthwhile — otherwise an adapter cannot capture the benefit:
- upsert with payload metadata, delete,
- payload/SQL-FILTERED search (not just top-k cosine) — dir2mcp currently filters in Go via overfetch (`internal/retrieval/service.go`); that filtering must be expressible to the backend,
- embed-identity mapping (recreate/reindex on mismatch).

## Declined backends (do NOT add — recorded so they are not re-proposed)
- **sqlite-vec** — appealing (lives in sqlite) but it is a **C extension incompatible with dir2mcps `modernc.org/sqlite` (pure Go) + `CGO_ENABLED=0`**. Adopting it means CGO (loses trivial cross-compilation) or shipping per-platform `.so/.dylib` (re-enters the docling/libspatialindex native-lib distribution pain). The Tier-B need it seemed to fill is instead met by pure-Go on-disk vectors (#246).
- **Weaviate** — same tier as Qdrant, no meaningful gain.
- **Milvus** — k8s/billion-scale overkill for dir2mcps scale.
- **Pinecone / managed SaaS** — vendor lock-in + cost; revisit only if dir2mcp becomes a hosted product (x402 angle).
- **Elasticsearch/OpenSearch** — heavy; lexical already covered by FTS5.
- **FAISS** — in-process like HNSW but pulls in CGO/C++, breaking the pure-Go static binary.

## Wildcard (note, not a task)
An S3/object-storage-backed vector store (turbopuffer-style) would align with the S3-corpus direction (#242) — cheap-at-rest for a huge COLD archive. Park until archive economics demand it.

## Seams
- `internal/model/interfaces.go` (`Index`), `internal/index/hnsw_index.go`, `internal/retrieval/service.go`.

## Acceptance
- A non-default backend serves search without loading all vectors into the dir2mcp process; HNSW remains the default with unchanged behavior and `CGO_ENABLED=0`.
- Depends on the spec gate (#239/#251). Enables #246 (Tier B), #268 (Qdrant), #269 (pgvector).

yidakra:
## Summary

Formalizes a pluggable `model.Index` interface (issue #247) so external/on-disk vector backends can implement it, keeping the pure-Go in-memory HNSW as the conforming default with unchanged behavior and `CGO_ENABLED=0`.

Closes #247. Part of epic #250. Enables Tier B on-disk (#246), Qdrant (#268), and pgvector (#269) — those backends build on this interface.

## What changed

- **Core `Index`** (`internal/model/interfaces.go`): `Upsert(ctx, vector, IndexPayload)`, `Delete(ctx, chunkIDs)`, `Search(ctx, vector, k, Filter) []IndexHit`, `Identity`, `Reset`, `Close`.
- **Optional capabilities** (type-asserted, like `LexicalSearcher`/`MultimodalEmbedder`):
  - `Persistable` (`Save`/`Load`) — HNSW implements it; `PersistenceManager` type-asserts and skips backends that own their storage.
  - `FilteringIndex` (`CanFilter`) — when true, retrieval pushes the `Filter` down and trusts backend-filtered results; otherwise falls back to overfetch-then-filter.
- **New types** (`internal/model/types.go`): `IndexPayload`, `IndexHit`, `Filter` (with `IsZero`/`Match` reproducing `matchFilters`: path prefix, `path.Match` glob, case-insensitive doctype set, orphan exclusion).
- **HNSW conforms** (`internal/index/hnsw_index.go`): stores `payloads` + `identity`; `Search` scans cosine + `filter.Match` inline; `CanFilter` always true; `Save`/`Load` move under `Persistable`.
- **Retrieval** (`internal/retrieval/service.go`): `collectVectorCandidates` pushes the filter down via `FilteringIndex` when available (materializing from backend payload, falling back to in-memory chunk metadata), else keeps the widening overfetch loop. Hybrid RRF / media dedup / rerank unchanged.
- **Embedding worker**: `indexChunks` now `Upsert`s with an `IndexPayload` projected from `ChunkMetadata`.
- **Identity wiring**: `index.EnsureIdentity(ctx, idx, configuredIdentity)` per index at startup (`up`/`ask`), `Reset` on mismatch/empty; complements the existing process-level `VerifyEmbedIdentity`.

## Persistence-format decision

Bumped the on-disk filename to `vectors_{text,code}.v2.hnsw`, holding a self-describing gob struct (vectors + payloads + identity) instead of a bare `map[uint64][]float32`. A missing v2 file is treated as **fresh / repopulated by reindex** — legacy bare-map files are never decoded, avoiding gob ambiguity. `reindex` removes both v2 and legacy files.

## Interface deviations / blast radius

- `Index` method signatures changed (`Add`→`Upsert`, `Search` now takes `ctx`+`Filter` and returns `[]IndexHit`, `Save`/`Load` gained `ctx` and moved under `Persistable`). Updated all call sites: `internal/cli/{up,app,reindex}.go`, `internal/index/{persistence,embedding_worker}.go`, `internal/retrieval/service.go`, and every test fake.
- `ExcludeOrphans` is intentionally **not** pushed down: orphan/eviction state lives in the retrieval service's in-memory chunk metadata (not the backend payload), so it's enforced by the post-materialization `matchFilters` re-check. The in-memory metadata stays authoritative for the default HNSW path, so eviction behavior is preserved.
- Scope held: no Qdrant/pgvector/on-disk backends; submodule not re-pinned.

## Tests

New `tests/index/pluggable_index_test.go` (Upsert+Search with doctype/prefix/glob filters, Delete, Identity/Reset, EnsureIdentity, CanFilter), `tests/retrieval/pushdown_test.go` (push-down end-to-end with payload materialization via an external-backend-style fake), and a worker payload-projection test. Existing HNSW/retrieval/persistence tests and fakes updated to the new interface.

`make check` passes (exit 0); `CGO_ENABLED=0 go build ./...` clean.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -693,6 +693,24 @@ func (a *App) configureReranker(ret *retrieval.Service, cfg config.Config) {
 	ret.SetRerankEnabled(true)
 }
 
+// openHNSWForAsk constructs a persisted HNSW index, restores it from its v2
+// snapshot, and reconciles its recorded embed identity with the configured one
+// (issue #247). On any error it closes the index and returns the error.
+func openHNSWForAsk(ctx context.Context, path, identity string) (*index.HNSWIndex, error) {
+	ix := index.NewHNSWIndex(path)
+	if err := ix.Load(ctx, path); err != nil &&
+		!errors.Is(err, model.ErrNotImplemented) &&
+		!errors.Is(err, os.ErrNotExist) {
+		_ = ix.Close()
+		return nil, err
+	}
+	if err := index.EnsureIdentity(ctx, ix, identity); err != nil {
+		_ = ix.Close()
+		return nil, err
+	}
+	return ix, nil
+}
+
 // buildRetrieverForAsk constructs a retrieval service for the ask path:
 // it loads the text/code HNSW indexes, resolves the embed/chat clients,
 // wires reranking, and preloads embedded chunk metadata. It returns the
@@ -703,23 +721,16 @@ func (a *App) buildRetrieverForAsk(ctx context.Context, cfg config.Config, st mo
 		return a.newRetriever(cfg, st), nil, nil
 	}
 
-	textIndexPath := filepath.Join(cfg.StateDir, "vectors_text.hnsw")
-	codeIndexPath := filepath.Join(cfg.StateDir, "vectors_code.hnsw")
-
-	textIx := index.NewHNSWIndex(textIndexPath)
-	if err := textIx.Load(textIndexPath); err != nil &&
-		!errors.Is(err, model.ErrNotImplemented) &&
-		!errors.Is(err, os.ErrNotExist) {
-		_ = textIx.Close()
+	// v2 snapshot filenames + per-index embed-identity reconciliation (issue
+	// #247); see internal/index for the format rationale.
+	identity := cfg.Providers().EmbedIdentity()
+	textIx, err := openHNSWForAsk(ctx, filepath.Join(cfg.StateDir, index.TextIndexFileName), identity)
+	if err != nil {
 		return nil, nil, fmt.Errorf("load text index: %w", err)
 	}
-
-	codeIx := index.NewHNSWIndex(codeIndexPath)
-	if err := codeIx.Load(codeIndexPath); err != nil &&
-		!errors.Is(err, model.ErrNotImplemented) &&
-		!errors.Is(err, os.ErrNotExist) {
+	codeIx, err := openHNSWForAsk(ctx, filepath.Join(cfg.StateDir, index.CodeIndexFileName), identity)
+	if err != nil {
 		_ = textIx.Close()
-		_ = codeIx.Close()
 		return nil, nil, fmt.Errorf("load code index: %w", err)
 	}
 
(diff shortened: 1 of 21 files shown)


yidakra:
@coderabbitai review

yidakra:
@coderabbitai full review

Issue #383, opened 2026-06-24
docling: doc titles show as temp filename; open_file page=N fails on docling PDFs


yidakra:
Two docling quality bugs found testing the stas-legal corpus on 0.9.6 (both confuse Claude during legal Q&A — it can't identify docs and can't open pages).

## 1. Document titles are `dir2mcp-docling-<random>`
`docling.Document.Title()` falls back to the DoclingDocument **name**, which docling derives from the input filename — and dir2mcp feeds docling a temp file named `dir2mcp-docling-*`. So every doc's title becomes e.g. `dir2mcp-docling-1122486701` (visible in `list_files` and search hits). The model then can't recognize documents and guesses wrong paths.

**Fix:** `Title()` returns "" when there's no real title element instead of falling back to the name; the ingest service then runs its content-based title heuristic (`ExtractTitle`) — the same path Mistral OCR uses — yielding a real title like the document heading.

## 2. `open_file page=N` fails on docling-extracted PDFs (DOC_TYPE_UNSUPPORTED)
`sliceFromMetadata`'s `page` case only matched chunks with a plain `page` span, but docling chunks carry a **region** span (page range in `Region.StartPage/EndPage`). So page reads found no metadata match and fell through to reading raw PDF bytes → `looksLikeBinaryContent` → `DOC_TYPE_UNSUPPORTED: pass page=N` (even though page=N was passed).

**Fix:** the `page` lookup also matches region spans covering the requested page (new `regionCoversPage` helper).

## Verification
- `open_file page=1` on a docling PDF now returns the page text (was DOC_TYPE_UNSUPPORTED) — verified live.
- Title fix: `Title()` returns "" with no title element → content heuristic supplies the title; covered by tests.
- Both behaviors have regression tests; `make check` green.

Titles are corrected on the next (re)index; the open_file fix takes effect immediately on the fixed binary.

Issue #381, opened 2026-06-24
docling unusable in dir2mcp-full: probe timeout, bare-binary command, and output cap


yidakra:
End-to-end verification of docling on a real dir2mcp-full 0.9.5 install (89 legal PDFs, ingest.extractor=auto — the default) surfaced three issues that together made docling unusable, even after #371 (rt_detr_v2/cv2) and #376 (--output -):

## 1. Functional-check probe timeout too short
`doclingProbeTimeout` is 20s, but `docling --version` cold-imports the whole torch/transformers/cv2 stack and takes ~26s on a warm M-series Mac (slower on older hardware / first run). The probe times out → docling is deemed "failed functional check" → `auto` silently falls back to Mistral OCR. So dad never actually got docling.

## 2. dir2mcp-full wrapper passes a BARE docling path
The formula wrapper sets `DIR2MCP_DOCLING_COMMAND=/…/docling-venv/bin/docling` (bare path, no template). dir2mcp maps that to `cfg.DoclingCommand` and runs `docling <input>` — no `--to json --output {output}` — so docling writes Markdown into the working dir and nothing to stdout → every doc fails "empty output" and litters the corpus with `dir2mcp-docling-*.md` (which then get indexed). The #376 default-template fix doesn't apply because `DoclingCommand` is non-empty.

## 3. Structured-JSON output cap (100MB) too small for large legal PDFs
docling `--to json` for big multi-hundred-page acts exceeds `maxDoclingStdoutBytes` (100MB): e.g. "VI Rules of Civil Procedure" (920KB PDF) → 101MB JSON. Such docs error "docling command output exceeded limit" and aren't indexed (3 of 89, including the BVI Business Companies Act).

## Fixes
1. Raise `doclingProbeTimeout` to 90s (probe result is memoized per process; daemon readiness window scales off it).
2. Expand a bare-binary `DoclingCommand` (single token) to the default `--to json --output {output} {input}` template.
3. Raise the docling output cap to comfortably cover large legal PDFs.

Verified on the stas-legal corpus: with 1+2, 86/89 extract via docling (no fallback, no pollution); 3 fixes the remaining large docs.

Issue #376, opened 2026-06-23
docling extractor: '--output -' yields empty output + litters corpus with dir2mcp-docling-*.md artifacts


yidakra:
Found while verifying the #371 docling fix end-to-end (dir2mcp-full 0.9.4_1, docling 2.92.0). With rt_detr_v2 + cv2 now working, docling **loads and converts** standalone — but dir2mcp's docling extractor still fails for every document, for two compounding reasons:

## 1. `--to json --output -` no longer streams to stdout (docling 2.92.0)
`internal/ingest/docling_extractor.go` uses `defaultDoclingCommand = "docling --to json --output - {input}"` and reads the JSON from **stdout**. In docling 2.92.0, `--output -` does **not** write to stdout — it exits 0 and writes the JSON to a file (a directory literally named `-`, or the input basename in CWD). So `run()` sees empty stdout and returns `docling command produced empty output` for every doc → all docling extractions fail (doc status=error), the corpus never re-extracts/embeds.

Repro:
```
docling --to json --output - some.pdf > out.json   # out.json is 0 bytes, exit 0
docling --to json --output /tmp/d  some.pdf         # /tmp/d/some.json is 4.5MB  ✓
```

## 2. docling output artifacts are written into the corpus and indexed
During a docling reindex, `dir2mcp-docling-*.md` files (and a `-/<name>.json`) landed in the **corpus root** (`~/.../stas-legal/`) and were picked up as documents (md doc count jumped 2→9). So the broken output path also pollutes the corpus with its own artifacts, which then get indexed (RAG answers from docling temp files).

## Impact
docling is effectively unusable in 0.9.4 even with the formula fixed; `ingest.extractor=mistral` is the working path. (Linux likely same, since this is the CLI invocation, not platform-specific.)

## Fix direction
- Invoke docling with `--output <tempdir>` (a real temp dir, NOT the corpus/CWD) and read the produced `.json`/`.md` file, instead of relying on `--output -` + stdout.
- Ensure all docling temp input/output lives under an os.MkdirTemp dir and is cleaned up, never the corpus root.
- Add the real-conversion smoke (see #372) to catch "empty output" regressions.

Related: #371 (formula rt_detr_v2/cv2 fix, merged), #372 (functional check should actually convert).

Issue #373, opened 2026-06-23
BM25 lexical search broken: bm25() returns NULL → Scan into float64 fails, no keyword fallback


yidakra:
## Problem
Every hybrid search logs:
```
hybrid: BM25 search failed, falling back to vector-only:
sql: Scan error on column index 7, name "score": converting NULL to float64 is unsupported
```
`SQLiteStore.SearchBM25` (internal/store/sqlite_bm25.go) selects `bm25(chunks_fts) AS score` and scans it into a `float64`. On the external-content FTS5 table (`content='chunks', content_rowid='chunk_id'`) `bm25()` returns **NULL** for matched rows — reproduced directly:

```sql
SELECT c.chunk_id, bm25(chunks_fts) AS score
FROM chunks_fts JOIN chunks c ON c.chunk_id=chunks_fts.rowid
WHERE chunks_fts MATCH '"report" OR "suspicious"' ORDER BY score LIMIT 5;
-- chunk_ids returned, score column NULL for every row
```
The scan then errors and the **entire lexical path is dropped** — so when the vector index is empty/unavailable there is no keyword fallback at all (search returns 0).

## Fix
- Defensive: scan into `sql.NullFloat64` (treat NULL as 0 / lowest rank) so a NULL score degrades to unranked-but-present rather than killing the whole query.
- Root: investigate why `bm25()` yields NULL on this external-content FTS5 config (FTS index population / rebuild / `content_rowid` mapping) and fix so real BM25 ranking is restored.

Found while diagnosing #364.

PR #385, opened 2026-06-24
chore(release): end-to-end MCP smoke gate before tagging releases

yidakra:
## Why
`make check`/`make ci` validate code but never exercise the **live retrieval surface** (`ask`/`search`/`open_file`) — which broke repeatedly across v0.9.x and only surfaced when a human asked the Claude client:
- #364 empty embeddings, #376 docling `--output -`, #381 docling probe/bare-binary/cap, #373 BM25 NULL, #383 open_file page on docling PDFs.

Every one of those would have been caught by actually running the stas-legal guide's verification questions through an MCP client before shipping.

## What
`scripts/release_smoke.py` — speaks the MCP streamable-HTTP protocol to a **running** dir2mcp daemon and asserts:
- `stats`: indexing stopped, `errors==0`, `embedded_ok>0`
- `ask` (each question): non-empty answer **and ≥1 citation**, no tool error
- `search`: ≥1 hit
- `open_file(real pdf, page=1)`: non-empty, non-binary text

Exits non-zero on any failure. Questions are overridable via `--question`.

Wired as `make release-smoke STATE_DIR=…` and documented in CLAUDE.md's Releasing section as a **pre-tag gate**: build the candidate binary, reindex a representative corpus on it, wait for `pending=0`, then run the gate; don't tag unless all-pass. It needs a live daemon + provider creds, so it's a manual gate (not in `make ci`).

## Verification
- **Fails** on a 0.9.6 daemon (correctly catches the missing #383 open_file fix).
- **All-pass** on a v0.9.7 daemon against the stas-legal corpus (3 cited answers, 5 search hits, open_file page text).

`test-release-tools` only runs `test_*.py`, so this script is ignored there; `make check` stays green.

## Summary by CodeRabbit

* **New Features**
  * Added a new pre-release smoke check for live release candidates.
  * Release instructions now include a manual end-to-end validation flow before tagging.

* **Bug Fixes**
  * Added checks that verify indexing has finished, no errors occurred, search returns results, and file text extraction works as expected.
  * The smoke check now fails clearly if any live retrieval path does not behave correctly.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -75,14 +75,19 @@ run the gate against the running daemon:
 ```bash
 make build
 # point a daemon at a freshly-reindexed corpus on THIS binary, wait for pending=0, then:
-make release-smoke STATE_DIR=/path/to/corpus/.dir2mcp
+make release-smoke STATE_DIR=/path/to/corpus/.dir2mcp                  # server contract (HTTP)
+make release-smoke STATE_DIR=/path/to/corpus/.dir2mcp TRANSPORT=stdio  # full mcp-remote bridge
 ```
 
 It speaks MCP to the daemon and asserts: indexing stopped + `errors=0` +
 `embedded_ok>0`; each question returns a grounded answer **with citations**;
 `search` returns hits; `open_file page=1` returns text. Any failure exits
-non-zero — do not tag until it is all-pass. (Manual gate: needs a live daemon +
-the corpus's provider credentials, so it is not part of `make ci`.)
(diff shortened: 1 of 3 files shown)


yidakra:
Added `--transport stdio`: the gate can now drive the actual `bunx mcp-remote` bridge Claude Desktop uses (not just direct HTTP), so it also catches client/bridge-layer regressions — the 'Failed to call tool' class. `make release-smoke TRANSPORT=stdio`. Verified all-pass on a v0.9.7 daemon against the stas-legal corpus on both transports.

Issue #439, opened 2026-06-26
quality-gate: quarantined garbage is STILL served via BM25/hybrid (+cited); language/repetition/density heuristics misfire on Russian/CJK/music


yidakra:
**Part of #395.** The quality gate quarantines bad OCR/STT output, but the quarantine is incomplete and several heuristics are Latin-centric (hits the keep-general-purpose / handle-Russian mandate). Distinct from #422 (status not set).

## F1 [HIGH] Quarantined content still returned by BM25/hybrid search + cited in ask (verified)
Quarantine sets `embedding_status='error'` so the embed worker skips it — but that only blocks the **vector** path. The FTS5 trigger `chunks_ai AFTER INSERT` indexes **every** chunk's text into `chunks_fts` regardless of status (verified sqlite_store.go:539), and `SearchBM25` filters only `WHERE chunks_fts MATCH ? AND c.deleted = 0` — **no `embedding_status` filter** (verified sqlite_bm25.go:27). Hybrid is on by default. So gibberish OCR / STT hallucination loops / wrong-language transcripts the gate "quarantined" are **still returned by search and appear as citations in ask answers**. The gate gives false confidence. **Fix:** add `AND c.embedding_status != 'error'` to SearchBM25 (or make the FTS trigger conditional / delete the FTS row on quarantine).

## F2 [MED] Whole-representation granularity loses good pages / keeps garbage
`screenOutputQuality` evaluates the **full** rep text and applies one decision to all chunks (service.go:2110/2154/2350 → represent.go:225). A 100-page PDF with 30 gibberish pages quarantines all 100 (70 good pages lost); 15% garbage stays under threshold and all garbage is indexed. OCR/structured chunking is page/block-aware so it *could* screen per-page. **Fix:** per-chunk/per-page screening.

## F3 [MED] Pinned STT language quarantines legitimate other-language transcripts (Russian-corpus footgun)
`languageDetector` (detectors.go:119) quarantines when >50% of letters are off the expected script, where "expected" is a single fixed `STTLanguage` for the whole corpus. Pin `language: ru` (TV Rain) → any English clip/interview/song → ~100% off-script → **quarantined, content lost**. Contradicts the "auto-detect language, stay general-purpose" mandate. (If STTLanguage is empty/auto, the detector self-skips — so it only bites operators who pin a language, which is encouraged.) **Fix:** per-file language detection feeding the gate, or only flag when *both* detected and expected disagree.

[...]

PR #448, opened 2026-06-27
fix(store,config): exclude quarantined chunks from BM25 + broaden default secret patterns

### LINKED ISSUE #443: SECURITY: default secret-scan patterns miss real provider key formats (sk-…, MISTRAL_API_KEY=, Cohere/Gemini) → a key pasted into a corpus file gets indexed and leaks via ask/search
**Part of #395.** The content secret-scan (which prevents indexing files that contain secrets) does not match the key formats this product actually issues.

## Verified
Default `SecretPatterns` (`config.go:916-921`): `AKIA[0-9A-Z]{16}`, an aws-secret shape, a JWT/bearer shape, `token[:=]{20,}`, and `sk_[a-z0-9]{32}|api_[A-Za-z0-9]{32}`. These do **not** match:
- **OpenAI/Mistral `sk-…`** (hyphen, mixed case) — the pattern requires `sk_` (underscore, lowercase).
- **`MISTRAL_API_KEY=<32 alnum>`**, Cohere, Gemini, ElevenLabs key shapes — no covering pattern (`API_KEY=` ≠ `api_<32>`).

So a user who pastes a real provider key into any **non-`.env`** text/data file in the corpus (`providers.yaml`, `notes.md`, `config.json` → classified `data`/`md`/`text`, not `ignore`) is **not** flagged by the scan (`service.go:1587`), so the file is chunked, embedded, and **retrievable via `search`/`ask`** → the API key leaks through the retrieval surface. Secondary: the scan only inspects the first `secretScanSampleBytes=64KB` (`service.go:1594`), so a secret deeper in a large file is missed regardless.
**Fix:** add `sk-[A-Za-z0-9-]{20,}` and provider-key shapes (incl. `[A-Z_]*API_KEY\s*[:=]`) to the default patterns; scan the whole file (or a larger sample) for secrets.

## Wizard key handling (related, lower severity)
- [MED] The setup wizard saves keys on **non-empty alone** — no format check, no live API probe (`wizard.go:214`); `emitSetupVerification` only checks the env var resolves ("no network calls", `config_cmd.go:305`). So a typo'd/expired key is written and "✓ provider ready" printed, then every index/ask fails with auth later. **Fix:** optional probe-validate before save.
- [LOW-MED] `saveEnvLocalKey` (`app.go:597`) writes `prefix+value` verbatim with no newline guard (the env-capture sibling path rejects `\r\n`, `service.go:361`) → a multi-line paste corrupts/injects into `.env.local`.
- [LOW] `saveEnvLocalKey` uses `os.WriteFile(...,0o600)` which only sets the mode on **create** → a pre-existing looser-perm `.env.local` keeps its mode → secret in a world-readable file. **Fix:** explicit `os.Chmod(path,0o600)`.

yidakra:
Two small, disjoint fixes in one PR.

## #439 (F1): quarantined garbage still served via BM25/hybrid
`internal/store/sqlite_bm25.go` `SearchBM25` filtered only on `c.deleted = 0`, with no embedding-status guard. Chunks the quality gate quarantined (`embedding_status='error'`) were still returned by the lexical/hybrid path and cited.

Fix: add `AND c.embedding_status != 'error'` to the WHERE clause. Quarantined chunks are now excluded; `'pending'` chunks (which need no embedding to match lexically) remain searchable. The FTS trigger is unchanged — the WHERE filter is the surgical fix.

This is only the BM25-quarantine-leak part of #439. The issue also tracks other heuristic items, so this PR **partially addresses #439 (F1: BM25 quarantine filter)** rather than closing it.

## #443: default secret-scan misses real provider key formats
`internal/config/config.go` default `SecretPatterns` matched `sk_`/`api_` but not OpenAI/Mistral `sk-...` (hyphen, mixed case) or `PROVIDER_API_KEY=<value>` assignments.

Fix: add two patterns to the defaults:
- `sk-[A-Za-z0-9-]{20,}` — OpenAI/Mistral hyphenated keys (e.g. `sk-proj-...`)
- `[A-Z][A-Z0-9_]*API_KEY\s*[:=]\s*\S{16,}` — generic `MISTRAL/COHERE/GEMINI_API_KEY` assignments

Both compile under Go's RE2 engine and stay surgical enough to avoid flagging ordinary prose. The 64KB scan-sample bound is intentionally left untouched (separate concern).

## Tests
- `tests/store/sqlite_bm25_quarantine_test.go`: asserts an `embedding_status='error'` chunk is NOT returned by `SearchBM25` while an `'ok'` chunk matching the same term IS.
- `tests/config/secret_patterns_test.go`: asserts the new defaults match `sk-proj-...` and `MISTRAL_API_KEY=<32chars>` (plus Cohere/Gemini shapes) and do NOT match ordinary sentences.

`gofmt`, `go build ./...`, `go vet`, and the focused `internal/store` / `internal/config` / `tests/store` / `tests/config` suites all pass, as does `make check` (the lone failure was a pre-existing uninitialized `dirstral-spec` submodule, unrelated to this change).

Closes #443.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -919,6 +919,16 @@ func Default() Config {
 			`(?i)(?:authorization\s*[:=]\s*bearer\s+|(?:access|id|refresh)_token\s*[:=]\s*)[A-Za-z0-9_-]{8,}\.[A-Za-z0-9_-]{8,}\.[A-Za-z0-9_-]{8,}`,
 			`(?i)token\s*[:=]\s*[A-Za-z0-9_.-]{20,}`,
 			`sk_[a-z0-9]{32}|api_[A-Za-z0-9]{32}`,
+			// OpenAI/Mistral-style hyphenated secret keys, e.g. "sk-proj-..."
+			// and "sk-...". Case-sensitive lowercase "sk-" prefix; the {20,}
+			// run of key chars keeps this from matching ordinary hyphenated
+			// prose (issue #443).
+			`sk-[A-Za-z0-9-]{20,}`,
+			// Generic provider-key assignment, e.g. "MISTRAL_API_KEY=<32 alnum>",
+			// "COHERE_API_KEY: ...", "GEMINI_API_KEY=...". Requires an
+			// uppercase provider prefix plus a 16+ char non-whitespace value so
+			// it gates real credentials without flagging plain text (issue #443).
+			`[A-Z][A-Z0-9_]*API_KEY\s*[:=]\s*\S{16,}`,
 		},
 		ElevenLabsAPIKey:     "",
 		ElevenLabsBaseURL:    "",
--- a/internal/store/sqlite_bm25.go
+++ b/internal/store/sqlite_bm25.go
@@ -28,6 +28,11 @@ import (
 // a vector-only fallback (returning zero results when the vector index was
 // empty). We now scan into sql.NullFloat64 and order NULL-scored rows LAST so a
 // missing score degrades a single hit's rank rather than failing the query.
+//
+// Quarantine filter (#439, F1): chunks the quality gate quarantined carry
+// embedding_status='error'. They must not surface via the lexical/hybrid path,
+// so the WHERE clause excludes them (embedding_status != 'error'). 'pending'
+// chunks remain searchable lexically (they need no embedding to match BM25).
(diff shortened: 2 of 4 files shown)


yidakra:
@coderabbitai review

yidakra:
@coderabbitai full review

Issue #356, opened 2026-06-21
mcp: tool-handler panics drop the connection ("Failed to call tool") and bypass server.log — no recover in tool dispatch


yidakra:
## Symptom (field report, v0.8.0, darwin/arm64)

Intermittent **`Failed to call tool "dir2mcp_open_file"`** and **`Failed to call tool "dir2mcp_list_files"`** in Claude Desktop on a corpus of ~89 OCR'd PDFs. `search`/`ask` succeed; `open_file`/`list_files` sometimes spin for minutes then fail, sometimes succeed. Restarting the chat/server changes which calls fail.

Support bundle: `version=v0.8.0`, `index_backend=memory`, `ingest_pdf_mode=ocr` (docling CLI), HTTP MCP server (`listen_addr 127.0.0.1:0`, `mcp_path /mcp`). **`server.log` was empty** despite the server having answered other queries in the same session.

## Root cause

"Failed to call tool" is a **transport-level** failure (a clean tool error is returned as a result the model reads, e.g. `OCR_NOT_READY`). The MCP tool dispatch has **no panic recovery**:

- `internal/mcp/tools.go:249` — `result, toolErr := tool.handler(ctx, params.Arguments)` is called with no `defer/recover`.
- There is **no `recover()` anywhere in `internal/mcp/`** (dispatch or HTTP middleware chain).

So if an `open_file`/`list_files` handler panics on some input, Go's `net/http` recovers it at the *server* level: it **abruptly closes the connection** (client → "Failed to call tool") and writes the stack to **stderr**, never to the captured `server.log`. This explains all three observations: the transport error, the empty `server.log`, and the process surviving (other queries keep working).

The "spun 5 minutes" is the model retrying the dropped call.

## Fix

1. **Wrap tool-handler dispatch in `defer/recover`** (`tools.go:249`): convert a panic into a proper `INTERNAL_ERROR` `toolExecutionError` (returned as a normal tool-error result, HTTP 200) **and log the stack to the server logger** so it lands in `server.log`. This turns the opaque connection-drop into an actionable error the model can report, and — crucially — **captures the real panic** so the underlying trigger can be fixed.
2. **Support bundle: capture the server's stderr**, not just `server.log`. The actual panic/HTTP-error output has been going to stderr, which the bundle omits — that's why field diagnosis has been blind.

## Secondary (verify separately)

[...]

PR #449, opened 2026-06-27
mcp: route SDK transport tool calls through panic-recovery guard (#401)

### LINKED ISSUE #401: mcp: tool-handler panic crashes the whole daemon on the default (x402-off) path — invokeToolHandler's recover() is bypassed
## Severity: Critical (single tool call can take down the daemon for all clients)

The production SDK transport's tool closure calls the handler **directly**, bypassing the `recover()` guard:

`internal/mcp/transport_sdk.go` (~:399, in `buildSDKServer`'s `AddTool` closure):
```go
res, toolErr := td.handler(ctx, args)   // <-- raw handler, no recovery
```

`Server.invokeToolHandler` (`internal/mcp/tools.go:267`) is the function that holds the `defer func(){ if r := recover(); r != nil {...} }()` added for #356/#357 — but this closure does **not** call it. The go-sdk runs the tool handler in a background goroutine (`jsonrpc2` async dispatch) with **no `recover()` of its own** (verified by grep of the SDK package), and an unrecovered panic in a non-ServeHTTP goroutine is not caught by net/http's per-request recovery → it **aborts the process**.

Asymmetry: the x402-**enabled** path (`payment.go` → `processToolsCall` → `invokeToolHandler`) *is* protected. `x402Enabled` defaults to false, so the **default deployment is the unprotected one**.

False confidence: `tool_panic_recovery_test.go` exercises only `invokeToolHandler`, which the default path never calls — so the regression test passes while the hot path is unguarded.

## Trigger / symptom
Any handler bug that panics (nil-map write, slice OOB, bad type assertion in `handleAskTool`/`handleSearchTool`/`handleTranscribeTool`/etc.) on a single tool call crashes the daemon for **all** clients — the dropped-connection / "Failed to call tool" class #356/#362 aimed to eliminate, still live on the default path.

## Fix
Have the SDK closure call `s.invokeToolHandler(ctx, td, td.Name, args)` instead of `td.handler(ctx, args)` (or add a `recover()` in the closure / via `AddReceivingMiddleware`). Extend the panic-recovery test to drive the SDK transport path, not just `invokeToolHandler`.

yidakra:
## Problem

A panic in any MCP tool handler crashed the entire daemon on the default (x402-off) path.

The go-sdk runs tool handlers in a background goroutine with no `recover()` of its own. The SDK transport closure in `buildSDKServer` (`internal/mcp/transport_sdk.go`) called `td.handler(ctx, args)` **directly**, bypassing `invokeToolHandler` (`internal/mcp/tools.go`) which holds the `recover()` guard added for #356/#357. Only the x402-enabled path routed through `invokeToolHandler`, so the default path was unprotected and a single handler panic took down the process.

## Fix

- In the `buildSDKServer` `AddTool` closure, call `s.invokeToolHandler(ctx, td, td.Name, args)` instead of `td.handler(ctx, args)`. The `toolErr → newToolErrorResult` and `convertToolCallResult` handling stays equivalent, so a panicking handler now yields a clean `INTERNAL_ERROR` isError tool result on every transport path.
- Add `TestSDKTransport_RecoversPanic`, which drives the **production** SDK transport end-to-end (`buildSDKServer` + in-memory client/server sessions) against a panicking handler and asserts the call returns an isError result rather than crashing the process. The existing tests only covered `invokeToolHandler` directly.

## Verification

- `go build ./...`, `go vet ./...` clean
- `make fmt vet cyclo ineffassign misspell` clean (gocyclo under 15)
- `go test ./internal/mcp/... ./tests/mcp/...` passes

Closes #401.

### DIFF
--- a/internal/mcp/tool_panic_recovery_test.go
+++ b/internal/mcp/tool_panic_recovery_test.go
@@ -105,14 +105,14 @@ func TestSDKTransport_RecoversPanic(t *testing.T) {
 	if err != nil {
 		t.Fatalf("server connect: %v", err)
 	}
-	defer serverSession.Close()
+	defer func() { _ = serverSession.Close() }()
 
 	client := sdkmcp.NewClient(&sdkmcp.Implementation{Name: "test-client", Version: "v0"}, nil)
 	clientSession, err := client.Connect(ctx, clientTransport, nil)
 	if err != nil {
 		t.Fatalf("client connect: %v", err)
 	}
-	defer clientSession.Close()
+	defer func() { _ = clientSession.Close() }()
 
 	// The call must return a normal (in-band) error result, not a transport
 	// error or a crashed daemon.
--- a/internal/mcp/tool_panic_recovery_test.go
+++ b/internal/mcp/tool_panic_recovery_test.go
@@ -2,7 +2,12 @@ package mcp
 
 import (
 	"context"
+	"strings"
 	"testing"
+
+	"github.com/dirstral/dir2mcp/internal/config"
+	"github.com/dirstral/dir2mcp/internal/protocol"
+	sdkmcp "github.com/modelcontextprotocol/go-sdk/mcp"
 )
 
 // TestInvokeToolHandler_RecoversPanic verifies that a panicking tool handler is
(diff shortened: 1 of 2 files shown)


yidakra:
@coderabbitai review

yidakra:
@coderabbitai full review

PR #450, opened 2026-06-27
fix(cli): write Claude Desktop config atomically with 0600 perms

### LINKED ISSUE #410: cli: install claude writes the client config non-atomically (interrupt corrupts ALL the user's MCP servers) + token written 0644
**Affects the primary install path (your dad runs `dir2mcp install claude`).**

## 1 [HIGH] Non-atomic config write can corrupt the user's entire Claude Desktop config
`internal/cli/claude_cmd.go` `writeJSONFile` (~:406) does a plain truncate-in-place `os.WriteFile(path, raw, 0o644)`, used by both install (~:150) and uninstall (~:240). Install loads `claude_desktop_config.json`, merges the dir2mcp entry **alongside the user's other MCP servers**, then rewrites the whole file in place. A process kill / disk-full / write error mid-write leaves the file **truncated → the user loses every MCP server they had configured**, not just dir2mcp. The careful "preserve other servers" merge is defeated by the unsafe write.
- **Fix:** use the package's existing `writeFileAtomic` (`internal/cli/export.go:250`, sibling-temp + `os.Rename`) — the same atomic pattern already used for connection.json.

## 2 [LOW-MED] Bearer token written world-readable (0644)
`writeJSONFile` uses `0o644`, and the entry embeds `Authorization:Bearer <token>` (~:329). The source `secret.token` is `0600`, but install copies the token into `claude_desktop_config.json` at `0644` → group/other-readable on a multi-user host (secret-exposure downgrade). Write the config `0600`.

## 3 [LOW] `print-config claude` emits `bunx` while `install` resolves an npx fallback
`buildClaudeMCPRemoteEntry` hardcodes `Command:"bunx"` (~:334); `runClaudeInstall` overrides with a resolved path incl. npx fallback (~:119), but `runClaudePrintConfig` prints the raw `bunx` (~:63). A user with only `npx` gets a working `install` but a copy-paste snippet that won't run. Also the npx args omit `-y` (~:335), so an uncached package can stall on npm's confirm prompt in Claude Desktop's non-TTY child. Make the two surfaces agree and add `-y`.

yidakra:
## Problem (#410)

`install claude` / `uninstall claude` rewrite the user's entire `claude_desktop_config.json` — which holds **all** of their MCP servers, not just dir2mcp's. The rewrite went through `writeJSONFile`, which did a non-atomic `os.WriteFile(path, raw, 0o644)`:

- **Corruption risk:** truncate-in-place. An interrupted write (crash, Ctrl-C, full disk) corrupts every MCP server entry in the file, not just ours.
- **Token leak:** the config embeds the `Authorization: Bearer <token>` header, but `0644` left it world-readable.

## Fix

- Route `writeJSONFile` through the package's existing `writeFileAtomic` (sibling temp file + `fsync` + `rename`), so a reader/crash never observes a partial file and an interrupt can't truncate the existing one.
- Pin the temp file to `0600` explicitly (rather than relying on the implicit `os.CreateTemp` mode) so the token-bearing config never lands world-readable. This also applies to the export callers, which were already effectively `0600`.

## Test

Added `TestClaudeInstallWritesConfigAtomically0600AndPreservesOtherServers` in `tests/cli`, asserting the install path:
- writes the config `0600`,
- leaves no leftover temp file in the config dir on success,
- preserves an unrelated `mcpServers` entry from another tool.

## Verification

- `gofmt` clean, `go build ./...`, `go vet ./internal/cli/... ./tests/cli/...`
- `go test ./internal/cli/... ./tests/cli/...` green
- `golangci-lint run` on changed packages: 0 issues
- no functions over gocyclo 15

Scope note: this PR addresses the atomic-write + perms parts of #410. The `bunx`/`npx` and missing `-y` divergence (item 3) is intentionally deferred to keep this change focused.

Closes #410.

### DIFF
--- a/internal/cli/claude_cmd.go
+++ b/internal/cli/claude_cmd.go
@@ -403,13 +403,18 @@ func loadJSONFileOrEmpty(path string) (map[string]interface{}, error) {
 	return out, nil
 }
 
+// writeJSONFile serializes payload and writes it atomically with owner-only
+// (0600) permissions. Atomicity matters because this rewrites the user's whole
+// Claude Desktop config (containing *all* of their MCP servers): a truncate-in-
+// place write interrupted mid-stream would corrupt every entry, not just ours.
+// The 0600 mode keeps the embedded bearer token out of world-readable files.
 func writeJSONFile(path string, payload map[string]interface{}) error {
 	raw, err := json.MarshalIndent(payload, "", "  ")
 	if err != nil {
 		return err
 	}
 	raw = append(raw, '\n')
-	return os.WriteFile(path, raw, 0o644)
+	return writeFileAtomic(path, raw)
 }
 
 func jsonMarshalMap(v interface{}) (map[string]interface{}, error) {
--- a/internal/cli/export.go
+++ b/internal/cli/export.go
@@ -245,7 +245,8 @@ func transcriptRepLanguage(metaJSON string) string {
 }
 
 // writeFileAtomic writes data to path via a sibling temp file + rename so a
-// reader never observes a partially written subtitle document. The temp file is
+// reader never observes a partially written file and an interrupted write can
+// never truncate/corrupt an existing one. The temp file is created 0600 and
 // removed on any failure before the rename completes.
 func writeFileAtomic(path string, data []byte) error {
 	dir := filepath.Dir(path)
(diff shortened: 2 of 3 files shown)


yidakra:
@coderabbitai review

yidakra:
@coderabbitai full review

PR #453, opened 2026-06-27
fix: retry transient embed/index errors (#412) and defer empty embed model to the provider adapter (#396)

### LINKED ISSUE #396: embeddings: swapping to OpenAI/Cohere/openrouter/local silently fails the whole corpus (no default embed model → mistral-embed sent to wrong provider)
**Part of #395.** Directly breaks the "swap providers by supplying a credential" promise — and fails in the silent corpus-wide way #395/#364 are meant to eliminate.

## Verified bug
- `EmbeddingWorker.modelForKind` (`internal/index/embedding_worker.go:919`) falls back to the **hardcoded literals** `"mistral-embed"` (text) / `"codestral-embed"` (code) whenever `ModelForText`/`ModelForCode` is empty.
- The built-in provider profiles for **`openai`, `cohere`, `openrouter`, `local`** set **no** `embed_text_model`/`embed_code_model` (`internal/config/providers.go:68,73,69,75`) — only `mistral` (`:65-66`) and `gemini` (`:71-72`) pin theirs.
- Embed-provider precedence is `mistral, mistral-ocr, openai, gemini, cohere` (`providers.go:111`).

So a user who sets only `OPENAI_API_KEY` (or `COHERE_API_KEY`) selects that provider for embeddings with an **empty model**, and the worker sends the string **`mistral-embed`/`codestral-embed` to OpenAI/Cohere** → 400 "model not found" → non-retryable → chunks marked failed corpus-wide (`embedding_worker.go:705-723`).

The query side mirrors it: `retrieval.NewService` defaults the query model to `mistral-embed`/`codestral-embed` and `SetQueryEmbeddingModel("")` is a no-op on empty (`internal/retrieval/service.go:245-246,640-647`), so query embedding also goes to the wrong provider.

Preflight does not catch it: `checkMistralAPIKey` (`internal/cli/up.go:432-452`) only resolves the profile and builds the adapter; it never makes a probe embed call, and `providerfactory.Embedder` succeeds regardless of model name.

## Symptom
`status` shows `errors>0`, `embedded_ok=0`; `ask`/`search` return nothing — the #364 class — for a config the capability-driven-activation promise says should work out of the box. Only Mistral and Gemini work today.

## Fix direction
- `modelForKind` should fall back to `""` (let the provider adapter apply its own correct default) rather than to Mistral literals; and/or the built-in `openai`/`cohere` profiles should carry their correct default embed models.
- Query path should symmetrically fall back to the provider default, not `mistral-embed`.
- Consider a one-shot probe embed in preflight so a wrong/oversized/invalid model surfaces as a clear CONFIG error at startup (see sibling issue on embedding graceful degradation).

### LINKED ISSUE #412: embeddings: transient provider/network errors are marked permanently failed and never re-queued (connection refused / 503 / overloaded / EOF / vector-upsert)
**Part of #395.** A brief provider/network blip during indexing → permanent silent recall loss.

## 1 [HIGH] `isTransientEmbedError` misses common transient faults → they're marked permanent
`internal/index/embedding_worker.go` `isTransientEmbedError` (~:878) only treats as transient: `net.Error` timeout/temporary, `context.Canceled/DeadlineExceeded`, and substrings `"rate limit"`/`"timeout"`. It does **not** match `"connection refused"`, `"connection reset"`, `"no such host"`, HTTP `503`/`529`, `"service unavailable"`, `"overloaded"`, or bare `"EOF"`. So a chunk failing on any of those is marked `embedding_status='error'` at `:715`. Error chunks are **never re-queued** (`NextPending` selects `'pending'` only, `sqlite_store.go:1820`) → permanent until a manual full reindex.
- Direct contradiction: `errcat.go:76` *does* classify `connection refused`/`reset`/`no such host` as `transient_net`, so the category label says "transient/retryable" while the behavior is "permanent."

## 2 [HIGH] Vector-index upsert errors have no transience guard at all
`embedding_worker.go` `indexChunks` (~:605): any `w.Index.Upsert` error → `MarkFailedWithCategory` immediately, with **no `isTransientEmbedError` check** (unlike the embed path). A transient Qdrant/pgvector outage (connection refused, 503) permanently marks chunks `error` — and the embedding already succeeded, so the vector is discarded.

## Symptom
Provider/index blip during indexing → `embedded_ok` stays low forever, `FailureSummary` shows `transient_net`/`unknown`, and rerunning `up` never recovers them. The smoke gate's `embedded_ok>0` passes while real content is missing.

## Fix
- Complete `isTransientEmbedError` (reuse the `errcat.go` transient_net rules: refused/reset/no-such-host/503/529/unavailable/overloaded/EOF).
- Add the same transience guard to the vector-upsert path.
- Add a bounded re-queue/backoff for `error` chunks classified transient (or a `dir2mcp retry-failed` path), so a recovered provider self-heals without a full reindex.

yidakra:
## Summary

Fixes two silent embed-failure bugs that both flow through `internal/index/embedding_worker.go`, so they are fixed together.

### #412 — transient embed/index errors were marked permanently failed (never retried)

- **`isTransientEmbedError`** only matched `rate limit` / `timeout` / `net.Timeout`. It now also recognizes `connection refused`, `connection reset`, `no such host`, `503`, `529`, `service unavailable`, `overloaded`, and `EOF` (including a wrapped `io.EOF` via `errors.Is`). The transient-network keyword set is centralized in `internal/store/errcat.go` as `transientNetKeywords` and exposed through a new `store.IsTransientError`, so the worker's retry gate and the `transient_net` diagnostics label (`dir2mcp status`/`doctor`) classify exactly the same errors. A transient match leaves the chunk **PENDING** for retry (no `MarkFailed`), matching existing embed-path behavior.
- **The vector-index upsert path (`indexChunks`)** called `MarkFailedWithCategory` on *any* `Upsert` error with no transience check, so a brief Qdrant/pgvector blip could bury chunks in the error state corpus-wide. It now applies the same `isTransientEmbedError` guard and leaves transient upsert failures pending. Non-transient upsert errors still mark the chunk failed (regression-guarded).

### #396 — swapping the embed provider to OpenAI/Cohere silently failed the corpus

- **`modelForKind`** fell back to the hardcoded `mistral-embed` / `codestral-embed` ids when the model was empty — sending Mistral model names to whatever provider was actually active. It now returns `""` when unset so the provider **adapter** applies its own correct default. Verified all four adapters already default sanely on empty (`openai` → `text-embedding-3-small`, `cohere` → `embed-v4.0`, `gemini` → `gemini-embedding-001`, `omniembed` → its `DefaultModel`), so no profile changes were needed.
- **Query side:** `SetQueryEmbeddingModel` / `SetCodeEmbeddingModel` were no-ops on an empty model, so the query stayed pinned to the `mistral-embed`/`codestral-embed` default in `NewService` even when the corpus was embedded with a different provider. They now record the trimmed value (including empty), clearing the Mistral-era default so the active adapter resolves the same default the corpus was embedded with. The built-in Mistral profile always supplies a concrete model, so Mistral corpora are unaffected; the `NewService` defaults are retained for the existing retrieval tests.

### Tests
- `store.IsTransientError` + `ClassifyError` now cover `503` / `overloaded` (529) / `EOF` / connection-refused/reset and still reject permanent (auth/dimension) failures.
- Worker: transient embed errors (`connection refused`, `503`, `overloaded`, `unexpected EOF`, wrapped `io.EOF`) leave the chunk pending; a transient index `Upsert` error leaves it pending while a permanent one still marks it failed; `modelForKind` defers to the adapter (empty model) for both text and code axes.

### Validation
- `gofmt` clean; `go build ./...`; `go vet` clean; targeted suites (`internal/index`, `internal/store`, `internal/retrieval`, `tests/index|store|retrieval|config|provider|providerfactory|embedqueue`) all pass. The only `go test ./...` failure is the pre-existing `tests/security` case that requires the `dirstral-spec` submodule (not initialized in this worktree) — unrelated to this change. Modified functions remain under the gocyclo-15 budget.

Closes #412.
Closes #396.

### DIFF
--- a/internal/index/embedding_worker.go
+++ b/internal/index/embedding_worker.go
@@ -610,6 +610,15 @@ func (w *EmbeddingWorker) indexChunks(ctx context.Context, validTasks []model.Ch
 					w.logf("mark embedded warning: failed to mark %d chunks as embedded before index error: %v labels=%v", idx, err, labels[:idx])
 				}
 			}
+			// A transient index/upsert error (the vector store is momentarily
+			// unreachable — connection refused/reset, 503/529, EOF) must NOT
+			// permanently fail the chunk: leave it PENDING so the next cycle
+			// retries it, mirroring the embed-path behavior (issue #412).
+			// Without this guard a brief Qdrant/pgvector blip silently buried
+			// chunks in the error state corpus-wide.
+			if isTransientEmbedError(addErr) {
+				return idx, addErr
+			}
 			category := string(store.ClassifyError(addErr))
 			reason := store.SanitizeReason(addErr.Error())
 			if mfErr := w.Source.MarkFailedWithCategory(ctx, labels[idx:idx+1], category, reason); mfErr != nil {
@@ -891,6 +900,14 @@ func isTransientEmbedError(err error) bool {
 	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 		return true
 	}
+	// A bare EOF / unexpected-EOF (connection dropped mid-response) is a
+	// transient transport failure that a later cycle recovers from. Match it
+	// structurally (errors.Is) in addition to the keyword set below, since
+	// io.EOF survives wrapping (e.g. *model.ProviderError.Cause) where the
+	// string form may not.
+	if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
+		return true
+	}
 	// net.Error can indicate timeouts or temporary network failures.
 	var ne net.Error
 	if errors.As(err, &ne) {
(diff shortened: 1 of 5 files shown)


yidakra:
@coderabbitai review

yidakra:
@coderabbitai full review

PR #454, opened 2026-06-27
fix(ingest): stop docling subprocess wedging indexing (#406)

### LINKED ISSUE #406: ingest: docling/ffmpeg subprocess can wedge indexing — stderr-cap deadlock + no per-operation timeout (pending never drains)
**Part of #395.** External-tool subprocesses can hang the ingest worker indefinitely — and one path is reachable on the large legal PDFs that motivated #381.

## 1 [HIGH] docling stderr-cap deadlock
`internal/ingest/docling_extractor.go`: `limitedBuffer.Write` returns an **error** (`errDoclingOutputTooLarge`) once the buffer exceeds its limit (line ~86):
```go
func (w *limitedBuffer) Write(p []byte) (int, error) {
    if w.limit > 0 && w.buf.Len()+len(p) > w.limit { return 0, errDoclingOutputTooLarge }
    return w.buf.Write(p)
}
```
It's wired as `cmd.Stderr` with `maxDoclingStderrBytes = 1 MiB` (line ~184). When this `Write` returns an error, Go's `os/exec` stops draining the stderr pipe; docling then blocks on a full 64 KB OS pipe buffer and never exits; `cmd.Run()`'s `Wait()` blocks on process exit forever (no `cmd.WaitDelay` is set). docling emits tqdm/loguru progress chatter to stderr and the default command has no quiet flag, so a **large multi-hundred-page PDF (the #381 BVI-acts class) can exceed 1 MiB of stderr and deadlock**.
- Secondary: on the legacy stdout path (line ~218), output > 1 GiB is *meant* to be rejected cleanly as `errDoclingOutputTooLarge` (line ~221), but it deadlocks the same way instead — the child can't exit while blocked on the undrained stdout pipe, so the graceful rejection never runs.
- Note: the default file-output path's JSON size guard in `readDoclingOutputDir` (stats before read) IS safe; the deadlock is specifically the stderr capture.
- **Fix:** after hitting the cap, keep draining and discard (don't return an error from `Write`); and/or set `cmd.WaitDelay`.

## 2 [MEDIUM] No per-document timeout on docling extraction
`runFileOutput`/`runStdout` (`docling_extractor.go:168-236`) use `exec.CommandContext(ctx, …)` with the **long-lived ingest ctx** — no `context.WithTimeout` and no `WaitDelay` (the `--version` probe is bounded at 90s, but extraction is not). A hung docling (wedged venv, blocked weights download) stalls indexing until daemon shutdown → `embedded_pending` never reaches 0.

## 3 [MEDIUM] avutil ffprobe/ffmpeg calls lack the hang-guard applied to silencedetect
`internal/avutil/avutil.go`: `Duration` (~:40), `ProbeMediaInfo` (~:124), and `extractSegment` (~:447) pass the caller's ctx straight through with no internal cap, while `DetectLeadingSilence` explicitly wraps a 60s `silenceProbeTimeout` "so a pathological input cannot hang ingest." A malformed/hostile media file that makes ffprobe/ffmpeg spin blocks the worker indefinitely. **Fix:** apply the same bounded-timeout pattern to all three.

## Why it matters
All three turn one pathological/large input into a stuck indexer — the "pending never drains" class the release-smoke gate exists to catch, but here it hangs rather than errors so even the gate would just time out. Add a per-document extraction timeout (config-tunable) as a blanket backstop, plus the stderr-drain fix.

yidakra:
## Problem

The docling extractor (`internal/ingest/docling_extractor.go`) could **wedge the entire indexer** on large documents — exactly the large legal PDFs this corpus is built around. Two independent failure modes:

### 1. stderr-cap deadlock
`limitedBuffer.Write` returned an error (`errDoclingOutputTooLarge`) once the 1 MiB stderr cap was exceeded. `limitedBuffer` is wired as `cmd.Stderr`, and when a `Write` returns an error, Go's `os/exec` **stops draining that pipe**. docling then blocks writing to a full pipe and `cmd.Run()` never returns. Since docling streams verbose `tqdm`/`loguru` progress to stderr, a large PDF can easily exceed 1 MiB of stderr and hang indexing forever.

### 2. no per-document timeout
`runFileOutput`/`runStdout` ran `exec.CommandContext` with the long-lived ingest context — no per-extraction bound. A single hung/slow child blocked indexing indefinitely.

## Fix

1. **`limitedBuffer.Write` never errors.** It keeps accepting every write, discarding bytes once over the cap but always returning `(len(p), nil)` so the pipe keeps draining. Overflow is tracked via a new `Truncated()` method. The legacy stdout over-cap rejection (`runStdout`) now checks `Truncated()` **after** `cmd.Run()` returns instead of erroring mid-`Write`. The separate file-output size guard in `readDoclingOutputDir` is unchanged.
2. **Per-document timeout.** Both run paths wrap the context with `context.WithTimeout(ctx, doclingExtractTimeout)` where `doclingExtractTimeout = 15m` — generous headroom above the slowest real legal PDF on CPU-only hosts, while still guaranteeing forward progress. `cmd.WaitDelay = 10s` is a kill backstop so a wedged child that ignores cancellation can't block `Run()` indefinitely. On expiry a clear `"docling command timed out"` error marks the doc failed (retryable) rather than hanging the indexer.

## Tests

New `internal/ingest/docling_extractor_internal_test.go` (white-box, `limitedBuffer` is unexported):
- `TestLimitedBuffer_OverCapNeverErrors` — over-cap writes return no error, report full length, retain only up to the cap, and flag truncation (the deadlock guard).
- `TestLimitedBuffer_ZeroLimitUnbounded` — non-positive limit disables capping.
- `TestRunFileOutput_TimesOutOnHungCommand` — a fake docling that floods stderr then hangs is killed promptly and surfaces a timeout error (exercises both fixes together).

## Validation
- `gofmt`, `go build ./...`, `go vet ./...`, `make cyclo` (nothing over 15) — clean.
- `go test ./internal/ingest/... ./tests/ingest/...` — pass.
- `make ci` — only failure is the pre-existing, unrelated `tests/security` spec-pattern test that needs the `dirstral-spec` submodule checked out (not initialized in this worktree).

Closes #406.

### DIFF
--- a/internal/ingest/docling_extractor.go
+++ b/internal/ingest/docling_extractor.go
@@ -3,12 +3,12 @@ package ingest
 import (
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/dirstral/dir2mcp/internal/ingest/docling"
 )
@@ -44,7 +44,22 @@ const (
 	maxDoclingStderrBytes = 1 * 1024 * 1024
 )
 
-var errDoclingOutputTooLarge = errors.New("docling output exceeded configured limit")
+// doclingExtractTimeout bounds a single document's docling subprocess so one
+// pathological file can never wedge the whole indexer (issue #406). docling
+// streams tqdm/loguru progress for the lifetime of a slow extraction, so the
+// limit must sit comfortably above the slowest real document: large multi-
+// hundred-page legal PDFs (the corpus this protects) can take several minutes
+// on CPU-only hosts, so 15 minutes leaves generous headroom while still
+// guaranteeing forward progress. On expiry the document is marked failed
+// (retryable) instead of hanging indexing. It is a var (not a const) only so
+// tests can shorten it; production never reassigns it.
+var doclingExtractTimeout = 15 * time.Minute
+
+// doclingWaitDelay is a backstop after the context is cancelled (timeout or
+// shutdown): os/exec sends the kill signal, and if the child has not exited
+// within this grace period the I/O pipes are force-closed so cmd.Run() cannot
+// block indefinitely on a wedged subprocess that ignores cancellation.
+const doclingWaitDelay = 10 * time.Second
 
 // SanitizeDoclingEnv returns env (in os.Environ "KEY=VALUE" form) with the
 // Python path-injection variables removed and user site-packages disabled, so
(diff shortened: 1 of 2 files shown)


yidakra:
@coderabbitai review

yidakra:
@coderabbitai full review

PR #456, opened 2026-06-27
store: enforce document status invariants (#425)

### LINKED ISSUE #425: store: document status invariants violated — secret_excluded silently persisted as ok; a status=error doc keeps live searchable chunks
**Part of #395.** Two verified document-state invariant violations.

## 1 [HIGH] `secret_excluded` is silently rewritten to `ok` on persist (audit loss + inflated indexed count)
`internal/store/sqlite_store.go` `normalizeStatus` recognizes only `skipped`/`error`; everything else → `ok`:
```go
func normalizeStatus(status string) string {
    switch strings.ToLower(strings.TrimSpace(status)) {
    case "skipped": return "skipped"
    case "error":   return "error"
    default:        return "ok"
    }
}
```
Ingest sets `doc.Status = "secret_excluded"` (`internal/ingest/service.go:1508` archive members, `:1588` main path) for documents withheld because their content matches a secret pattern. `UpsertDocument` (`sqlite_store.go:657`) runs `normalizeStatus` → the persisted row becomes **`status='ok'` with zero representations/chunks** — indistinguishable from a healthy indexed document. `CorpusStats` counts `status='ok'` as **indexed** (`sqlite_store.go:1778`), so secret-excluded docs inflate the `indexed` total instead of `skipped`, and the security-audit signal (which documents were withheld for secrets) is **permanently lost**. (`documents.status` has no CHECK constraint, so the column could hold the value — only `normalizeStatus` downgrades it.)
**Fix:** add `secret_excluded` (and `pending`) as recognized cases in `normalizeStatus`; count it as skipped, not indexed; surface it in the coverage report (#414).

## 2 [HIGH] A `status=error` document keeps live, embeddable, searchable chunks
`generateRepresentations` writes each representation in its **own** transaction (`internal/ingest/represent.go:218` per-rep `WithTx`) — there is no outer transaction across reps. On a mid-way failure, `processDocument` sets `doc.Status="error"` and upserts (`service.go:1282-1290`) but does **not** roll back / tombstone the reps+chunks earlier reps already committed. And `NextPending` (`sqlite_store.go:1820`) selects pending chunks with **no parent-document status check** (verified: `FROM chunks c WHERE c.embedding_status='pending' AND c.deleted=0 AND c.chunk_id>0` — no join to `documents`).
**Trigger:** a doc where rep A (raw_text/OCR) succeeds and a later rep (media/translation/structured) fails.
**Symptom:** rep A's chunks get embedded → vectors + `chunkByLabel` → `search`/`ask` return hits from a document that `status`/doctor report as **failed**; and every subsequent incremental run sees `status==error` → `resolveNeedsProcessing` true → re-upserts those chunks to `pending` and **re-embeds them every cycle**, while the doc stays permanently `error`.
**Fix:** wrap doc+reps in one atomic unit (or tombstone partial reps on failure); have `NextPending` exclude chunks whose parent doc is `error`/`deleted`.

yidakra:
## Summary

Fixes two document status-invariant violations in `internal/store/sqlite_store.go` that let withheld/failed documents masquerade as healthy indexed content.

### 1. `normalizeStatus` collapsed unknown statuses to `ok`
It only recognized `skipped`/`error`; everything else (including `secret_excluded`, set by ingest when a file is withheld for containing secrets) was rewritten to `ok`. The doc was then persisted as an indexed `ok` document with zero chunks — counted as indexed, audit signal lost.

- Recognize `secret_excluded` and `pending` so they round-trip faithfully.
- Count `secret_excluded` as **skipped** (not indexed) in `CorpusStats` (`status IN ('skipped','secret_excluded')`).

### 2. `NextPending` ignored the parent document's lifecycle
It selected pending chunks with no parent-document check, so a document that later went `status='error'` (a representation failed *after* an earlier rep's chunks committed) kept **live embeddable/searchable chunks** that the embed worker re-embedded every cycle.

- `LEFT JOIN documents ON rel_path` and skip chunks whose parent is `status='error'` or tombstoned (`deleted=1`).
- A `d.rel_path IS NULL` guard preserves bare chunks with no document row (e.g. `UpsertChunkTask`-seeded chunks), so existing behavior is not regressed.

### Tests
New `tests/store/document_status_invariants_test.go`:
- (a) a doc upserted with `status='secret_excluded'` round-trips as `secret_excluded` (not `ok`) and is counted as skipped, not indexed.
- (b) chunks of a `status='error'` document (and a tombstoned document) are NOT returned by `NextPending`, while an `ok` document's chunk IS.

Full `make check` passes (gofmt, vet, golangci-lint, gocyclo, ineffassign, misspell, tests, build). Only `internal/store/sqlite_store.go` and the new test file changed.

Closes #425.

### DIFF
--- a/internal/store/sqlite_store.go
+++ b/internal/store/sqlite_store.go
@@ -1776,7 +1776,7 @@ func (s *SQLiteStore) CorpusStats(ctx context.Context) (model.CorpusStats, error
 		SELECT
 			COUNT(*) AS scanned,
 			COALESCE(SUM(CASE WHEN deleted = 0 AND status = 'ok' THEN 1 ELSE 0 END), 0) AS indexed,
-			COALESCE(SUM(CASE WHEN deleted = 0 AND status = 'skipped' THEN 1 ELSE 0 END), 0) AS skipped,
+			COALESCE(SUM(CASE WHEN deleted = 0 AND status IN ('skipped', 'secret_excluded') THEN 1 ELSE 0 END), 0) AS skipped,
 			COALESCE(SUM(CASE WHEN deleted = 1 THEN 1 ELSE 0 END), 0) AS deleted,
 			COALESCE(SUM(CASE WHEN deleted = 0 AND status = 'error' THEN 1 ELSE 0 END), 0) AS errors
 		FROM documents`,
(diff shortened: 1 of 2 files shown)


yidakra:
@coderabbitai full review

Issue #418, opened 2026-06-25
cli: lifecycle robustness — reindex has no live-daemon guard (corruption) + destructive-first ordering, pid-reuse in down/up, status reports stale running


yidakra:
**Part of #395.** CLI lifecycle correctness gaps.

## 1 [MED-HIGH] `reindex` has no running-daemon guard → two writers + index files unlinked under the live daemon
`internal/cli/reindex.go:24-59`; `prepareReindexStore:83-112` opens the sqlite store and `os.Remove`s the HNSW/index files (`:102-109`) without reading `pidFilePath(cfg.StateDir)` first (contrast `up_daemon.go:234`). If a `dir2mcp up` daemon is live for the same `.dir2mcp`, `dir2mcp reindex` runs concurrent sqlite writers and deletes vector-index files the daemon still holds open → index/state corruption; the daemon keeps serving a freed/stale index. **Fix:** detect the live pid and refuse (or signal the daemon to reindex), like `up`.

## 2 [MED] `reindex` clears destructively up front → interrupt leaves an empty corpus
`prepareReindexStore` wipes content hashes (`:94`) and removes index files (`:102`) *before* `ing.Reindex(ctx)` runs (`:55`); on SIGINT/error `finishReindex` returns generic failure with no rollback. Ctrl-C on a long OCR/embed reindex → hashes wiped + prior index deleted but rebuild incomplete → corpus left empty/half-indexed. **Fix:** build into a staging location and swap on success, or at least don't delete the old index until the rebuild is durable.

## 3 [MED] pid-reuse: `down` can SIGKILL an unrelated process; `up` refuses to start
`down.go:54-63` / `daemon.go:336-351` use a bare signal-0 existence check on the integer pid (`claimPIDFile` stores only the pid). After a crash without cleanup, the OS can recycle the pid → `dir2mcp down` kills that unrelated process; `dir2mcp up` refuses with "already running (pid N)". **Fix:** record start-time/cmdline in the pid file and verify identity before signaling.

## 4 [MED] `status` reports stale "running" after a daemon crash
`status.go:88-92` reads `snapshot.Indexing.Running` straight from `corpus.json` (daemon-written) with no liveness cross-check against pid/connection.json/endpoint. A daemon that wrote `Running=true` then crashed shows `Indexing … running` (green) forever. **Fix:** reconcile against pid liveness.

[...]

PR #457, opened 2026-06-27
cli: refuse reindex under a live daemon (#418)

yidakra:
## Summary

`internal/cli/reindex.go` `prepareReindexStore` opens the sqlite store and `os.Remove`s the HNSW/index files **without first checking whether a `dir2mcp up` daemon is live for the same state dir**. If a daemon is running, `dir2mcp reindex` runs concurrent sqlite writers AND unlinks index files the daemon still holds open → index/state corruption, with the daemon continuing to serve a freed, stale index.

This adds a guard at the very start of the reindex flow — **before** the confirmation prompt and any destructive operation — that detects a live daemon and refuses with a clear, actionable error (non-zero exit) telling the user to stop the daemon first with `dir2mcp down`.

## Implementation

- New `refuseReindexIfDaemonRunning` reuses the **exact** liveness mechanism `up`/`up_daemon.go` use — `readPIDFile(pidFilePath(stateDir))` + `processIsAlive` — rather than reimplementing any pid logic.
- A missing pid file (common case), a malformed pid file, or a stale pid file naming a dead process all let reindex proceed exactly as before (no false positives on a crashed daemon).
- Config load was hoisted above the confirmation prompt so the guard can refuse before asking the user to confirm a destructive action.

## Scope

Partially addresses #418 (reindex live-daemon guard). The other #418 items live in other files and are intentionally left for separate PRs:
- destructive-first ordering / staging-and-swap (`reindex.go` part 2) — deferred
- pid-reuse identity in `down`/`up` (`down.go`/`daemon.go`)
- stale `status` "running" reconciliation (`status.go`)

## Tests

`tests/cli/reindex_daemon_guard_test.go`:
- `TestReindex_RefusesWhenDaemonAlive` — live pid → non-zero exit, error mentions `dir2mcp down`, ingestor never constructed, and the seeded index file is **not** deleted.
- `TestReindex_ProceedsWithStalePidFile` — stale (dead) pid → reindex proceeds.

Full `make check` (incl. golangci-lint + gocyclo) passes locally.

### DIFF
--- a/internal/cli/reindex.go
+++ b/internal/cli/reindex.go
@@ -27,15 +27,21 @@ func (a *App) runReindex(ctx context.Context, global globalOptions, args []strin
 		return exitConfigInvalid
 	}
 
+	cfg, code := a.loadReindexConfig(global)
+	if code != exitSuccess {
+		return code
+	}
+	// Refuse before prompting/destroying anything: a reindex under a live
+	// daemon corrupts the shared index/state (issue #418).
+	if code := a.refuseReindexIfDaemonRunning(global, cfg); code != exitSuccess {
+		return code
+	}
+
 	if !a.confirmDestructive(global, "Re-index all documents?", "Discards the current index and rebuilds it from scratch (may re-run OCR/embeddings).") {
 		writeln(a.stderr, "reindex aborted")
 		return exitSuccess
 	}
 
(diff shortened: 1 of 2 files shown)


yidakra:
@coderabbitai full review

Issue #436, opened 2026-06-26
diagnostics: support-bundle ships corpus content/query text un-redacted + unbounded; query_metrics NDJSON emitter unsynchronized (log corruption); status snapshot stale running=true survives crash


yidakra:
**Part of #395.** Integrity/privacy gaps in the diagnostic surfaces (distinct from the filed secret-redaction + status-stale issues).

## Support bundle (`internal/cli/support_bundle.go`)
- **S1 [MED] corpus content / paths / query text shipped without redaction or consent.** `list-files.json` embeds every doc's `rel_path`, `Title`, and `error_message` (`:264-285`) → discloses the full corpus inventory + content fragments. `server.log` is added **raw** (`:127-128`, verified — `redactBundleSecrets` is applied to client logs `:404` and daemon.json `:439` but NOT server.log) and can contain query text, answer snippets, and absolute paths. No prompt/flag warns the user the bundle contains corpus content before they send it to a maintainer. **Fix:** redact server.log like client logs; gate corpus content behind a `--include-content` flag with a warning.
- **S2 [LOW-MED] partially unbounded.** `server.log` is tail-capped at 2 MB (good), but `list-files.json` uses `ListFiles(...,0,0)` (limit 0 = ALL docs, `:264`) and `collectClientMCPLogs` globs `mcp*.log` with no file-count cap (`:392`) → a large corpus / many rotated Claude logs produce a giant bundle.
- **S3 [LOW-MED] bundle status trusts a possibly-stale corpus.json** (`:199-221`, no staleness/liveness check) → misreports indexing state to the maintainer diagnosing it. **S4 [LOW]** opening `meta.sqlite` (with `Init` migrations) while the daemon is live → SQLITE_BUSY risk (degrades to a warning, not abort — acceptable).

## Query metrics
- **M1 [LOW-MED] the NDJSON emitter is unsynchronized (verified).** `QueryMetrics` accumulation is race-free (per-query + mutex'd Sink), but `ndjsonEmitter.Emit` writes via `fmt.Fprintln(e.out, …)` with **no lock** (`app.go:~2067`), and that one emitter is the shared sink for concurrent ask/search `query_metrics` events + the corpus writer. query_metrics lines exceed PIPE_BUF, so concurrent writes to server.log **interleave/corrupt**. **Fix:** lock the emitter write.
[...]

PR #458, opened 2026-06-27
fix(cli): synchronize ndjson emitter + atomic connection.json write (#436)

yidakra:
Addresses the emitter-race + connection.json-atomicity parts of #436 (bundle-redaction + snapshot-staleness deferred).

### FIX 1 — unsynchronized NDJSON emitter (M1)
`ndjsonEmitter.Emit` wrote to the shared `e.out` (server.log) with a bare `fmt.Fprintln` and no lock, while a single emitter is the shared sink for concurrent ask/search `query_metrics` events *and* the corpus writer. Those lines can exceed `PIPE_BUF`, so concurrent unsynchronized writes could interleave and corrupt lines. Added a `sync.Mutex` to the struct and lock around the only write site.

### FIX 2 — connection.json non-atomic write (D3)
`writeConnectionFile` used `os.WriteFile` (`O_CREATE|O_TRUNC`), so a concurrent reader (readiness poll, support bundle, `claude_cmd`) could observe a truncated/partial file. It now writes atomically via a sibling temp file + `fsync` + rename, preserving the existing `0o600` mode. Factored the temp+sync+rename logic into a shared `atomicWriteFile` helper and routed `writeCorpusSnapshot` through it too (de-duplicating its previously-inline implementation).

## Tests (internal/cli/corpus_writer_test.go)
- `TestNDJSONEmitter_ConcurrentEmitNoInterleave`: 32 goroutines × 50 emits of >PIPE_BUF payloads through one emitter; asserts every output line is individually-valid JSON and the exact expected count (no interleaving). Passes under `-race`.
- `TestWriteConnectionFile_AtomicNoTempLeftover`: success leaves a complete valid file at mode `0o600` and no `.tmp.` leftovers.
- `TestWriteConnectionFile_ConcurrentReadNeverPartial`: a reader hammering the file during 200 rewrites only ever sees a complete, valid JSON document (or no file).

New tests were placed in the already-allowlisted `corpus_writer_test.go` to keep the `internal/cli` file-ownership boundary test green without touching `tests/security`.

`make check` passes locally (vet, golangci-lint, gocyclo -over 15, misspell, full test suite incl. `-race`).

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -16,6 +16,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -245,7 +246,12 @@ type ndjsonEvent struct {
 
 type ndjsonEmitter struct {
 	enabled bool
-	out     io.Writer
+	// mu serializes writes to out. A single emitter is the shared sink for
+	// concurrent ask/search query_metrics events and the corpus writer;
+	// without synchronization their lines (which can exceed PIPE_BUF) may
+	// interleave and corrupt server.log.
+	mu  sync.Mutex
+	out io.Writer
 }
 
 type cliErrorPayload struct {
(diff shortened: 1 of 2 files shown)


yidakra:
@coderabbitai full review

Issue #442, opened 2026-06-26
extraction: ExtractTitle picks page-headers/shebang/code-fence lines and ignores YAML front-matter; image 'annotation' is OCR-text-only (can't describe pictures) + unspanned


yidakra:
**Part of #395.** Title-extraction and image-annotation correctness (distinct from the filed BOM #417 and docling-tempfile #383 title cases).

## Title extraction (`internal/ingest/title.go`)
- **5.1 [MED] All-caps page-header / running-head becomes the title.** `titleFromUppercaseLine` (title.go:82) rejects only a trailing `.` and ignores digits, so `PAGE 1 OF 10`, `CONFIDENTIAL`, `DRAFT`, `JANUARY 2026`, `ABSTRACT:`, `BREAKING NEWS:` all become document titles.
- **5.2 [MED] YAML front-matter title is ignored.** `ExtractTitle` has no front-matter handling: `---\ntitle: The Real Title\n---\nCHAPTER ONE` → `---` skipped, the `title:` line fails the uppercase test, `CHAPTER ONE` wins. The explicitly-declared title is discarded.
- **5.3 [MED] Shebang/hashtag/code-comment lines become markdown "headings".** `titleFromMarkdownHeading` (title.go:68) accepts any `#`-prefixed line incl. no-space forms (CommonMark requires a space): `#!/usr/bin/env python` → title `!/usr/bin/env python`; `#nowplaying` → `nowplaying`; a `# comment` inside a fenced code block (no fence tracking) → becomes the title.
- **5.4 [LOW]** Setext (underline `===`) H1 not recognized. **5.5 [LOW]** a <200-char all-caps body sentence (no trailing `.`) becomes a long "title".
**Fix:** parse YAML front-matter title first; require a space after `#`; track code fences; reject digit-heavy/colon-terminated/multi-word-header lines; recognize setext.

## Image annotation (`mcp/tools.go` annotate + `ingest/service.go`)
[...]

Issue #417, opened 2026-06-25
ingest/retrieval: silent quality bugs — rerank pool truncated to k, code chunking oversize, no charset detection (UTF-16/Latin-1), text reps not language-tagged, byte-vs-rune flush (Cyrillic/CJK), BOM


yidakra:
**Part of #395.** Silent retrieval-quality degraders verified in the text/chunking/fusion paths. (Audited-sound: `chunkTextByChars` is rune-correct with floored step + complete tail; RRF/MMR math correct; cross-file dedup is exact-hash, not fuzzy.)

## 1 [MED] Rerank only ever sees the top-k candidates (over-fetch pool defeated)
`internal/retrieval/service.go:1863` (hybrid: `fuseRRF(..., k)` truncates to k inside fusion) and `:2334` (`collectFilteredCandidates` stops at `len < k`) both feed `rerankPool(..., k)` — so the reranker receives ≤ k items and can only **reorder** them, never rescue a relevant doc at fused rank k+1..50. `defaultRerankCandidatePool=50` is thereby defeated on the common (hybrid, vector-only) paths; the HyDE path correctly feeds 50 (`:994`), proving intent. **Fix:** fetch `rerankCandidatePool` candidates before rerank on all paths.

## 2 [MED-HIGH] Code chunking is line-bounded only → minified/long-line files become one oversize chunk
`represent.go:970` `chunkCodeByLines` splits purely by line count (`maxLines=200`); char/token count is never bounded. A minified `.js`/`.css`, single-line bundled JSON-in-JS, or any long-line code file → one chunk = whole file. There's **no per-chunk size guard before embedding** (only the whole-file cap), so the provider rejects it (chunk → `embedding_status=error`, silently absent) or truncates it (most of the file unsearchable). Only the `code` doctype is exposed (text/data route to rune-bounded `chunkTextByChars`). **Fix:** add a char/token cap to code chunking + a pre-embed oversize guard/quarantine.

## 3 [MED] No charset detection → UTF-16 indexed as garbage, Latin-1/Windows-1252 destroyed
[...]

PR #460, opened 2026-06-27
ingest: harden ExtractTitle against shebangs, page headers, and front-matter (#442)

yidakra:
Addresses the title-extraction part of #442 (image-annotation part deferred).

Scoped to `internal/ingest/title.go` (+ its test). Fixes the `ExtractTitle` heuristic:

- **5.3 Shebang/hashtag are not headings.** ATX headings now require a space/tab after the 1-6 `#` run (CommonMark), so `#!/usr/bin/env python` and `#nowplaying` no longer become titles.
- **5.2 YAML front-matter title is honored.** A leading `---` front-matter block is parsed and its `title:` value becomes the highest-priority source (strips one layer of matching quotes). A block with no `title:` key is skipped so its values aren't mistaken for candidates.
- **5.1 Page headers / labels rejected.** The uppercase-line fallback now rejects digit-heavy running headers (`PAGE 1 OF 10`, `JANUARY 2026`) and colon-terminated labels (`ABSTRACT:`), in addition to the existing trailing-`.` rule.
- **5.4 Setext H1** (`Title` + `=====` underline) is recognized.
- Fenced code blocks are tracked so a `#` comment inside a fence is not treated as a heading.

Conservative: genuine all-caps titles (including ones with a trailing year, e.g. `... ACT, 2021`) and real `# Heading` lines still extract exactly as before. The existing #383 (docling-tempfile) and #417 (BOM) paths are untouched.

Adds regression tests covering: shebang/hashtag not a title; front-matter title used (and quote-stripped); front-matter without title falls through; `PAGE 1 OF 10` / `JANUARY 2026` / `ABSTRACT:` not titles; normal `# Heading` and `MAIN TITLE` still work; code-fence `#` comment ignored; setext H1.

`make check` passes locally (gofmt/vet/golangci-lint/gocyclo/tests/build).

The image-annotation part of #442 (`mcp/tools.go`, `ingest/service.go`) is intentionally out of scope for this PR.

### DIFF
--- a/internal/ingest/title.go
+++ b/internal/ingest/title.go
@@ -19,10 +19,14 @@ const titleMaxLen = 200
 // beginning of a text/markdown/OCR body. Used to humanize citations whose
 // rel_path is opaque (e.g. Windows 8.3-truncated PDF filenames).
 //
-// Heuristic, in order of preference (markdown headings always win, even when
-// they appear after an earlier uppercase-line candidate):
-//  1. The first markdown heading line (`# Title`, `## Title`, ...).
-//  2. The first line that is mostly uppercase and looks title-like.
+// Heuristic, in order of preference (earlier wins):
+//  1. A `title:` key inside a leading YAML front-matter block (the author's
+//     explicitly declared title).
+//  2. The first markdown ATX heading line (`# Title`, `## Title`, ...) or a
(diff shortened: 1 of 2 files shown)


yidakra:
@coderabbitai full review

PR #461, opened 2026-06-27
ingest: persist status=error when a provider failure yields zero reps (#413)

### LINKED ISSUE #413: ingest/observability: failure that yields zero representations leaves the doc status=ok, invisible to errors/recent_failures (audio silently unsearchable)
**Part of #395.**

## 1 [HIGH] Transcript-provider failure → doc stays status=ok with zero reps
`internal/ingest/service.go:1996-2003`: on `ErrTranscriptProviderFailure`, the code logs, calls `addErrors(1)`, and `return nil` — the audio doc keeps `status="ok"` (set at build time), produces no transcript chunk, and is **never persisted as error**. Consequences:
- not in `CorpusStats.Errors` (counts `status='error'`, `sqlite_store.go:1781`)
- not in `RecentFailures` (filters `status='error'`, `:1693`)
- not in `FailureSummary` (chunk-level; no chunk exists)
- only trace: a log line + the in-memory `addErrors` counter (lost on restart)

Symptom: operator asks "why isn't my audio searchable?" — `status` shows the file under `audio` doc-count as healthy, `errors=0` after restart, nothing in failures. Undiagnosable without reading logs. Same silent-ok risk for any media that yields no representation.

## 2 [MEDIUM] Live vs persisted `errors` diverge
`addErrors` feeds the live `IndexingState` (`app.go:1834`), but the computed/persisted snapshot (`status` after restart, `corpus.json`) uses `corpusStats.Errors` from the `status='error'` query (`app.go:1838`). A transcript-failed corpus shows `errors=5` live and `errors=0` computed — contradictory, neither with a path list.

## Fix
Persist a real failure for any document that produces zero representations from a non-transient cause (`status='error'` + `error_message`, or a dedicated `skipped`+reason per the coverage-report issue), so it surfaces in `RecentFailures`/`status`. Reconcile the live and persisted error counts.

yidakra:
## Problem

On `ErrTranscriptProviderFailure`, `generateTranscriptOrSidecar` logged, called `addErrors(1)`, and `return nil` — leaving the audio document `status="ok"` with no transcript and **never persisting it as an error**. The failure was therefore invisible to:

- `CorpusStats.Errors` (counts `status='error'` rows)
- `RecentFailures`
- `FailureSummary`

So a transcription provider failure left the audio **silently unsearchable**, and the corpus reported `errors=0` after a daemon restart (the live counter is in-memory only).

## Fix

When a genuine (non-transient) provider failure leaves a document with **zero representations**, persist it as `status="error"` with a redacted `error_message` before `return nil`:

- New `persistNonFatalDocError` helper re-upserts the document as `status="error"` (logging+swallowing the upsert error, mirroring the existing `generateRepresentations` error path). The live `addErrors(1)` counter is kept.
- The persist is gated on `!mediaProduced`: a document that DID produce multimodal media chunks stays `"ok"` (still searchable — not a zero-representation failure). `secretPatterns` and `mediaProduced` are threaded through `generateRepresentations` / `generateTranscriptOrSidecar`.
- **Legitimately empty media is unaffected**: an empty transcript returns `nil` *without* the `ErrTranscriptProviderFailure` sentinel, so it never marks an error.
- The batch is **still not hard-failed** — the run continues (`return nil`), only the persisted status/visibility changes. The failed doc is now also retried on the next incremental run (`resolveNeedsProcessing` retries `status='error'`).

Scope: `internal/ingest/service.go` plus tests.

## Tests

- New `tests/ingest/zero_rep_failure_test.go`: an audio doc whose transcription fails with `ErrTranscriptProviderFailure` is persisted `status='error'` (+ `error_message`) and surfaces in `RecentFailures` and `CorpusStats.Errors`, while `ProcessDocument` does not hard-fail. (real SQLite store, end to end)
- Updated `TestServiceRun_AudioTranscriberFailure_DoesNotFailRun`: it previously asserted the document status *remained* `"ok"` — that encoded the exact #413 bug. It now asserts `status="error"` + non-empty `error_message`, keeping the "run does not fail" guarantee.

`make check` passes locally (build, vet, golangci-lint, gocyclo <15, full test suite, python tests).

Closes #413.

### DIFF
--- a/internal/ingest/service.go
+++ b/internal/ingest/service.go
@@ -1279,7 +1279,7 @@ func (s *Service) processDocument(ctx context.Context, f DiscoveredFile, secretP
 		return nil
 	}
 
-	if err := s.generateRepresentations(ctx, doc, content, forceReindex); err != nil {
+	if err := s.generateRepresentations(ctx, doc, content, secretPatterns, forceReindex); err != nil {
 		// Persist error status so the next incremental run retries this document
 		// and operators can identify it via status queries.  Silence the upsert
 		// error since the original rep error is the more actionable signal.
@@ -1526,7 +1526,7 @@ func (s *Service) processDocumentFromContent(ctx context.Context, relPath string
 	if !needsProcessing || doc.Status != "ok" {
 		return nil
 	}
-	if err := s.generateRepresentations(ctx, doc, content, forceReindex); err != nil {
+	if err := s.generateRepresentations(ctx, doc, content, secretPatterns, forceReindex); err != nil {
 		// Persist error status so the next incremental run retries this document
 		// and operators can identify it via status queries.  Silence the upsert
 		// error since the original rep error is the more actionable signal.
@@ -1934,7 +1934,7 @@ func isEmbeddableAudio(relPath string) bool {
 	}
 }
 
-func (s *Service) generateRepresentations(ctx context.Context, doc model.Document, content []byte, forceReindex bool) error {
+func (s *Service) generateRepresentations(ctx context.Context, doc model.Document, content []byte, secretPatterns []*regexp.Regexp, forceReindex bool) error {
 	if s.repGen == nil {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


yidakra:
@coderabbitai full review

Issue #433, opened 2026-06-26
embedqueue: poison task redelivered forever (reclaim ignores attempts), dead-letter never marks chunk status → infinite re-enqueue + unbounded growth, DLQ never surfaced


yidakra:
**Part of #395.** Correctness gaps in the async embed queue (`internal/embedqueue/`), distinct from #412/#399. Root: the broker is decoupled from the chunk store and **nothing writes `embedding_status` on dead-letter**; the coordinator re-derives work purely from `embedding_status='pending'`. (Duplicate *vectors* are prevented by id-keyed `Upsert` — that holds; these are about queue mechanics.)

## F1 [HIGH] Reclaim path ignores `attempts` → crash-inducing chunk redelivered forever (verified)
The `maxAttempts`→dead-letter gate exists only in `Nack` (`membroker.go:149`, `sqlitebroker.go:232`). The lease-reclaim path moves an expired in-flight job **straight back to pending with no attempts check** (verified `reclaimExpiredLocked` membroker.go:119: `if now.After(deadline){ ... b.pending = append(...) }`; sqlitebroker.go:160). A chunk whose embedding reliably kills/hangs the worker (OOM-kill, media-decoder segfault, wedged ffmpeg) dies before Ack/Nack → lease expires → reclaimed → kills the next worker → **poison-task infinite loop + pool-wide crash loop**, the exact failure the lease/DLQ mechanism is sold to bound.

## F2 [HIGH] Dead-letter never writes chunk status → coordinator re-enqueues dead jobs forever + unbounded growth
On embed failure the worker Nacks and returns; `Nack` may dead-letter internally but returns nil — the worker never marks the chunk (no `MarkFailed*` anywhere in `embedqueue`). For a transient failure the chunk stays `embedding_status='pending'`. Enqueue dedup blocks only `pending`/`inflight`, **not `dead`** (sqlitebroker.go:127). So `coordinator.go:46` re-reads the pending head every 2s → re-enqueues a fresh job → 5 attempts → dead-letter → repeat. Symptom: infinite re-enqueue burning a worker slot every cycle; SQLite `dead` rows never deleted (~6 rows/min/poison-chunk forever); MemBroker `deadLetter` slice grows unbounded (**memory leak** in default mode). A low-`chunk_id` poison chunk sits permanently at the `NextPending` head.

## F3 [MED] Dead-letter queue never surfaced
[...]

PR #462, opened 2026-06-27
embedqueue: respect maxAttempts on lease-reclaim + set broker DB pragmas (#433)

yidakra:
Addresses the reclaim-attempts (F1) + SQLiteBroker-pragmas (F5) parts of #433 (dead-letter write-back F2/F3 deferred)

## F1 [HIGH] — reclaim path ignored `attempts` → poison task redelivered forever
The `maxAttempts`→dead-letter gate existed only in `Nack`. The lease-reclaim path moved an expired in-flight job straight back to `pending` with no attempts check, so a chunk whose embedding reliably kills/hangs the worker (dying before Ack/Nack) was re-leased forever in an infinite crash loop. Reclaim now mirrors `Nack`'s gate: a lease that expires after exhausting `maxAttempts` is dead-lettered instead of redelivered. `MemBroker` (`reclaimExpiredLocked`) and `SQLiteBroker` (new shared `reclaimExpired` helper, used by both `Lease` and `Stats`) are kept consistent.

## F5 [MED] — SQLiteBroker opened its DB with no busy_timeout/WAL → SQLITE_BUSY storms
`NewSQLiteBroker` now sets `busy_timeout=5000` then `journal_mode=WAL` (same order/values as `internal/store/sqlite_store.go`), so the reclaim-write-first `Lease` waits on contention under the multi-worker pool instead of failing immediately with `database is locked`.

## Scope
Touches only `internal/embedqueue/` (membroker.go, sqlitebroker.go) + tests. The dead-letter-write-back-to-chunk-status (F2) and surface-the-DLQ (F3) parts need cross-package store interaction and are deferred to a separate PR.

## Tests
- Reclaim dead-letters a job once it reaches `maxAttempts` and its lease expires (not redelivered to pending) — both `MemBroker` and `SQLiteBroker`.
- The broker DB reports `busy_timeout=5000` and `journal_mode=wal` after construction.

`make check` passes locally (gofmt, vet, golangci-lint, gocyclo, ineffassign, misspell, tests, build).

### DIFF
--- a/internal/embedqueue/membroker.go
+++ b/internal/embedqueue/membroker.go
@@ -115,12 +115,19 @@ func (b *MemBroker) Lease(_ context.Context, visibility time.Duration) (Lease, e
 
 // reclaimExpiredLocked moves in-flight jobs whose lease deadline has passed back
 // to pending so a crashed/abandoned worker cannot strand a chunk (SPEC §8.7.3
-// lease expiry). Caller holds b.mu.
+// lease expiry). A job whose lease expires after it has already exhausted
+// maxAttempts is dead-lettered instead of redelivered, mirroring Nack's gate, so
+// a chunk that reliably kills/hangs the worker before it can Ack/Nack cannot be
+// re-leased forever. Caller holds b.mu.
 func (b *MemBroker) reclaimExpiredLocked(now time.Time) {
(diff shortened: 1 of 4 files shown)


yidakra:
@coderabbitai full review

Issue #437, opened 2026-06-26
pgvector/qdrant: path_prefix pushdown diverges from Filter.Match (silent under-return), pgvector HNSW 2000-dim limit breaks gemini-3072, existing store adopted without dim check, qdrant sentinel


yidakra:
**Part of #395.** Tier-C real-ANN backend gaps. (Verified SOUND: id-keyed idempotent upsert, REAL delete/eviction — better than the local backend, cosine metric consistent, no SQL injection, doc_types pushdown parity, TLS/auth, identity-guard `Reset` on model/dim change.)

## F1 [MED-HIGH] pgvector PathPrefix pushdown diverges from `Filter.Match` → silent under-return
`pgvectorindex/sql.go:202` pushes `rel_path LIKE 'prefix%' ESCAPE '\'` and `CanFilterFilter` (`:259`) claims pgvector can evaluate a `PathPrefix`. But Postgres `LIKE` is **case-sensitive** while `MatchesPathPrefix` (`model/pathprefix.go`) folds case (mirroring `list_files`, #286), and the pushdown uses the **raw** prefix while `Match` first runs `NormalizePathPrefix` (`./`→strip, `\`→`/`, `path.Clean`, strip leading `/`). So `path_prefix:"Docs"` or `"./docs"` matches under `Match` but the pushed-down LIKE drops the rows server-side; the retrieval recheck can only filter *down* what the backend returned and the widening loop re-issues the same restrictive LIKE → **silent empty/under-returned results** for case/form-mismatched prefixes. Qdrant is unaffected (it *declines* PathPrefix → Go-side Match is authoritative). **Fix:** `lower(rel_path) LIKE lower(...)` over a normalized prefix, or decline pushdown like qdrant.

## F2 [MED-HIGH] pgvector HNSW index has a hard 2000-dim limit → high-dim models permanently break the backend
[...]

PR #463, opened 2026-06-27
Addresses the path-prefix pushdown (F1) + HNSW 2000-dim guard (F2) parts of #437 (qdrant-sentinel/DSN/NaN deferred)

yidakra:
Fixes the two MED-HIGH parts of #437 in the pgvector backend. Scoped strictly to `internal/index/pgvectorindex/` (`sql.go`, `index.go`) + its tests. The smaller qdrant-sentinel / DSN-leak / NaN parts are intentionally left for a follow-up.

## F1 [MED-HIGH] — path_prefix pushdown diverged from `Filter.Match` -> silent under-return

`BuildFilterPredicates` pushed `rel_path LIKE 'prefix%' ESCAPE '\'` over the **raw** prefix. But:

- Postgres `LIKE` is case-**sensitive**, while `model.MatchesPathPrefix` is ASCII case-**insensitive**.
- The pushdown skipped normalization, while `Match` runs `NormalizePathPrefix` first (`./`-strip, `\`->`/`, `path.Clean`, leading-`/`-strip).

So a case- or form-mismatched prefix (e.g. `./Docs`, `docs/`, `DOCS/`) silently dropped valid rows that `Filter.Match` would keep.

**Fix:** normalize the prefix with the same `NormalizePathPrefix` and emit a case-insensitive `lower(rel_path) LIKE lower($n) ESCAPE '\'`. A prefix that normalizes away (`.`, `./`) imposes no constraint, mirroring the matcher. The backend pushdown and the authoritative Go-side recheck now agree (the lowercase LIKE can only over-return on non-ASCII folds, which the Go recheck removes — never under-return).

## F2 [MED-HIGH] — HNSW 2000-dim limit permanently broke high-dim models

`CreateHNSWIndexSQL` unconditionally created `USING hnsw (...)`. pgvector's hnsw/ivfflat indexes are capped at 2000 dims, so for `dim > 2000` (e.g. gemini-embedding-001 = 3072) `CREATE INDEX` failed, `tableReady` stayed false, and every `Upsert` re-failed corpus-wide.

**Fix:** `CreateHNSWIndexSQL` now takes the dimension and returns `ok=false` (empty SQL) above `HNSWMaxDim` (2000). `ensureVectorsTable` skips the index and logs a clear warning recommending a smaller Matryoshka dimension. The table stays fully queryable via exact (sequential-scan) search instead of breaking the backend. Dimension is threaded through from `ensureVectorsTable` where it is already known.

## Tests

- `BuildFilterPredicates` builds a case-insensitive predicate over a normalized prefix for `Docs` / `./docs` / `docs/` / backslash / leading-slash inputs, and emits no predicate when the prefix normalizes away.
- Cross-check: the SQL pushdown keeps exactly the rows `Filter.Match` keeps across case/form-mismatched prefixes.
- HNSW dim guard: created at/below 2000, skipped above; `HNSWMaxDim == 2000`.
- High-dim (3072) `Upsert` succeeds and skips the HNSW DDL while still running the INSERT.

`make check` passes fully (vet, golangci-lint, gocyclo<=15, misspell, all suites).

### DIFF
--- a/internal/index/pgvectorindex/index.go
+++ b/internal/index/pgvectorindex/index.go
@@ -15,6 +15,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"sync"
 
 	"github.com/jackc/pgx/v5"
@@ -97,7 +98,7 @@ type Index struct {
 
 	mu         sync.Mutex
 	dim        int  // 0 until known (config or first vector)
-	tableReady bool // vectors table + HNSW index created
+	tableReady bool // vectors table created (+ HNSW index when dim <= HNSWMaxDim)
 }
 
 // Open connects to Postgres, verifies pgvector is available, and ensures the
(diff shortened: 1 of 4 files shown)


yidakra:
@coderabbitai full review

Issue #416, opened 2026-06-25
providers: HTTP client hardening — gemini key leaks across redirects, unbounded success-body OOM, plaintext-http credentials, no Retry-After/jitter, no zero-vector validation


yidakra:
**Part of #395.** Provider-client robustness/security hardening. (Audited-sound: every client sets `http.Client.Timeout`, bodies are closed on all paths, error bodies are `io.LimitReader`-capped at 4096, no `InsecureSkipVerify`, keys not in URLs/logs.)

## 1 [MED-HIGH, security] Gemini API key leaks across HTTP redirects
`internal/gemini/client.go:363` sets the key in the custom header `x-goog-api-key`. No client sets `http.Client.CheckRedirect`, so Go follows up to 10 redirects and strips only `Authorization`/`Www-Authenticate`/`Cookie`/`Cookie2` cross-host — **custom headers are copied to the redirect target**. A malicious/compromised `base_url`, an upstream open-redirect, or an injected 302 forwards `x-goog-api-key` to an attacker host. Bearer clients (mistral/openai/omniembed) are safe (Authorization is stripped); Gemini is uniquely exposed via its non-standard header. **Fix:** set `CheckRedirect` to refuse redirects (or strip the key on host change).

## 2 [MED] No `io.LimitReader` on the SUCCESS response body → OOM
Error bodies are capped at 4096 bytes, but 200-OK bodies are decoded/read unbounded: `mistral/client.go:322,631,800,938`; `gemini:320,470,609,676`; `openai:206,316,422,469`; `omniembed:245`; `elevenlabs:187,288`. Clients send `Accept: application/json` but never set `Accept-Encoding`, so Go adds gzip and transparently inflates — a gzip-bombed or enormous 200 (most realistic for self-hosted omniembed or any custom `base_url`) is buffered with no ceiling → daemon OOM. **Fix:** wrap success bodies in `io.LimitReader` with a finite cap.

## 3 [MED, security] Credentialed requests allowed over plaintext http://
`NewClient` (openai `:91`, omniembed `:89`, gemini custom base `:122`, elevenlabs `:39`) accepts any `base_url` scheme with no enforcement/warning that a credentialed endpoint use https. A custom `http://` gateway with an api_key transmits `Authorization`/`xi-api-key` in cleartext → sniffable. **Fix:** warn or refuse when a non-empty key is paired with a non-https, non-loopback base URL.

## 4 [LOW-MED] 429 ignores Retry-After; backoff has no jitter
[...]

PR #464, opened 2026-06-27
Addresses the gemini-redirect-key-leak + success-body-OOM parts of #416 (Retry-After/jitter, plaintext-http, zero-vector deferred)

yidakra:
## Summary

Two of the security/robustness items from #416, scoped strictly to the provider HTTP clients (`internal/gemini`, `internal/openai`, `internal/omniembed`, `internal/mistral`, `internal/elevenlabs`). No retrieval/embedding/config code touched.

### FIX 1 — Gemini API key leaks across HTTP redirects (MED-HIGH, security)

`internal/gemini` authenticates with the custom `x-goog-api-key` header. Go follows redirects by default and copies custom headers onto the redirect target (it only strips `Authorization`/`Cookie` cross-host), so a malicious/compromised `base_url` or an injected `302` would forward the Gemini key to an attacker-controlled host.

Fix: set `CheckRedirect` on every Gemini HTTP client to **refuse** redirects (returns `http.ErrUseLastResponse`, surfaced as a non-2xx provider error). A direct provider API call never legitimately needs a redirect, so refusing is safe and the key is never sent onward. (Bearer-auth clients are not vulnerable — Go strips `Authorization` — so this hardening is applied to gemini, the affected client.)

### FIX 2 — no `io.LimitReader` on success bodies → gzip-bomb / huge-response OOM (MED)

Error bodies were already capped at 4096 bytes, but `200 OK` bodies were decoded/read **unbounded** (`json.NewDecoder(resp.Body)` / `io.ReadAll`) across all five clients. A malicious or buggy provider (especially a self-hosted omniembed or a custom `base_url`) could return a giant/gzip-bombed `200` and OOM the daemon.

Fix: every success-body read/decode now goes through `io.LimitReader` with a finite cap — **64 MiB** for JSON (embeddings/chat/OCR/STT), **256 MiB** for TTS audio (higher so legitimate large audio is not truncated). Exceeding the cap returns a clear, non-retryable provider error instead of buffering unbounded. Implemented as a small `readLimitedBody(resp, limit)` helper per client (reads one byte past the cap to detect overflow without buffering the whole body).

## Tests

- **gemini redirect**: a test server returning `302` to a second server — the client errors and the redirect target is never contacted, so the `x-goog-api-key` is never leaked.
- **over-limit body** (all five clients): the limited reader returns a non-retryable `*_FAILED` error on an over-cap body and returns within-cap bodies in full.
- Refactored `mistral.embedBatch` (extracted `collectEmbedVectors`) to keep `gocyclo` under 15.
- `make check` passes locally (fmt, vet, lint, cyclo, ineffassign, misspell, full test suite, release-tools, build).

## Deferred (still open in #416)

Retry-After/jitter, plaintext-http guard, and zero-vector handling are intentionally **not** in this PR.

### DIFF
--- a/internal/elevenlabs/client.go
+++ b/internal/elevenlabs/client.go
@@ -21,6 +21,15 @@ const (
 	defaultBaseURL  = "https://api.elevenlabs.io"
 	defaultTimeout  = 30 * time.Second
 	defaultSTTModel = "scribe_v1"
+
+	// maxResponseBytes caps a JSON success response body (STT) so a malicious
+	// or buggy upstream (e.g. a gzip bomb behind a custom base_url) cannot
+	// drive unbounded memory use when we buffer it (issue #416).
+	maxResponseBytes = 64 << 20 // 64 MiB
+	// maxAudioResponseBytes caps a TTS audio body. Audio is legitimately
+	// larger than JSON, so it gets a higher ceiling while still bounding the
+	// read.
+	maxAudioResponseBytes = 256 << 20 // 256 MiB
 )
 
 type Client struct {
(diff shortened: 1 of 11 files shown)


yidakra:
@coderabbitai full review

PR #465, opened 2026-06-27
fix(status): reconcile stale "running" snapshot against daemon liveness

yidakra:
Addresses the status-stale-running part of #418 (reindex-guard in #457; pid-reuse deferred).

## Problem

`internal/cli/status.go` read `snapshot.Indexing.Running` straight from `corpus.json` (daemon-written) with no liveness cross-check. A daemon that wrote `running=true` and then crashed left `status` reporting indexing as "running" (green) forever — it never consulted the pid file.

## Fix

In the status command, reconcile the reported indexing-running state against daemon liveness using the existing pid-liveness mechanism (`pidFilePath` + `readPIDFile` + `processIsAlive`, reused — not reimplemented):

- If `corpus.json` says `running=true` but no live daemon process owns the state dir (pid file absent or process not alive), report indexing as **stopped** and note the snapshot is stale.
- The normal case (a live daemon writing `running=true`) still reports running.
- Output contract is stable: human output unchanged except a `(stale snapshot; daemon not running)` note; JSON `indexing.running` is reconciled to `false` with an additive, optional `stale_running` flag.

Scope kept to `internal/cli/status.go` plus tests.

## Tests

`tests/cli/command_behavior_test.go`:
- `TestStatusStaleRunningReportsStopped` — `running=true`, no pid file → reports stopped + stale note.
- `TestStatusStaleRunningJSONReconcilesRunningFalse` — JSON envelope reports `running=false`, `stale_running=true`.
- `TestStatusLiveDaemonReportsRunning` — live pid (test process) → still reports running, no stale flag.

`make check` passes (build, vet, golangci-lint, gocyclo -over 15 clean, misspell, full test suite).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `status` output to better detect when indexing appears running but no live process is actually using the state directory.
  * Added a clearer “stale snapshot” message in human-readable output for this case.
  * JSON status output now includes a `stale_running` field and reports the indexing state more accurately when the snapshot is outdated.

### DIFF
--- a/internal/cli/status.go
+++ b/internal/cli/status.go
@@ -58,15 +58,44 @@ func (a *App) runStatus(ctx context.Context, global globalOptions, args []string
 		source = "computed"
 	}
 
-	return a.renderStatusOutput(global, cfg.StateDir, snapshot, source)
+	// Reconcile a daemon-written "running" snapshot against process liveness.
+	// corpus.json is written by the daemon, so a daemon that recorded
+	// running=true and then crashed would otherwise leave status reporting
+	// indexing as "running" forever. If no live daemon owns this state dir, the
+	// snapshot is stale: report indexing as stopped instead (#418).
+	staleRunning := false
+	if snapshot.Indexing.Running && !daemonIsLive(cfg.StateDir) {
+		snapshot.Indexing.Running = false
+		staleRunning = true
+	}
+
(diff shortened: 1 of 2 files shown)


yidakra:
@coderabbitai full review

Issue #415, opened 2026-06-25
SECURITY: elevenlabs-bridge has zero inbound authentication — anonymous callers run ask/search against the private corpus and drive LLM spend


yidakra:
**Security.** The ElevenLabs bridge is a REST→MCP reverse proxy that holds the dir2mcp bearer token and makes **authenticated** MCP calls on behalf of **any unauthenticated** caller (auth-laundering).

## Verified
`internal/elevenlabsbridge/server.go` `routes()` registers `/ask`, `/search`, `/list_files`, `/stats` guarded **only** by `methodGuard` (HTTP method check). A grep for `Authorization`/`hmac`/`signature`/`secret` on the inbound path finds **nothing** — no Authorization check, no HMAC/signature verification, no timestamp/replay window, no shared secret on any route. The bridge resolves the dir2mcp MCP token (`config.go:105`) and uses it for the **outbound** call regardless of who called in.

`--listen` (`internal/cli/bridge.go:90,120`) accepts any address (e.g. `0.0.0.0:8088`) with **no guard and no warning** — unlike the main server, which forces auth for `--public` unless `--force-insecure`. The documented use case (ElevenLabs Conversational AI calling the bridge) **requires** a publicly reachable URL, so real-world deployment is exactly the exposed, unauthenticated configuration.

## Symptom
Anyone who can reach the listen socket runs `ask`/`search`/`list_files`/`stats` against the private corpus (data exfiltration) and drives unbounded LLM/embedding spend (cost-amplification DoS). Remote-exploitable whenever bound non-loopback.

## Fix
Require a shared secret / signature (ideally the documented ElevenLabs webhook HMAC, verified constant-time with a timestamp/replay window) on inbound requests, AND refuse non-loopback binds without an explicit `--force-insecure`-style override mirroring the main server.

yidakra:
Fixed by #466 (merged to main). Closing — the PR's scope fully addresses this issue; it wasn't auto-closed (no `Closes` keyword in the PR).

PR #466, opened 2026-06-27
fix(elevenlabsbridge): require inbound auth + guard non-loopback binds (#415)

yidakra:
## Problem

The ElevenLabs bridge (`internal/elevenlabsbridge`) is a REST→MCP reverse proxy that holds the dir2mcp bearer token and made **authenticated** MCP calls on behalf of **any anonymous** caller (auth-laundering). The inbound routes `/ask`, `/search`, `/list_files`, `/stats` were guarded **only** by an HTTP method check — no `Authorization`, secret, HMAC, or signature on any route. `--listen` accepted any address (e.g. `0.0.0.0`) with no guard, and the documented ElevenLabs use case needs a publicly reachable URL. Result: anyone reaching the socket read the private corpus (data exfiltration) and drove unbounded LLM/embedding spend (cost-amplification DoS).

## Fix (two parts)

**1. Inbound authentication.** New shared-secret config field, set via `BRIDGE_INBOUND_SECRET` env or `--inbound-secret`. When configured, every protected route requires the secret presented as `Authorization: Bearer <secret>` or `X-Bridge-Secret: <secret>`, verified with `crypto/subtle.ConstantTimeCompare`. Missing/wrong credentials return `401`. `/health` stays unauthenticated (it returns no corpus data). Auth runs before the method guard and before any backend call.

**2. Non-loopback guard.** Mirrors the main server's `--public`/`--force-insecure` contract. Binding a non-loopback address (wildcard `0.0.0.0`/`::`, empty host, or a routable IP) **without** an inbound secret is refused with a clear, actionable error unless `--force-insecure` is set. Loopback binds (`127.0.0.1`, `::1`, `localhost`) with no secret stay allowed for local dev. Enforced both at the CLI (clean config error) and in `RunWithBridge` (defense-in-depth).

The CLI/JSON status now reports `inbound auth=secret|none`.

## Scope

Touches only `internal/elevenlabsbridge/{config.go,server.go}`, `internal/cli/bridge.go`, and `tests/elevenlabsbridge/bridge_test.go`.

## Tests

- Protected route returns `401` without the secret and `200` with the correct secret (both header forms); the backend is never reached by unauthenticated callers; `/health` stays open.
- `ValidateListenSecurity` table covering loopback/wildcard/empty-host/routable-IP × secret/force-insecure.
- `RunWithBridge` refuses a non-loopback bind without a secret and allows a loopback bind without one.

Full `make check` passes (gofmt, vet, golangci-lint, gocyclo, ineffassign, misspell, tests, build).

## Follow-up

The PREFERRED ElevenLabs webhook **HMAC** signature verification (HMAC-SHA256 over the raw body with a timestamp/replay window) is left as a follow-up; the constant-time shared-secret check fully closes the open hole and is a large improvement.

Addresses #415 (inbound auth + non-loopback guard; HMAC follow-up).

### DIFF
--- a/internal/cli/bridge.go
+++ b/internal/cli/bridge.go
@@ -36,10 +36,20 @@ func (a *App) runBridgeElevenLabs(ctx context.Context, global globalOptions, arg
 		return parseCode
 	}
 
+	if err := elevenlabsbridge.ValidateListenSecurity(listen, cfg.InboundSecretConfigured(), cfg.ForceInsecure); err != nil {
+		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid, err.Error())
+		return exitConfigInvalid
+	}
+
 	bridge, err := elevenlabsbridge.New(cfg)
 	if err != nil {
 		writeCLIError(a.stderr, global.jsonOutput, exitGeneric, err.Error())
 		return exitGeneric
+	}
+
+	inboundAuth := "none"
+	if bridge.InboundAuthEnabled() {
+		inboundAuth = "secret"
 	}
 
 	if global.jsonOutput {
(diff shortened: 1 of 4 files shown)


yidakra:
@coderabbitai full review

Issue #397, opened 2026-06-25
mcp: span with empty/unknown kind serializes to {"kind":""} → fails Span oneOf → strict-client 'Failed to call tool'


yidakra:
**Part of #395.** Another instance of the #387 class (strict MCP clients validate `structuredContent` against `outputSchema` and reject the whole call), distinct from #387's `additionalProperties` cause.

## Verified bug
- `buildOpenFileSpan` (`internal/mcp/tools.go:2368`, default branch) returns `{"kind": kind}` for any `kind` not in {lines,page,time,region} — **including `kind == ""`**.
- The `Span` schema (`spanDefinitionSchema`) is `oneOf` of five branches, each requiring `kind` to be a specific `const` (lines/page/time/region/document). An empty or unknown kind matches **zero** branches → `oneOf` fails → strict client (Claude Desktop) rejects the entire tool result with "Failed to call tool".
- Reachability is real: `serializeHit` (`tools.go:2569`) **unconditionally** emits `span: buildOpenFileSpan(h.Span)` for every hit, and `internal/retrieval/hybrid.go:131` explicitly guards `if cached.Span.Kind != ""` — i.e. BM25 hits whose cached metadata never populated a span kind ship with `Span.Kind == ""`. Affects `search`, `ask`, `ask_audio`, `transcribe_and_ask`, and `search_only`.

Data-dependent, which is why it hasn't bitten yet: corpora whose hits carry docling `region` spans (e.g. the stas-legal corpus) never hit the empty-kind path, so the release-smoke schema check passes. A lexical-heavy corpus/query that surfaces BM25 hits without span metadata would trip it → intermittent "Failed to call tool" on a different corpus.

## Fix
Default branch should emit a schema-valid span — `{"kind":"document"}` (a valid `oneOf` branch) — for empty/unknown kinds, instead of echoing the raw kind. `span` is `required` in the `Hit` schema, so omission isn't an option; the `document` fallback is the safe fix. Add a regression test asserting `buildOpenFileSpan` output validates against `spanDefinitionSchema` for empty/unknown kinds.

## Related (low / likely benign, note only)
`newToolErrorResult` (`tools.go:322`) sets `structuredContent.error`, which no `outputSchema` declares (all are `additionalProperties:false`) — structurally the #387 shape. Benign in practice because clients skip `outputSchema` validation when `isError==true`; flagged for completeness. No action unless a strict client is observed validating error results.

## Prevention
[...]

Issue #387, opened 2026-06-24
mcp: search/ask/open_file results rejected by strict clients — hit outputSchema missing modality/media_ref (Failed to call tool)


yidakra:
## Symptom
In Claude Desktop, `dir2mcp_search` / `dir2mcp_ask` (and `open_file`) fail **immediately** with **"Failed to call tool"**, while `dir2mcp_stats` and `dir2mcp_list_files` work — even though the daemon is healthy (direct curl returns HTTP 200 with a full answer in ~3s, and `make release-smoke` passes). The server-side `query_metrics` show the calls succeeding; the `mcp-remote` bridge log shows the full response coming back (`[Remote→Local]`). Yet Claude marks the tool failed.

## Root cause
Claude Desktop (and other strict MCP clients) **validate `structuredContent` against the tool's declared `outputSchema`** and reject the entire result on any violation. `hitDefinitionSchema` declares the hit object with `additionalProperties: false` and these properties: chunk_id, rel_path, title, doc_type, rep_type, score, snippet, span. But `serializeHit` also emits **`modality`** and **`media_ref`** (conditionally, for media/multimodal chunks, SPEC 8.1.7). Those undeclared fields violate `additionalProperties: false`, so the whole `search`/`ask` result is rejected → "Failed to call tool".

`stats`/`list_files` work because their structured content matches their schemas. Direct curl and the smoke test never caught it because they don't validate against `outputSchema` — only the client does. This is why the server looked perfectly healthy while Claude failed.

Validated against a live daemon (jsonschema): every `ask`/`search` hit raised `Additional properties are not allowed ('modality' was unexpected)`. Adding `modality` + `media_ref` to the hit schema makes ask/search/open_file all validate.

## Fix
- Add `modality` and `media_ref` (both `string`) to `hitDefinitionSchema` so the schema matches `serializeHit`.
- Add a regression guard so serializer/schema drift can't recur (every key a serializer can emit must be declared in the tool's outputSchema).

## Impact
This is THE cause of the recurring "Failed to call tool" on docling/media corpora (any corpus whose hits carry `modality`). High severity — search/ask are unusable in Claude Desktop despite a healthy server.

PR #467, opened 2026-06-27
test(mcp): validate tool structuredContent against declared outputSchema (#428)

### LINKED ISSUE #428: tests: critical paths untested let the recurring regressions ship green — no outputSchema validation, no production-transport e2e, no vector-misalignment/auth/payment-verdict coverage
**Meta-issue.** The suite is "donut-shaped": strong on isolated pure functions, weak/absent exactly where bugs have historically shipped (#387 "Failed to call tool", #364 empty embeddings, #400 payment fail-open, #401 panic-crash). Most recurring regressions are caught only by the manual `make release-smoke`, not by `make ci`.

## Critical gaps
- **C1 — No test validates `structuredContent` against the declared `outputSchema`** (the #387 class). Zero tests import a JSON-Schema validator; `tests/mcp/tools_test.go:78` checks only that a schema *exists*, and `tools_internal_test.go:36` checks only that `serializeHit` keys are *declared* (no types/required/enum/additionalProperties, search/ask only). dir2mcp registers via raw `Server.AddTool`, so the SDK doesn't validate server-side — any serializer change reproduces #387 with the suite green.
- **C2 — No e2e test does a *successful* `tools/call` over the production SDK transport.** Every output-inspecting test drives the legacy `.Handler()` path users don't connect to; `convertToolCallResult` (`transport_sdk.go:411`, incl. the audio base64 decode/drop + empty-content shaping) and the SDK result serialization — the exact "Failed to call tool" surface — are never exercised in a passing tool call.
- **C3 — Embed batch vector-count mismatch untested.** `fakeEmbedder.Embed` (`embedding_worker_test.go:248`) returns preset vectors regardless of input length, so a provider returning N-1 vectors for N inputs (→ chunk A's text indexed under chunk B's vector, citation poisoning) can never be produced; the `ErrFatal` guard (`embedding_worker.go:181`) is never exercised.

## High gaps
- **H1 — Auth decision untested:** only the missing-header → 401 path is tested; the accept path and the security-critical wrong-token reject via `subtle.ConstantTimeCompare` (`server.go:443`) are not — invert the compare and tests pass.
- **H2 — Invalid-payment-at-HTTP-200 never tested** (the #400 class): all "invalid" cases use a 4xx facilitator; none returns `{isValid:false}` at 200, so the fail-open is uncatchable by tests.
- **H3 — Panic-recovery test covers the wrong path** (#401): `tool_panic_recovery_test.go` calls `invokeToolHandler` directly, whose `recover()` the default (x402-off) transport bypasses (`transport_sdk.go:400`).
- **H4 — Hybrid BM25+vector fusion never executes:** no retrieval test constructs a `model.LexicalSearcher`, so `runHybridSearch` always returns the vector-only fallback; the BM25-error/empty/metadata-reattach paths are uncovered.
- **H5-H9:** partial-batch index-failure recovery, failure-category wiring, stale-vector/shrink eviction (no real-store test of `SoftDeleteChunksFromOrdinal`), open_file page slicing (only middle page of 3), and the daemon+embed-worker integration tests are env-gated out of `make ci`.

## Highest-leverage fix
**One e2e test** that (a) drives `NewSDKTransport.Serve` with x402 off, (b) `initialize` → `tools/call` each tool with realistic data, (c) reads each tool's `outputSchema` from `tools/list`, (d) validates the returned `structuredContent` against it with a real JSON-Schema validator — closes C1, C2, and H3 together and would have caught the recurring #387 / "Failed to call tool" classes in CI. Add it, plus auth accept/reject (H1) and the 200-`isValid:false` payment case (H2), and wire `-race` (#419) + the integration tests into CI.

yidakra:
## What

Adds a CI-side conformance test (`internal/mcp/output_schema_conformance_test.go`, new file only) that validates each MCP tool's representative `structuredContent` against its declared `outputSchema`. This is the highest-leverage prevention from #428: it ports the `schema_errors` mini-validator from `scripts/release_smoke.py` to Go so the **#387 "Failed to call tool" class** (a serialized field not allowed by an `additionalProperties:false` / `oneOf` schema) is caught at unit-test time, not just in a live release-smoke run.

## How

`schemaErrors(schema, instance, root, path)` is a small, dependency-free recursive validator covering the JSON Schema subset the dir2mcp outputSchemas use:

- `$ref` → resolved against root `#/definitions/...`, **fail-closed** on an unresolved ref
- `oneOf` → instance must match **exactly one** branch
- `const`, `enum`
- object → `properties` + `additionalProperties:false` rejects undeclared keys + `required` must be present (also handles `additionalProperties: {schema}` for `stats.doc_counts`)
- `array` / `items`
- scalar `type` (string/integer/number/boolean) — **bool rejected** for integer/number, mirroring the Python guard

The recursion is split into small helpers so each function stays under the gocyclo budget.

## Coverage

- **search / ask / ask_audio / transcribe_and_ask** — built with the **real serializers** (`serializeHit` / `buildAskStructuredContent`) using a hit with every conditional field populated (`Title`, `Modality`, `MediaRef`, region span) — the exact docling-corpus shape that broke #387.
- **open_file / open_media_clip / list_files / transcribe / annotate / stats** — built as the handlers return them.
- **Negative control** — injects an undeclared hit field and asserts the validator *detects* it, proving it would catch #387.
- **Span variants** — `buildOpenFileSpan` is checked for every kind (lines / page / time / time-diarized / region / region-degraded / document) against the shared `Span` schema.

## Empty-kind span (#397)

On current `main`, `buildOpenFileSpan(Span{Kind: ""})` emits `{"kind": ""}`, which matches **no** `Span` `oneOf` branch — the #397 class. Per the issue scope, the test surfaces this as a **`t.Log` known gap** rather than failing, so the suite is green on `main` while documenting the invariant the #397 fix must restore. When that fix lands, the log branch flips to a hard assertion.

## Notes

- New test file only — no existing non-test `.go` files edited.
- `go.mod` / `go.sum` unchanged (no new dependency).
- Full `make check` passes (including lint/gocyclo and the Python validator unit tests).

Closes #428.

### DIFF
--- a/internal/mcp/output_schema_conformance_test.go
+++ b/internal/mcp/output_schema_conformance_test.go
@@ -1,0 +1,544 @@
+package mcp
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/dirstral/dir2mcp/internal/model"
+)
+
+// This file locks in the invariant that every MCP tool's representative
+// structuredContent validates against the tool's declared outputSchema. It is
+// the Go port of the schema_errors mini-validator in scripts/release_smoke.py
+// (and the highest-leverage prevention from issue #428): a strict MCP client
+// (Claude Desktop) validates structuredContent against outputSchema and rejects
+// the WHOLE call ("Failed to call tool") when a serialized field is not allowed
+// by an additionalProperties:false / oneOf schema. That is exactly the #387 bug
+// class (hits[].modality / media_ref undeclared) and the #397 class (a Span with
+// an unknown kind matching no oneOf branch). Asserting it here catches the class
+// in CI instead of in a release smoke run.
+
+// schemaErrors is a minimal recursive validator for the subset of JSON Schema the
+// dir2mcp outputSchemas use: $ref (resolved against root #/definitions/...),
+// oneOf (exactly one branch must match), const, enum, object (properties +
+// additionalProperties:false rejects undeclared keys, required must be present),
+// array/items, and scalar type (string/integer/number/boolean). It is
+// deliberately small and dependency-free; conditional keywords (allOf/anyOf/
+// if/then) are ignored, matching the release-smoke validator. It fails closed on
+// an unresolved $ref. The returned strings are human-readable error paths; an
+// empty slice means the instance conforms.
+func schemaErrors(schema interface{}, inst interface{}, root map[string]interface{}, path string) []string {
+	sch, ok := schema.(map[string]interface{})
+	if !ok {
+		return nil
+	}
+	if ref, ok := sch["$ref"].(string); ok {
+		target := resolveRef(ref, root)
+		if target == nil {
+			return []string{path + ": unresolved $ref " + ref}
+		}
+		return schemaErrors(target, inst, root, path)
+	}
+	if branches, ok := sch["oneOf"].([]interface{}); ok {
+		matched := 0
+		for _, b := range branches {
+			if len(schemaErrors(b, inst, root, path)) == 0 {
+				matched++
+			}
+		}
+		if matched == 1 {
+			return nil
+		}
+		return []string{path + ": matched " + itoa(matched) + " oneOf branches (want 1)"}
+	}
+	if c, ok := sch["const"]; ok {
+		if reflect.DeepEqual(inst, c) {
+			return nil
+		}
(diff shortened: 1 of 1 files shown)


yidakra:
@coderabbitai full review

Issue #399, opened 2026-06-25
embeddings: graceful-degradation gaps — batch poison-chunk, no input token/size cap, invalid creds pass preflight, reranker on truncated snippets


yidakra:
**Part of #395.** Robustness gaps in the embedding/retrieval path where a recoverable condition instead fails (or silently degrades) more than it should. (Agent-audited with file:line; the top items below are worth re-confirming during implementation.)

## 1 [HIGH] One poison chunk fails its whole batch (no per-item isolation)
`internal/index/embedding_worker.go:705-724`: on a non-transient `embedTasks` error, `EmbedAndIndex` marks **all** labels in the batch failed. Providers embed by HTTP batch (openai 64, gemini 64, omniembed 32) and any single rejected input (oversized / bad encoding / over token limit) 400s the whole request — so one bad chunk marks up to 31 healthy siblings `embedding_failure`. **Fix:** on a batch-level non-retryable error, bisect/retry per-item so only the offending chunk(s) fail.

## 2 [MED] No per-provider input token/size cap → oversized input hard-fails
The embed clients slice by item count only (`openai/client.go:146-163`, `gemini/client.go:247-265`, `omniembed/client.go:186-202`); none enforce a per-provider token budget or per-input length cap (`omniembed` has only a 50MB media byte ceiling). `providerfactory.validateEmbedDim` validates the *output* dim, not *input* size. A chunk over the model's max input tokens → provider 400 → (via #1) whole batch fails. **Fix:** model per-provider input limits; truncate or split oversize inputs before sending.

## 3 [MED] Present-but-invalid credentials pass preflight, then fail the whole corpus
`provider.Profile.Eligible()` (`internal/provider/provider.go:197-199`) returns true on any non-empty key; preflight builds the adapter but never authenticates (`up.go:448-451`). A typo'd/expired key passes startup, then every embed 401/403s → `*_AUTH` non-retryable → corpus-wide failure. **Fix:** a one-shot probe embed in preflight (also catches the wrong-model case in #396) → clear CONFIG/credential error at startup instead of silent corpus-wide failure.

## 4 [MED] LexicalSearcher assertion silently degrades BM25
`internal/retrieval/hybrid.go:112-115`: `ls, ok := store.(model.LexicalSearcher); if !ok { return nil, false }` with no log — hybrid silently drops to vector-only if the store stops satisfying the interface (the BM25-regression class, invisible at runtime). **Fix:** warn on the negative branch.

[...]

PR #478, opened 2026-07-01
fix(index): isolate embedding batch poison chunks via bisection (#399)

yidakra:
## Addresses #399

Robustness gaps in the embedding / retrieval path where a recoverable condition
failed (or silently degraded) more than it should. This PR implements the
HIGH-priority poison-chunk item plus one cheap observability item; the remaining
items are deferred to focused follow-ups (see below).

### Item 1 [HIGH] — one poison chunk no longer fails its whole batch
`internal/index/embedding_worker.go`

Providers embed by HTTP batch (openai 64, gemini 64, omniembed 32) and a single
rejected input (oversized / bad encoding / over token limit) 400s the whole
request. The worker previously marked **all** labels in the batch
`embedding_failure` on a non-transient error, so one bad chunk took down up to
batch-size-1 healthy siblings.

`EmbedAndIndex` now delegates to a recursive `embedIndexBatch`: on a
non-transient, non-fatal, multi-item batch error it **bisects** the batch and
re-embeds each half, narrowing to the individual offending chunk — which is
marked `embedding_status=error` (retryable-vs-permanent classification preserved
from #412) — while every healthy sibling still embeds and indexes.

Contract preserved:
- **Transient** errors (network / rate-limit / cancellation, #412) still leave
  chunks PENDING and are never marked failed. A transient error from a bisected
  half takes precedence in the return value so still-pending chunks are retried.
- **ErrFatal** (a local media/config failure, not a provider input rejection)
  keeps its existing whole-batch behaviour so genuine misconfiguration stays
  loud rather than degrading into a silent per-chunk failure storm.
- The single-item path (used by the distributed embed-worker,
  `internal/embedqueue`) is unchanged — bisection only engages for multi-item
  in-process batches.

### Item 4 [LOW] — warn when hybrid silently degrades to vector-only
`internal/retrieval/hybrid.go`, `internal/retrieval/service.go`

When hybrid is enabled but the store does not satisfy `model.LexicalSearcher`,
retrieval silently dropped to vector-only (the BM25-regression class, invisible
at runtime). It now logs a warning **once** (sync.Once-guarded, so no per-query
flood) so the degradation is diagnosable.

### Tests
`tests/index/embed_poison_isolation_test.go`:
- one poison chunk in a 4-chunk batch → the 3 healthy chunks embed+index, only
  the poison chunk is marked failed (with a classification category), and the
  worker bisected (asserted via embed-call count).
- all-poison batch → each chunk marked failed individually, no run-loop error.
- transient batch error → all chunks left PENDING, none marked failed, error
  surfaced for retry.

### Deferred (separate follow-ups)
- **Item 2 [MED]** per-provider input token/size caps — needs per-provider
  limits modelled across the openai/gemini/omniembed clients; oversized inputs
  now degrade gracefully (isolated + marked error) via item 1 instead of failing
  the batch.
- **Item 3 [MED]** preflight probe-embed to catch present-but-invalid creds.
- **Item 5 [LOW]** rerank against full chunk text instead of the truncated BM25
  snippet.
- **Item 6 [LOW]** include native vector width in the HNSW `EmbedIdentity`.

### Checks
`make check`: lint (0 issues), `go vet`, gocyclo (≤15), ineffassign, misspell,
and the embedding/retrieval suites all pass. (One unrelated `tests/cli`
timeout — `TestUpCreatesSecretTokenAndConnectionFile`, "initialize metadata
store: context deadline exceeded" — was a load-induced flake under heavy
parallel test load; it passes in isolation.)

https://claude.ai/code/session_01QHuQEKSW3nzU1Pn7KXXBhj

<h3>Greptile Summary</h3>

This PR fixes two robustness gaps in the embedding/retrieval path. The core change in `embedding_worker.go` introduces a recursive bisection strategy so a single provider-rejected "poison" chunk (oversized input, bad encoding) no longer marks all batch-mates failed; the fix adds an `errChunkMarkedFailed` sentinel that is absorbed within bisection but remains opaque to the run loop so post-embed write errors still propagate correctly. A companion observability fix in `hybrid.go`/`service.go` surfaces a one-time warning when hybrid search silently degrades to vector-only.

- **`embedIndexBatch` + `bisectEmbedBatch`** — on non-transient, non-fatal, non-auth batch embed failure, the batch is split recursively until the offending item is isolated and marked `embedding_status=error`; every healthy sibling still embeds and indexes.
- **`errChunkMarkedFailed` sentinel** — distinguishes \"chunk already terminal (absorbable by bisection)\" from \"post-embed write failure (must propagate to run loop)\"; the previous thread's concern that `indexChunks`/`markEmbeddedWithRetry` errors were silently dropped is addressed and pinned by `TestBisect_PostEmbedWriteErrorPropagates`.
- **`hybridNoLexicalWarned atomic.Bool`** — replaces the unresetable `sync.Once` approach; `SetHybridEnabled(true)` resets the flag so a hot-reload cycle re-arms the degradation warning, addressing the previous thread's concern about the once staying spent.

<h3>Confidence Score: 5/5</h3>

Safe to merge — the bisection logic is correct, the error propagation contracts are preserved, and the two issues raised in previous review rounds are both demonstrably fixed and test-pinned.

The `errChunkMarkedFailed` sentinel correctly separates terminal chunk failures (absorbed within bisection so healthy siblings still succeed) from post-embed write failures (propagated to the run loop so tight re-embed loops are avoided). `mergeBisectErrors` gives `ErrFatal` priority over any sibling transient error, transient errors surface unchanged, and the sentinel cannot wrap either of those two cases. The `atomic.Bool` reset in `SetHybridEnabled` correctly re-arms the degradation warning on hot-reload. All new code paths are covered by dedicated tests including an explicit regression test for the prior post-embed write error propagation concern.

No files require special attention.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| internal/index/embedding_worker.go | Adds `embedIndexBatch`/`bisectEmbedBatch`/`mergeBisectErrors` and the `errChunkMarkedFailed` sentinel; both previous-thread P1s (post-embed write errors silently absorbed, no bisection) are addressed; transient/fatal error contracts preserved. |
| internal/retrieval/hybrid.go | Adds a `CompareAndSwap`-guarded one-time warning when the store does not implement `LexicalSearcher`; correctly reads `hybridNoLexicalWarned` after releasing `metaMu`, safe because the field is an `atomic.Bool`. |
| internal/retrieval/service.go | Adds `hybridNoLexicalWarned atomic.Bool`; `SetHybridEnabled(true)` resets it so hot-reload cycles re-arm the degradation warning; addresses previous-thread concern about `sync.Once` staying spent after toggle. |
| tests/index/embed_poison_isolation_test.go | Five new tests covering: healthy-sibling isolation, all-poison batch, transient-leaves-PENDING, ErrFatal propagation through bisection, and the previous-thread P1 (post-embed write error propagates). |

<h3>Flowchart</h3>

<a href="#gh-light-mode-only">

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A["EmbedAndIndex\n(batch of N chunks)"] --> B["embedIndexBatch"]
    B --> C{embedTasks}
    C -- "transient error" --> D["Return error\nChunks stay PENDING"]
    C -- "non-transient error\nlen>1, !ErrFatal, !auth" --> E["bisectEmbedBatch"]
    C -- "single item OR\nErrFatal OR auth" --> F["MarkFailedWithCategory\n(whole batch)"]
    F -- "ErrFatal" --> G["Return ErrFatal\n(worker stops)"]
    F -- "other" --> H["Return errChunkMarkedFailed\n(absorbed by parent bisection)"]
    E --> I["embedIndexBatch\n(left half)"]
    E --> J["embedIndexBatch\n(right half)"]
    I & J --> K["mergeBisectErrors"]
    K -- "errChunkMarkedFailed only" --> L["Return nil\nHealthy siblings indexed ✓"]
    K -- "ErrFatal present" --> G
    K -- "transient or write error" --> D
    C -- "success" --> M["indexChunks"]
    M -- "error" --> N["Return write error\n(propagates to run loop)"]
    M -- "success" --> O["markEmbeddedWithRetry"]
    O -- "error" --> N
    O -- "success" --> P["Return n, nil"]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
flowchart TD
    A["EmbedAndIndex\n(batch of N chunks)"] --> B["embedIndexBatch"]
    B --> C{embedTasks}
    C -- "transient error" --> D["Return error\nChunks stay PENDING"]
    C -- "non-transient error\nlen>1, !ErrFatal, !auth" --> E["bisectEmbedBatch"]
    C -- "single item OR\nErrFatal OR auth" --> F["MarkFailedWithCategory\n(whole batch)"]
    F -- "ErrFatal" --> G["Return ErrFatal\n(worker stops)"]
    F -- "other" --> H["Return errChunkMarkedFailed\n(absorbed by parent bisection)"]
    E --> I["embedIndexBatch\n(left half)"]
    E --> J["embedIndexBatch\n(right half)"]
    I & J --> K["mergeBisectErrors"]
    K -- "errChunkMarkedFailed only" --> L["Return nil\nHealthy siblings indexed ✓"]
    K -- "ErrFatal present" --> G
    K -- "transient or write error" --> D
    C -- "success" --> M["indexChunks"]
    M -- "error" --> N["Return write error\n(propagates to run loop)"]
    M -- "success" --> O["markEmbeddedWithRetry"]
    O -- "error" --> N
    O -- "success" --> P["Return n, nil"]
```

</a>

<sub>Reviews (2): Last reviewed commit: ["review: address PR #478 comments"](https://github.com/dirstral/dir2mcp/commit/195c19d270ecc12562a8acd51faa90187980b814) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41247262)</sub>

### DIFF
--- a/internal/index/embedding_worker.go
+++ b/internal/index/embedding_worker.go
@@ -708,22 +708,44 @@ func (w *EmbeddingWorker) EmbedAndIndex(ctx context.Context, indexKind string, t
 	if err != nil {
 		return 0, err
 	}
+	return w.embedIndexBatch(ctx, indexKind, modelName, validTasks, labels)
+}
 
+// embedIndexBatch embeds validTasks (aligned 1:1 with labels), upserts the
+// resulting vectors into the index, and marks the chunks embedded. It is the
+// recursive core of EmbedAndIndex.
+//
+// On a non-transient, non-fatal batch embed error it bisects the batch
+// (bisectEmbedBatch) so a single provider-rejected "poison" input — an
+// oversized / bad-encoding / over-token-limit chunk that 400s the whole HTTP
+// request, since openai/gemini/omniembed all embed by batch — is isolated and
+// marked failed while its healthy siblings still embed (issue #399). Previously
+// one bad chunk marked up to batch-size-1 healthy siblings embedding_failure.
+//
+// The classification contract from issue #412 is preserved: transient errors
+// (network hiccup, rate limit, cancellation) leave the chunks PENDING for a
+// later cycle and are never marked failed; ErrFatal is a local media/config
+// failure (not a provider input rejection) and keeps the existing whole-batch
+// behavior so genuine misconfiguration stays loud rather than degrading into a
+// silent per-chunk failure storm.
+func (w *EmbeddingWorker) embedIndexBatch(ctx context.Context, indexKind, modelName string, validTasks []model.ChunkTask, labels []uint64) (int, error) {
 	// Text chunks embed via Embed; media chunks (SPEC 8.1.7) embed via
 	// EmbedMedia. embedTasks returns vectors aligned 1:1 with validTasks.
 	vectors, err := w.embedTasks(ctx, modelName, validTasks)
 	if err != nil {
-		// distinguish between transient errors (which we want to retry later)
-		// and permanent failures for which the chunks should be marked as
-		// irrecoverable.  A transient error could be a network timeout,
-		// rate‑limit response, or context cancellation.  We intentionally keep
-		// the interface simple; by returning the error without marking the
-		// chunks as failed they will remain in the pending state and be
-		// re‑fetched on the next cycle.  Permanent errors fall through to the
-		// existing MarkFailed behaviour.
+		// A transient error could be a network timeout, rate-limit response, or
+		// context cancellation. Returning it without marking the chunks failed
+		// leaves them PENDING so they are re-fetched on the next cycle (#412).
 		if isTransientEmbedError(err) {
 			return 0, err
 		}
+		// Non-transient, multi-item, and not a local/config fatal: this is the
+		// poison-chunk case. Bisect the batch so only the offending input(s)
+		// fail instead of the whole HTTP batch (#399).
+		if len(validTasks) > 1 && !errors.Is(err, ErrFatal) {
+			return w.bisectEmbedBatch(ctx, indexKind, modelName, validTasks, labels, err)
+		}
+		// Single item, or a fatal batch error: mark failed (existing behavior).
 		category := string(store.ClassifyError(err))
 		reason := store.SanitizeReason(err.Error())
 		if mfErr := w.Source.MarkFailedWithCategory(ctx, labels, category, reason); mfErr != nil {
@@ -753,6 +775,34 @@ func (w *EmbeddingWorker) EmbedAndIndex(ctx context.Context, indexKind string, t
 	}
 
 	return len(labels), nil
+}
+
+// bisectEmbedBatch isolates the provider-rejected input(s) in a batch that
+// failed a non-transient, non-fatal embed by splitting it in half and
+// re-embedding each half through embedIndexBatch. Recursion narrows to the
+// individual poison chunk, which is marked failed (embedding_status=error)
+// while every healthy sibling still embeds and indexes (issue #399).
+//
+// Error precedence in the return value: a transient error surfacing from either
+// half means those chunks stayed PENDING, so it is propagated to make the run
+// loop back off and re-fetch them. An isolated non-transient failure is already
+// terminal (its chunk is marked failed) and need not propagate, so once the
+// poison is contained the batch reports success for the healthy remainder.
+// Returns the total number of chunks successfully indexed across both halves.
+func (w *EmbeddingWorker) bisectEmbedBatch(ctx context.Context, indexKind, modelName string, validTasks []model.ChunkTask, labels []uint64, batchErr error) (int, error) {
+	w.logf("embed batch of %d failed non-transiently (%s); bisecting to isolate poison chunk(s) so healthy siblings still embed [kind=%s]",
+		len(validTasks), store.SanitizeReason(batchErr.Error()), indexKind)
+	mid := len(validTasks) / 2
+	nLeft, errLeft := w.embedIndexBatch(ctx, indexKind, modelName, validTasks[:mid], labels[:mid])
+	nRight, errRight := w.embedIndexBatch(ctx, indexKind, modelName, validTasks[mid:], labels[mid:])
+	n := nLeft + nRight
+	if errLeft != nil && isTransientEmbedError(errLeft) {
+		return n, errLeft
+	}
+	if errRight != nil && isTransientEmbedError(errRight) {
+		return n, errRight
+	}
+	return n, nil
 }
 
 // Run starts a background loop that periodically calls RunOnce. A small
--- a/internal/index/embedding_worker.go
+++ b/internal/index/embedding_worker.go
@@ -739,26 +739,45 @@ func (w *EmbeddingWorker) embedIndexBatch(ctx context.Context, indexKind, modelN
 		if isTransientEmbedError(err) {
 			return 0, err
 		}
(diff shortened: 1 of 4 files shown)


yidakra:
Addressed all review threads in 195c19d. `make check` passes clean (CGO_ENABLED=0, golangci-lint v2.10.1, gocyclo ≤15).

**Embedding-worker bisection (`internal/index/embedding_worker.go`)** — the root of three threads was that `bisectEmbedBatch` only ever propagated *transient* errors and returned `nil` for everything else, silently swallowing anything non-transient in a bisected half. Reworked the error contract around a new `errChunkMarkedFailed` sentinel:

- **`ErrFatal` swallowed in bisection** (optibot, copilot) — Fixed. `embedIndexBatch` now propagates `ErrFatal` unwrapped (never sentinel-wrapped), and `mergeBisectErrors` gives it precedence over a sibling's transient error, so a local media/config failure stays loud and stops the worker (the #412 contract) even when it surfaces mid-bisection.
- **Bisection drops post-embed write errors** (greptile P1) — Fixed. Only the `errChunkMarkedFailed` sentinel (an isolated poison chunk already terminally marked failed) is absorbed; every other error — transient, `ErrFatal`, and post-embed write failures from `indexChunks`/`markEmbeddedWithRetry` — now propagates so the run loop backs off instead of tight-looping on PENDING chunks.
- **Auth errors explode into O(n) bisection** (copilot) — Fixed. Batches whose error classifies as `ErrorCategoryAuth` (401/403) skip bisection entirely and keep the existing whole-batch mark-failed behavior, since bisecting a corpus-wide auth rejection isolates nothing and just hammers the provider (1+2+4+…).

The #399 poison-isolation and #412 transient contracts are preserved (existing `TestEmbedAndIndex_*` still green). Added tests: `TestBisect_FatalErrorPropagates`, `TestAuthError_NotBisected`, `TestBisect_PostEmbedWriteErrorPropagates`.

**Hybrid degradation warning (`internal/retrieval/{service,hybrid}.go`)** (greptile P2) — Fixed. Replaced the `sync.Once` guard with an `atomic.Bool` that `SetHybridEnabled(true)` resets, so a hot-reload toggling hybrid off and back on re-arms the "store does not implement LexicalSearcher" warning instead of staying silent for the Service lifetime.

No threads rebutted.

PR #480, opened 2026-07-01
SECURITY: enforce ingest exclude/secret policy on on-demand transcribe/annotate (#407)

### LINKED ISSUE #407: SECURITY: on-demand transcribe/annotate bypasses ingest exclude policy + unbounded file read; annotate skips open_file's secret-scan
**Part of #395.** The on-demand audio-init path and the annotate/transcribe text path are not hardened to the same policy as open_file. (Traversal itself is NOT affected — root-containment is correctly enforced everywhere via Clean + `..`/abs reject + EvalSymlinks + Rel.)

## 1 [MEDIUM] On-demand transcribe bypasses `ingest.exclude` and reads unbounded
`internal/mcp/tools.go`: `handleTranscribeTool` → `lookupOrInitAudioDocumentForTool` → `initAudioDocumentOnDemand`. When the audio doc is not already indexed, it resolves any audio-extension path under root via `resolveDocumentPath` and ingests it:
```go
absPath, pathErr := s.resolveDocumentPath(normalizedRel)   // root-containment only
...
content, readErr := os.ReadFile(absPath)                    // no size bound
```
- **Exclude bypass:** `resolveDocumentPath` enforces only root containment — it does **not** consult `pathExcludes`/`ingest.exclude`, unlike open_file's `resolveFilePath` (which returns `ErrForbidden` on an excluded match). Because excluded files are never indexed, the on-demand branch is *always* taken for them — so an operator-excluded audio file under root (e.g. `private/*.mp3`) can be transcribed on demand and its content disclosed, plus it incurs STT/provider cost.
- **Unbounded read:** `os.ReadFile(absPath)` reads the whole file into memory with no cap (ingest's `MaxSizeBytes` is bypassed here; open_file uses `readFileBounded`). A large within-root audio file → daemon OOM.
- **Reach:** remote when `--public`; confined to within-root (no traversal).
- **Fix:** run the `pathExcludes` check inside `resolveDocumentPath` (or `initAudioDocumentOnDemand`), and bound the read.

## 2 [LOW-MEDIUM] Secret-scan asymmetry: annotate/transcribe skip the gate open_file enforces
open_file refuses to emit content matching `secret_patterns` (`hasSecretMatch` gates `openFileFromResolvedPath`/`openFileFromMetadata` → `ErrForbidden`). annotate (`sourceTextForAnnotation` → `readDocumentContent`) and transcribe return `annotation_text_preview` / transcript `segments` with **no** `hasSecretMatch` gate. A document open_file refuses (because it contains an AWS key/JWT/etc.) can still be exfiltrated via annotate. The doc must be indexed (not arbitrary-file), so this is a consistency/defense-in-depth gap, not a traversal escape. **Fix:** apply the same secret scan on the annotate/transcribe output paths.

## 3 [LOW] `open_file page` has no upper bound
`parsePageArg` enforces `page > 0` but no ceiling (k/max_chars/limit/clip-duration are all bounded). Impact is minimal (a huge page just misses → `DOC_TYPE_UNSUPPORTED`, no proportional allocation), but cap it for symmetry.

yidakra:
## Summary

Fixes the SECURITY issue in #407: the on-demand `dir2mcp_transcribe` /
`dir2mcp_annotate` tool handlers processed a file **without** applying the
corpus exclusion/secret-pattern policy that normal ingestion (and
`open_file`) enforce. A caller could therefore transcribe/annotate a file the
operator deliberately excluded (e.g. `private/*.mp3`, `**/.env`, `**/*.key`)
and disclose its content — plus the on-demand audio read was unbounded.

## Changes

- **Exclude-policy bypass (part 1).** `resolveDocumentPath` — the shared choke
  point for the on-demand transcribe/annotate read paths — now applies the same
  path-exclusion policy ingestion uses (`ingest.MatchesAnyPathExclude` over
  `cfg.PathExcludes`), checked against **both** the requested rel path and the
  symlink-resolved real path (zip-slip / symlink-escape into an excluded
  target). Excluded → `model.ErrForbidden` → canonical `PERMISSION_DENIED`.
- **Unbounded read (part 1).** The on-demand audio init now bounds the read to
  the ingest file-size cap (`IngestMaxFileMB`) instead of an unbounded
  `os.ReadFile`, refusing oversize files with `FILE_TOO_LARGE` rather than
  risking daemon OOM. A `readBoundedFile` guard also catches a file that grows
  between stat and read.
- **Secret-scan asymmetry (part 2).** transcribe/annotate now apply the same
  `secret_patterns` gate `open_file` enforces before emitting derived text
  (transcript segments, OCR/annotation previews), closing a defense-in-depth
  exfiltration gap. The offending text is never logged.
- **open_file page cap (part 3).** `parsePageArg` gains an upper bound for
  symmetry with the other numeric limits.

Reuses the existing `internal/ingest` exclusion/secret helpers
(`MatchesAnyPathExclude`, `CompileSecretPatterns`, `HasSecretMatch`) rather
than duplicating logic. Existing tool/error contracts and canonical codes are
preserved.

## Tests

New `tests/mcp/tool_exclude_policy_test.go` covers:
- transcribe refuses an excluded audio path (on-demand branch),
- transcribe refuses an oversize audio file (`FILE_TOO_LARGE`),
- annotate refuses an excluded document,
- annotate refuses secret-matching content (and does not echo the secret),
- transcribe refuses a transcript containing a secret (and does not echo it).

`make check` passes (gofmt, go vet, golangci-lint v2 → 0 issues, full test
suite, python unittests, build).

Closes #407

<h3>Greptile Summary</h3>

This PR closes a security gap in the on-demand `dir2mcp_transcribe` and `dir2mcp_annotate` tool handlers: until now, neither path applied the corpus path-exclusion or `secret_patterns` policy that normal ingestion and `open_file` already enforced. A caller could therefore transcribe or annotate any deliberately excluded file (e.g., `private/*.mp3`, `**/.env`) or exfiltrate secrets via derived text.

- **Exclude-policy bypass** is fixed by adding `ingest.MatchesAnyPathExclude` checks to `resolveDocumentPath` (the shared choke point), covering both the requested relative path and the symlink-resolved real path to guard against zip-slip / symlink-escape into excluded targets.
- **Unbounded audio read** is fixed by replacing `os.ReadFile` with a new `readBoundedFile` helper capped at `IngestMaxFileMB`, with a TOCTOU guard that catches files that grow between stat and read.
- **Secret-scan asymmetry** is closed by `refuseIfSecretContent` (pattern compilation cached via `sync.Once`) applied after transcript/OCR derivation; when a secret fires, `PurgeTranscriptCache`/`PurgeOCRCache` are called immediately so the refused text is not left on disk.
- **`open_file` page cap** is added to `parsePageArg` for symmetry with other numeric bounds.

<h3>Confidence Score: 5/5</h3>

Safe to merge. The changes are narrowly scoped security fixes that close documented exfiltration paths without altering existing tool contracts or error codes.

Every changed code path has been traced: resolveDocumentPath now enforces PathExcludes at both the raw normalized path and the symlink-resolved path, covering transcribe, annotate, and media-clip reads. readBoundedFile provides TOCTOU-safe bounded reads. The secret gate is correctly applied after derivation with immediate cache purge. The sync.Once pattern-compilation fix, the prior-review cache-before-gate concern for both transcript and OCR paths, and the open_file page-cap are all correctly addressed. Integration tests verify each scenario including disk non-persistence of refused secrets.

No files require special attention. All four changed files implement their stated purpose correctly.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| internal/mcp/tools.go | Core security fixes: adds exclusion-policy check to resolveDocumentPath (both pre- and post-symlink resolution), bounded audio read via readBoundedFile, refuseIfSecretContent with sync.Once pattern compilation, OCR/transcript secret gate with immediate cache purge on match, and open_file page cap. All paths through readDocumentContent now enforce PathExcludes. |
| internal/ingest/service.go | Adds PurgeOCRCache and PurgeTranscriptCache public helpers. Both methods use the same active-derivation cache key (ocrCacheKey/transcriptCacheKey) that the underlying write paths use, ensuring the purge targets the exact file(s) just written. PurgeTranscriptCache also removes the .words.json word-timing sidecar. |
| internal/mcp/server.go | Adds secretPatternOnce sync.Once, secretPatterns []*regexp.Regexp, and secretPatternErr error fields to Server to cache compiled secret-pattern regexes across requests, addressing the per-call recompilation noted in prior review. |
| tests/mcp/tool_exclude_policy_test.go | New integration tests covering all three fix areas: excluded path → PERMISSION_DENIED for transcribe and annotate, oversize audio → FILE_TOO_LARGE, secret-matching content → PERMISSION_DENIED without echoing the secret value, and disk-persistence verification that PurgeTranscriptCache removes the cache entry on a secret match. |

<h3>Flowchart</h3>

<a href="#gh-light-mode-only">

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[dir2mcp_transcribe / dir2mcp_annotate request] --> B[lookupDocumentForTool\nor initAudioDocumentOnDemand]
    B --> C{resolveDocumentPath}
    C -- MatchesAnyPathExclude\nnormalized rel path --> D[ErrForbidden → PERMISSION_DENIED]
    C -- EvalSymlinks --> E{MatchesAnyPathExclude\nsymlink-resolved rel path}
    E -- excluded --> D
    E -- allowed --> F[readDocumentContent\nreturns file bytes]
    F --> G{info.Size > ingestMaxFileBytes?}
    G -- YES --> H[FILE_TOO_LARGE]
    G -- NO --> I[readBoundedFile\nLimitReader maxBytes+1]
    I -- tooLarge=true --> H
    I -- OK --> J[ReadOrComputeTranscript\nor ReadOrComputeOCR\nwrites to cache]
    J --> K{refuseIfSecretContent\nsecretPatternOnce.Do}
    K -- match --> L[PurgeTranscriptCache\nor PurgeOCRCache\ndelete from disk]
    L --> D
    K -- no match --> M[Return transcript/OCR text\nto caller]
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
flowchart TD
    A[dir2mcp_transcribe / dir2mcp_annotate request] --> B[lookupDocumentForTool\nor initAudioDocumentOnDemand]
    B --> C{resolveDocumentPath}
    C -- MatchesAnyPathExclude\nnormalized rel path --> D[ErrForbidden → PERMISSION_DENIED]
    C -- EvalSymlinks --> E{MatchesAnyPathExclude\nsymlink-resolved rel path}
    E -- excluded --> D
    E -- allowed --> F[readDocumentContent\nreturns file bytes]
    F --> G{info.Size > ingestMaxFileBytes?}
    G -- YES --> H[FILE_TOO_LARGE]
    G -- NO --> I[readBoundedFile\nLimitReader maxBytes+1]
    I -- tooLarge=true --> H
    I -- OK --> J[ReadOrComputeTranscript\nor ReadOrComputeOCR\nwrites to cache]
    J --> K{refuseIfSecretContent\nsecretPatternOnce.Do}
    K -- match --> L[PurgeTranscriptCache\nor PurgeOCRCache\ndelete from disk]
    L --> D
    K -- no match --> M[Return transcript/OCR text\nto caller]
```

</a>

<sub>Reviews (2): Last reviewed commit: ["review: address PR #480 comments"](https://github.com/dirstral/dir2mcp/commit/464a9e8b8346398e626ec645de9f6025543c082c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41247773)</sub>

### DIFF
--- a/internal/ingest/service.go
+++ b/internal/ingest/service.go
@@ -2905,6 +2905,18 @@ func (s *Service) ReadOrComputeOCR(ctx context.Context, doc model.Document, cont
 	return s.readOrComputeOCR(ctx, doc, content)
 }
 
+// PurgeOCRCache removes the OCR cache entry for the given content, using the
+// same active-extractor key readOrComputeOCR writes under. Callers that gate the
+// returned OCR text (e.g. the MCP secret-pattern gate) use this so a refused
+// result is never left persisted in {StateDir}/cache/ocr. Missing files are
+// ignored.
+func (s *Service) PurgeOCRCache(content []byte) {
+	cachePath := filepath.Join(s.cfg.StateDir, "cache", "ocr", s.ocrCacheKey(content)+".md")
+	if err := os.Remove(cachePath); err != nil && !errors.Is(err, os.ErrNotExist) {
+		s.getLogger().Printf("purge ocr cache: %v", err)
+	}
+}
+
 // TranscriptLangSuffix returns a safe filename suffix for the given language,
 // used to key the transcript cache by content+language. Empty language returns
 // an empty suffix so language-unaware callers share the same cache file.
@@ -3043,6 +3055,22 @@ func (s *Service) writeCachedWords(path string, words []model.TimedWord) {
 // ReadOrComputeTranscript exposes transcript cache lookup/computation for tests.
 func (s *Service) ReadOrComputeTranscript(ctx context.Context, doc model.Document, content []byte, language string) (string, error) {
 	return s.readOrComputeTranscript(ctx, doc, content, language)
+}
+
+// PurgeTranscriptCache removes the transcript cache entry (and its word-timing
+// sidecar) for the given content+language, using the same active-STT key
+// readOrComputeTranscriptWithWords writes under. Callers that gate the returned
+// transcript (e.g. the MCP secret-pattern gate) use this so refused text is
+// never left persisted in {StateDir}/cache/transcribe. Missing files are
+// ignored.
+func (s *Service) PurgeTranscriptCache(content []byte, language string) {
+	cacheDir := filepath.Join(s.cfg.StateDir, "cache", "transcribe")
+	base := s.transcriptCacheKey(content) + TranscriptLangSuffix(language)
+	for _, p := range []string{base + ".txt", base + ".words.json"} {
+		if err := os.Remove(filepath.Join(cacheDir, p)); err != nil && !errors.Is(err, os.ErrNotExist) {
+			s.getLogger().Printf("purge transcript cache: %v", err)
+		}
+	}
 }
 
 // flattenJSONForIndexing walks an arbitrary JSON-like structure and
--- a/internal/mcp/server.go
+++ b/internal/mcp/server.go
@@ -18,6 +18,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime/debug"
 	"sort"
 	"strings"
@@ -122,6 +123,14 @@ type Server struct {
 	execKeyMu map[string]*keyMutex
 
 	eventEmitter func(level, event string, data interface{})
+
+	// secretPatterns is the compiled secret_patterns slice used by
+	// refuseIfSecretContent. Compilation is deferred to first use and performed
+	// at most once via secretPatternOnce so annotate/transcribe requests reuse
+	// the same []*regexp.Regexp instead of recompiling every call.
+	secretPatternOnce sync.Once
+	secretPatterns    []*regexp.Regexp
+	secretPatternErr  error
 
 	// extractSegment cuts [startMS, endMS) from a local media file and returns
 	// the container bytes for dir2mcp_open_media_clip (SPEC §15.11). It defaults
--- a/internal/mcp/tools.go
+++ b/internal/mcp/tools.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"log"
 	"math"
@@ -27,6 +28,12 @@ import (
 	"github.com/dirstral/dir2mcp/internal/provider"
 	"github.com/dirstral/dir2mcp/internal/providerfactory"
 )
+
+// maxOpenFilePage caps the open_file page argument so a caller cannot request
+// an absurd page index (issue #407 part 3). Real documents stay well under
+// this; the ceiling only bounds pathological input for symmetry with the other
+// numeric limits (k/max_chars/clip-duration).
+const maxOpenFilePage = 1_000_000
 
 const (
 	defaultEmbedTextModel = "mistral-embed"
@@ -1080,6 +1087,12 @@ func (s *Server) handleAnnotateTool(ctx context.Context, args map[string]interfa
 	if toolErr != nil {
 		return toolCallResult{}, toolErr
 	}
+	// Gate the source text against secret_patterns, matching open_file (issue
+	// #407). The audio branch is already gated inside the transcript path; this
+	// covers OCR/raw-text documents whose content open_file would refuse.
+	if gate := s.refuseIfSecretContent(sourceText); gate != nil {
+		return toolCallResult{}, gate
+	}
 	if strings.TrimSpace(sourceText) == "" {
 		return toolCallResult{}, &toolExecutionError{Code: "ANNOTATE_FAILED", Message: "no source text available for annotation", Retryable: false}
 	}
@@ -1797,6 +1810,9 @@ func parsePageArg(args map[string]interface{}) (page int, hasPage bool, toolErr 
 	if p <= 0 {
 		return 0, false, &toolExecutionError{Code: "INVALID_FIELD", Message: "page must be > 0", Retryable: false}
 	}
+	if p > maxOpenFilePage {
+		return 0, false, &toolExecutionError{Code: "INVALID_FIELD", Message: fmt.Sprintf("page must be <= %d", maxOpenFilePage), Retryable: false}
+	}
 	return p, true, nil
 }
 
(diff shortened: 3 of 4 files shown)


yidakra:
Addressed the review comments (commit 464a9e8):

**Secret-pattern regex recompiled per request** (optibot, copilot, greptile P2) — FIXED. `refuseIfSecretContent` now compiles `secret_patterns` at most once via a `sync.Once` on `Server` (`secretPatterns`/`secretPatternErr`) and reuses the compiled `[]*regexp.Regexp` on every annotate/transcribe call. Also moves compile-error surfacing off the hot path.

**Audio transcript re-scanned on annotate** (copilot) — FIXED. The gate is now applied once at the source: the audio branch gates inside `ensureTranscriptForAudioDoc`, and the OCR/raw branches gate inside `sourceTextForAnnotation`. Removed the duplicate `refuseIfSecretContent` call in `handleAnnotateTool`, so audio transcripts are no longer double-scanned.

**Secret written to transcript cache before the gate** (greptile P1, security) — FIXED. When the gate refuses a transcript, `ensureTranscriptForAudioDoc` now calls the new `ingest.Service.PurgeTranscriptCache` (using the same active-STT key `ReadOrComputeTranscript` wrote under) to remove the `.txt` and `.words.json` entries, so refused secrets no longer linger in `{StateDir}/cache/transcribe`.

**OCR cache written before the gate** (greptile P1, security) — FIXED. Symmetric fix: `sourceTextForAnnotation` gates the OCR text and, on a match, calls the new `ingest.Service.PurgeOCRCache` to remove the `{StateDir}/cache/ocr/{hash}.md` entry.

Added `TestMCPTranscribe_PurgesSecretTranscriptCache`, which walks the cache dir and asserts the secret is not left on disk after a refusal (verified it fails without the purge).

**`readBoundedFile` duplicates `retrieval.readFileBounded`** (optibot, non-blocking) — declined. The two helpers have deliberately different semantics: `mcp.readBoundedFile` reads `maxBytes+1` to distinguish a file exactly at the cap from one over it (`tooLarge` sentinel), takes `int64`, and is used for the byte-bounded audio read; `retrieval.readFileBounded` takes `int`, treats `<=0` as unbounded, and reports truncation via `len==maxBytes` (an intentional heuristic for its char-bounded `open_file` path). `mcp` has no dependency on `retrieval`; sharing would either pull a `retrieval` import in for a ~10-line helper or force retrieval's exact-length reporting onto the audio path. Kept separate.

`make check` passes (fmt/vet/lint v2.10.1/gocyclo ≤15/tests).

Issue #493, opened 2026-07-02
docs: first-class 'fully local / no-egress' setup path + doctor egress check (defaults imply Mistral cloud)


yidakra:
## Summary
The fully self-hosted ("no data leaves the host") provider configuration is supported and first-class (SPEC §8.5, `docs/dual-machine-deployment.md`), but it is not discoverable from the default onboarding. `.env.example` and the quickstart show only **cloud** providers (`MISTRAL_BASE_URL=https://api.mistral.ai`, ElevenLabs), so an operator following the happy path silently sends corpus content to third parties — even when their requirements forbid it.

## Why it matters
Users of this tool frequently have data-residency / compliance constraints (e.g. a media archive that must be processed on-prem). Today the path of least resistance is the cloud pipeline; the local path requires reading into `providers:` / SPEC §8.5 to discover. The capability exists — the docs just don't lead there.

## Requested changes
1. A prominent **"Fully local / no-egress" quickstart** in the README (and a commented `.env.local.example` or a `providers:` snippet) showing STT via `kind: whisper` + a self-hosted `kind: openai` chat/translation profile, with `stt_provider: whisper` and no cloud `base_url`s.
2. A **`dir2mcp doctor` egress check** — report whether any resolved provider `base_url` is a public/third-party host, so operators can verify "nothing leaves the host" rather than infer it. (Complements the existing per-extractor status reporting.)
3. Cross-link from the cloud `.env.example` to the local recipe so the tradeoff is visible at the point of configuration.

## Relevant references
- `docs/dual-machine-deployment.md` (self-hosted provider contract)
- README `providers:` block (`kind: whisper`, self-hosted `kind: openai`)
- `.env.example` (currently cloud-only)

Issue #444, opened 2026-06-26
retrieval: LooksLikeCodeQuery mis-routes plain English to the code index; query expansion fires up to ~9 serial LLM calls/query with no max_tokens, no cache, no budget


yidakra:
**Part of #395.** Query-expansion (HyDE + cross-lingual + adaptive gate) correctness/cost gaps.

## F1 [MED] `LooksLikeCodeQuery` mis-routes ordinary English prose to the CODE index (verified)
`service.go:29-30`: `codeKeywordRe = \b(func|class|package|import|return|if|for|while|switch|case)\b` and `codePunctRe = [(){}\[\];]`; the shortcut is `if hasKw && hasPunct → code` (`:2534`). The keyword set is full of common English words and the punct set is near-universal, so ordinary questions classify as code → routed to the code index (`auto` branch, `:942`) → poor/empty recall on a text corpus:
- "How do I import data (CSV) into the system?" (`import` + `(`)
- "What is the case for using X; and why?" (`case` + `;`)
- "Which class did she attend (and when)?" (`class` + `(`)
It also feeds the adaptive gate's `codey` signal and is re-run per cross-lingual variant. **Fix:** require a stronger code signal (backticks/fences, file extensions, or keyword *adjacent* to code punctuation), not "any keyword anywhere + any punct anywhere."

## F2 [MED, cost] Up to ~9 serial LLM calls + ~10 retrievals per query, unbounded in aggregate
With HyDE + cross-lingual on, a single `ask`/`search` issues serially: 1 HyDE generation + 2 retrievals, then **up to 8** translation generations (`crossLingualMaxVariants=8`, `crosslingual.go:21,103`) each with its own retrieval + rerank. With `target-langs: auto` the corpus's detected-language set drives this automatically — an operator who never chose 8 languages still gets 8 translation calls. No aggregate budget; translations not parallelized. **Fix:** parallelize + cap aggregate LLM calls/latency per query.

## F3 [LOW-MED] Generation has NO `max_tokens` → cost amplification
[...]

PR #501, opened 2026-07-03
fix(openai): bound chat completions with a default max_tokens (#500)

### LINKED ISSUE #500: ingest/translate: media-translate chat request sends no max_tokens → runaway generation → fatal OPENAI_FAILED on self-hosted/CPU LLM backends
## Summary
The media **translation** step (and, it appears, the per-cue chat calls behind it) sends the chat request with **no `max_tokens` cap**. Against a fast cloud model this is invisible, but against a self-hosted / CPU-bound OpenAI-compatible backend the model keeps generating well past the intended subtitle line, so a single translation exceeds the client timeout and the file fails fatally with `OPENAI_FAILED`. This makes the fully-local / no-egress path (the one #493 documents) effectively unusable for translation on realistic local hardware.

## Environment
- `dir2mcp-full` 0.9.9 (Homebrew / Linuxbrew, Linux, 2× NVIDIA A2)
- Fully-local config (no cloud base_urls):
  - STT: self-hosted faster-whisper `large-v3`, `kind: whisper`
  - chat/translate + embed: Ollama (`kind: openai`, OpenAI-compatible), `qwen2.5:3b-instruct` / `bge-m3`
  - `media_translate_enabled: true`, `media_translate_target_langs: [en]`
- Sample: a 3-min 16 kHz mono WAV (51 speech segments, Russian).

## What happens
1. STT succeeds — 51 segments produced, `export --format srt` yields a correct original-language SRT.
2. The inline translate step then hangs and the file eventually fails with `OPENAI_FAILED` (fatal for that file). The embed queue also sits at `pending=51`.

## Root cause (reproduced)
The chat/translate request omits `max_tokens`. On a local/CPU backend a single cue's completion runs away (the small model continues/repeats past the one subtitle line) until the socket times out.

Standalone reproduction against the **same** model + endpoint:
- No `max_tokens`: one translate call ran **9m38s** before the client socket timed out (`TimeoutError: timed out`).
- With `"max_tokens": 200`: each call returns in ~30s reliably; all 51 cues then translate cleanly.

So the failure isn't the backend being "wrong" — it's an unbounded generation request meeting a bounded client timeout.

## Impact
- The self-hosted-LLM translation path — the deployment the README's no-egress guidance implies (see #493) and the whole point of "locally deployed LLMs" — cannot complete on ordinary local hardware.
- One slow cue fails the **entire file**, not just that cue.

## Suggested fix
- Send a bounded `max_tokens` on the translate (and per-cue chat) requests, sized from the input length; optionally expose `media_translate_max_tokens`.
- Prefer a per-request timeout + retry/skip of the single cue over failing the whole file.

Related: #473 (distinct `TRANSLATE_FAILED` code), #444 (query-expansion also issues no-`max_tokens` LLM calls) — same class of bug, different call site.

yidakra:
Fixes #500.

## Problem
The OpenAI-compatible `Generate` path (`internal/openai`) marshaled `/chat/completions` with only `model` + `messages` — **no `max_tokens`**. Against a fast cloud model this is invisible, but against a small self-hosted / CPU chat backend the model can generate far past the requested output — e.g. a subtitle-line translation that never stops — until `GenerationTimeout` fires and the file fails fatally with `OPENAI_FAILED`. That makes the fully-local / no-egress translation path (the "locally deployed LLMs" setup, #493) unusable on realistic local hardware, and one runaway cue fails the **whole file**.

Reproduced against Ollama `qwen2.5:3b-instruct` on CPU: a no-cap translate call ran **9m38s** before the client socket timed out; adding a `max_tokens` cap made each call return reliably (~30s) and the full 51-cue clip translated cleanly.

## Fix
- Add a client-level `GenerationMaxTokens` (default **1024**), mirroring the existing `GenerationTimeout` default pattern — a client-level default, not a new user-config field.
- Always emit `max_tokens` on the chat request; `<= 0` falls back to the default so it is **never sent unbounded**.
- Applies to every `model.Generator.Generate` caller — translation and query expansion (same class as #444).

A finite cap converts an unbounded runaway into a bounded request. 1024 is generous for translations and typical answers; operators on very slow backends can lower `Client.GenerationMaxTokens`.

## Tests
`tests/openai`: `TestGenerate_SendsBoundedMaxTokens` asserts the request carries `max_tokens` (default 1024, and an explicit override is honored). `go build ./...`, `go vet`, and the `openai` suites pass.

## Notes / follow-ups
- A per-call or per-config cap (e.g. a tighter `media_translate_max_tokens` for the translate path specifically, vs. `ask` answer synthesis) would be a natural refinement, but is out of scope here — this PR removes the unbounded-runaway failure with a single safe default.

## Summary by CodeRabbit

* **New Features**
  * Added bounded generation support so translation and text generation can cap output length per request, with sensible fallbacks when a cap isn’t provided or isn’t supported.

* **Bug Fixes**
  * Prevented unbounded chat completion generation by always sending a valid maximum token value in requests.

* **Tests**
  * Added coverage to verify translation uses the tighter per-line cap (512) when available.
  * Added OpenAI client tests to confirm `max_tokens` is always included and honors default/per-call override behavior.

### DIFF
--- a/internal/ingest/translate.go
+++ b/internal/ingest/translate.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+
+	"github.com/dirstral/dir2mcp/internal/model"
 )
 
 // readOrComputeTranslation returns the source transcript translated into
@@ -88,6 +90,14 @@ func (s *Service) translateTranscriptText(ctx context.Context, sourceText, targe
 	return strings.Join(out, "\n"), nil
 }
 
+// translateLineMaxTokens caps a single translated transcript line. One source
+// segment maps to one short output line, so a tight bound is plenty here; it
+// keeps the #500 runaway path tight on chat backends that respect max_tokens
+// WITHOUT lowering the generous default that ask/annotate rely on. It is applied
+// only when the translator implements model.BoundedGenerator; otherwise the call
+// falls back to Generate (the provider's own default cap still bounds it).
+const translateLineMaxTokens = 512
+
 // translateLine translates a single line of transcript text into targetLang via
 // the chat Generator. Empty/whitespace input short-circuits to empty so the chat
 // provider is never called for a marker-only line. The prompt asks for the
@@ -98,6 +108,9 @@ func (s *Service) translateLine(ctx context.Context, text, targetLang string) (s
 		return "", nil
 	}
 	prompt := buildTranslatePrompt(text, targetLang)
+	if bg, ok := s.translator.(model.BoundedGenerator); ok {
+		return bg.GenerateWithMaxTokens(ctx, prompt, translateLineMaxTokens)
+	}
 	translated, err := s.translator.Generate(ctx, prompt)
 	if err != nil {
 		return "", err
--- a/internal/model/interfaces.go
+++ b/internal/model/interfaces.go
@@ -255,6 +255,19 @@ type Generator interface {
 	Generate(ctx context.Context, prompt string) (string, error)
 }
 
+// BoundedGenerator is an OPTIONAL capability a Generator MAY implement to cap a
+// single completion's output tokens for THIS call, without changing the
+// generous default the generator uses for unbounded callers (answer synthesis,
+// annotation). A caller with a known-short output (e.g. one translated
+// transcript line) type-asserts its Generator against this and, if satisfied,
+// passes a tight cap; a Generator that does not implement it falls back to
+// Generate and behaves exactly as before. A maxTokens <= 0 MUST behave like
+// Generate (use the generator's own default).
+type BoundedGenerator interface {
+	Generator
+	GenerateWithMaxTokens(ctx context.Context, prompt string, maxTokens int) (string, error)
+}
+
 // Reranked is one rescored candidate: Index is the position in the
 // documents slice passed to Rerank; RelevanceScore is the provider's
 // relevance (higher = better). Results are returned best-first.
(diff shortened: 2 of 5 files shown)


yidakra:
Reviewed this and pushed a fix to the branch (commit `63fbb51`) addressing the one blocker — hope that's helpful, revert if you'd rather take it yourself:

**The global `max_tokens=1024` truncated real output on `kind:openai` backends** — it capped *every* `Generate` caller, including `ask` answer synthesis (silent mid-answer truncation) and `annotate` (JSON cut off → `ANNOTATE_FAILED`, a functional regression) — and was 4× tighter than the anthropic sibling's 4096.

Fix (both parts, the interface threading was clean so nothing deferred):
- **Default 1024 → 4096** (matches `internal/anthropic`; still bounds the #500 runaway — any finite cap does — but no truncation of answers/annotations).
- **Per-call tight cap** via a new optional `model.BoundedGenerator` (`GenerateWithMaxTokens`, mirroring the `StructuredTranscriber` optional-capability pattern — no `Generator` signature change, non-openai implementers untouched). The ingest translate path (`translateLine`) passes `max_tokens=512` (a transcript line is short); `ask`/`annotate`/cross-lingual stay on 4096.
- Dropped the misleading `omitempty` on the `max_tokens` tag (value is always > 0 after clamping).

Added tests for the per-call cap (512 sent; 0/negative → 4096) and the translate path. `make cyclo`/`golangci-lint`/targeted tests all green. The mistral/gemini/cohere unbounded paths (#444/#473 class) are still out of scope — worth a separate follow-up.

Issue #524, opened 2026-07-03
subtitle: broadcast-grade cue re-segmentation from word timings (CPS/duration/line-length norms)


yidakra:
`subtitle.BuildCues` emits exactly one cue per stored transcript chunk, i.e. one per Whisper *segment*. That reads fine as a transcript but is not broadcast-legible: a single segment routinely runs 15–30 s and 150+ characters, so the cue violates every standard subtitle norm (≤~6–7 s on screen, ≤2 lines × 42 chars, ≤17 chars/sec reading speed).

Measured on the RFE/RL pilot (30-min Russian interview, `large-v3`, `task=translate`), current segment-level export vs. word-timing re-segmentation:

| metric | `BuildCues` (today) | word-timing re-seg |
|---|---|---|
| cues | 241 | 391 |
| max cue duration | **30.0 s** | 6.0 s |
| cues > 7 s | **99** | 0 |
| lines > 42 chars | **175** | 11 |
| cues > 2 lines | 0 | 0 |
| cues > 17 CPS | 12* | 26 (of which **21 are speech-rate-bound**) |

\*The segment-level CPS count is deceptively low only because 30 s mega-cues dilute the rate; the text is still unreadable.

The 21 residual over-CPS cues are genuine fast speech (18–30 CPS with no adjacent silence to borrow) — fixable only by editorial text condensation, which changes wording and is deliberately **out of scope** here (track separately if wanted).

## The data is already there
No new capture path is needed. Word timings already flow end-to-end (#252):
- whisper client parses `verbose_json` words into `model.TranscriptResult.Words`;
- `internal/ingest/represent.go` attaches them to each chunk's `Span.Words` (offset-shifted for chunked media);
- `internal/store/sqlite_store.go` persists them in `extra_json` as `{t,d,w}` and reads them back (`wordsFromExtraJSON`).

So `subtitle.TranscriptChunk.Span.Words` is populated on any whisper-`verbose_json` transcript. A broadcast builder consumes data that is already persisted, playback-ordered, and in absolute ms.

## Proposed design
New `internal/subtitle/broadcast.go`:

```
func BuildBroadcastCues(chunks []TranscriptChunk, opts BroadcastOpts) []Cue
```

- Flatten `chunk.Span.Words` across chunks into a time-ordered word stream.
[...]

PR #533, opened 2026-07-04
feat(subtitle): broadcast cue segmentation from word timings (#524)

yidakra:
## What

Adds an opt-in **broadcast subtitle segmentation** mode that re-segments a transcript's per-word timings into broadcast-legible cues, closing #524.

`subtitle.BuildBroadcastCues` consumes the already-persisted `model.Span.Words` (spec §8.6.1) and produces cues that obey standard subtitle norms:
- ≤ 6 s duration, ≤ 2×42 characters
- breaks at sentence ends and speech pauses
- buys display time from surrounding silence (end-extend, then a small lead-in) without ever overlapping a neighbour
- two-line balanced wrap

Cue **text** is the verbatim word tokens re-joined — only cue boundaries and timing are computed, never the words. Because `whisperapi` trims each word token, spacing is rebuilt with punctuation-aware detokenization rather than relying on whisper's native leading spaces.

## Config

New enum `media.subtitles.segmentation: chunk|broadcast` (default `chunk`), validated in `validateMediaSubtitles` and plumbed like the other `media.subtitles.*` keys. The VTT/SRT export path selects the builder via `buildCuesForSegmentation` and **falls back to `BuildCues` when a transcript carries no word timings**, so behavior is unchanged unless `broadcast` is explicitly selected. TTML/SMIL keep chunk segmentation (their bilingual alignment is out of scope here).

## Tests

- Unit: duration cap, pause break, min-duration extension without overlap, wrap + detokenization, empty-words fallback.
- Config: validation, case/whitespace normalization, empty→default, unknown-value rejection, save/load round-trip, nested-YAML apply.
- End-to-end: a 20 s word-timed chunk splits into capped cues under `broadcast` while the default renders one cue.

## Checklist
- [x] `make check`-level gates pass locally (build, vet, golangci-lint 0 issues, gocyclo -over 15 clean, gofmt)
- [x] `coderabbit review` run via CLI — **2 major findings addressed** in 94ff11f (preserve speaker metadata across broadcast cues; don't gate natural breaks on spoken duration)

## Summary by CodeRabbit

* **New Features**
  * Added a new subtitle segmentation option for exports, including a broadcast-style mode for more readable SRT/VTT cues.
  * Subtitle exports now better preserve speaker changes, pauses, and line wrapping for clearer captions.

* **Bug Fixes**
  * Improved word filtering so removed phrases no longer reappear in exported subtitle text.
  * Fixed subtitle export behavior for filtered transcripts and long cues, keeping output within expected duration and line-length limits.


yidakra:
Reviewed this and pushed a fix to the branch (commit `8bf14ca`) for the two blockers — revert if you'd rather handle them yourself:

**1. Hard-rule: broadcast segmentation corrupted spaceless scripts.** `appendBroadcastToken` injected an ASCII space between tokens, so for CJK/Japanese/Thai (Whisper emits per-char pieces, no inter-word spaces) `你好世界` became `你 好 世 界` — text corruption, contradicting the 'text is never modified' claim; those lines also never wrapped and used rune-not-cell width. Took the **minimal, low-risk fallback** (not a full east-asian-width rewrite): `wordsNeedSpaceJoining` detects ≥30% spaceless runes (`unicode.Han`/Hiragana/Katakana/Thai; Hangul excluded since Korean uses spaces) → `BuildBroadcastCues` returns nil → caller falls back to `BuildCues`, which preserves stored chunk text verbatim. Latin/Cyrillic broadcast path unchanged. Documented the limitation.

**2. `media.filter_words` leak.** Broadcast cues rebuild from `Span.Words`, but `attachWordsToTimeSpans` attached the *full* word list — filtered phrases were never removed from `Span.Words`, so watermark/credit phrases stripped at ingest re-appeared in broadcast subtitles (the export-time `FilterCues` backstop is fragile: re-segmentation/wrap splits a phrase and defeats the substring match). Fixed at the source: `filterTimedWords` drops filtered words from `Span.Words` at ingest (new `WordFilter.FilterTokens`), so any word-consumer is clean. Default chunk path (uses already-filtered `Text`) unchanged.

Tests added for both (CJK verbatim-fallback, Cyrillic still segments, straddling-phrase no-leak). `make cyclo`/`golangci-lint`/subtitle+ingest+cli tests all green.

**Left as-is (your call):** the 8 broadcast knobs from #524 are still hardcoded consts — a scope decision, not a defect, so I didn't touch it. Might be worth plumbing so a CJK operator could tune `max_line` if you later add cell-width support.

PR #534, opened 2026-07-04
feat(subtitle): export-time cue cleaning — glossary, repetition-collapse, URL drop

yidakra:
## What

Ports the pilot's `clean_srt.py` cue-hygiene passes into the dir2mcp Go export path as three **additive, opt-in** filters, applied on VTT/SRT export after `media.filter_words`. Complements the broadcast segmentation in #533 / #524.

| Config | Effect | Default |
|---|---|---|
| `media.subtitles.glossary: ["pattern=>replacement", …]` | Editorial term **replacement** — case-insensitive, ASCII-word-bounded regexp, so one entry absorbs spelling variants (`Aju?bei=>Adzhubei`). Distinct from `filter_words` (which deletes). | empty (off) |
| `media.subtitles.collapse_repeats: N` | Drop the Nth-and-later cue in a run of identical consecutive cues (whisper repetition-collapse artifact), keeping the first N-1. `< 2` disables. | `0` (off) |
| `media.subtitles.drop_urls: bool` | Drop cues that are hallucinated URLs / bare domains / credit lines. | `false` |

All default off → export is byte-identical unless configured.

## Design

- New `subtitle.Glossary` + `subtitle.CleanCues` live beside the existing `WordFilter`, run via `subtitle.CleanCues` in `renderTranscriptExport`.
- The glossary grammar is validated **at config load** (`subtitle.NewGlossary` is the single source of truth), so a malformed `pattern=>replacement` entry or bad regexp fails fast with `CONFIG_INVALID` rather than at export. `collapse_repeats` is guarded non-negative.
- Config plumbed at every `media.subtitles.*` site (runtime field, fileConfig, defaults, snapshot, apply, alias, accessor/list-append, serialize, validate).
- Scope: VTT/SRT export only (TTML/SMIL keep their own path), and export-only — ingest/embeddings are untouched.

## Tests

- Unit: `Glossary` (whole-word / case-insensitive / regex-variant / error cases) and `CleanCues` (URL drop, collapse run math, glossary, inactive no-op).
- Config: defaults-off, save/load round-trip, nested-YAML apply, fail-fast validation (bad glossary entry, invalid regexp, negative threshold).
- End-to-end: `export` command applies all three passes with config, and is a no-op without it.

## Checklist
- [x] `make check`-level gates pass locally (build, vet, golangci-lint 0 issues, gocyclo -over 15 clean, gofmt)
- [x] `coderabbit review` run via CLI against `main` — **no findings**

## Summary by CodeRabbit

* **New Features**
  * Added optional subtitle cleanup during export, including glossary-based text replacements, repeat collapsing, and URL/credit-line removal.
  * New subtitle export settings can now be saved and loaded in configuration.

* **Bug Fixes**
  * Export now fails fast when subtitle glossary settings are invalid, preventing broken exports.
  * Cleaned subtitle cues are re-numbered correctly after removals.

### DIFF
--- a/internal/cli/export.go
+++ b/internal/cli/export.go
@@ -79,7 +79,19 @@ func (a *App) runExport(ctx context.Context, global globalOptions, args []string
 	}
 
 	filter := subtitle.NewWordFilter(cfg.MediaFilterWords)
-	rendered, code := a.renderTranscriptExport(ctx, global, ts, opts, filter)
+	// The glossary was already validated at config load, so a parse error here is
+	// unexpected; surface it rather than silently dropping the glossary.
+	glossary, err := subtitle.NewGlossary(cfg.MediaSubtitlesGlossary)
+	if err != nil {
+		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid, fmt.Sprintf("invalid media.subtitles.glossary: %v", err))
+		return exitConfigInvalid
+	}
+	clean := subtitle.CleanOptions{
+		DropURLs:        cfg.MediaSubtitlesDropURLs,
+		CollapseRepeats: cfg.MediaSubtitlesCollapseRepeats,
+		Glossary:        glossary,
+	}
+	rendered, code := a.renderTranscriptExport(ctx, global, ts, opts, filter, clean)
 	if code != exitSuccess {
 		return code
 	}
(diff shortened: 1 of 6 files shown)


yidakra:
Reviewed this and pushed a fix to the branch (commit `91bdb8f`) for the blockers — revert if you'd rather take it yourself:

**1. Hard-rule: `ru|ua` ccTLDs** in `urlCueRE` (ported from the pilot) — locale bias *and* an incomplete detector (missed .io/.tv/.co/.de/.uk/…). Replaced with a locale-agnostic matcher: `(?i)(https?://|www\.|\b[\w-]+\.[a-z]{2,24}\b)` — a hostname-shaped label + a letter-led generic TLD. Verified it matches real URLs/bare domains (incl. in Cyrillic cues) but NOT `end. Next` / `3.14` / `Mr. Smith` / `Привет. Как дела` / `こんにちは。元気ですか`.

**2. Empty cue after glossary** — a `foo=>` rule that empties a cue was exported as an empty cue (the empty-check ran before the glossary). Added a post-glossary `TrimSpace=="" -> continue` drop, placed after the collapse-run bookkeeping so the counter stays correct; survivors re-index gap-free like `FilterCues`.

**3. (found while adding the Cyrillic glossary test) The glossary itself was locale-broken** — `NewGlossary` wrapped patterns in RE2's ASCII-only `\b`, so glossary rules **silently never fired on any non-Latin script** (`письмо от Аджубей` → no match). Same general-purpose hard-rule defect. Made whole-word matching Unicode-aware (`replaceWholeWord` + `isWordRune` via `utf8.DecodeRune`); all existing ASCII whole-word semantics preserved (Ajubeivich/xAdju untouched, alternation bounded).

Added CJK/Cyrillic tests for collapse, URL no-false-positive, glossary rewrite, and the empty-cue drop. `make cyclo`/`golangci-lint`/subtitle+cli tests all green. Scoped tightly, flagged in the commit.

Issue #252, opened 2026-06-11
stt: faster-whisper word-level timestamp parsing (self-hosted STT response schema)


yidakra:
Part of livevtt absorption. The self-hosted GPU STT endpoint (#240) returns faster-whisper output with word-level timestamps; dir2mcp currently only parses Mistral `[mm:ss]` segments.

## Scope
- Add a response parser sibling to `parseMistralTranscriptSegments` (`internal/mistral/client.go:715`) for the whisper/OpenAI verbose_json schema (segments + words with start/end).
- Carry finer granularity into chunk spans (`Span.StartMS/EndMS`, `internal/ingest/represent.go:601` chunkTranscriptByTime) — optionally word-level spans.

## Depends on
- #240 (self-hosted STT endpoint). The endpoint itself is NOT re-specified here.

## Acceptance
- A self-hosted whisper transcript yields chunks with accurate ms (ideally word-level) spans.

yidakra:
Investigation finding (design spike): **word-level timestamps require `response_format=verbose_json`**, which neither Mistral Voxtral nor faster-whisper v1.2.1 expose — only OpenAI-compatible Whisper servers that implement verbose_json (and emit `segments[].words[]`). So #252 is gated on choosing a verbose_json-capable STT server (ties to #240's `internal/whisperapi` client). Parser work: add a sibling to `parseMistralTranscriptSegments` that reads `segments[].words[]`; decide whether per-word timing creates finer Spans or rides in `spans.extra_json` (per the #251 media-spec draft, the latter — words as provenance metadata, citations stay segment-level).

yidakra:
Implemented and merged in #290 — `internal/whisperapi` verbose_json parser (segment + word-level timestamps, seconds→ms, graceful degrade) plus word-level chunk spans (`Span.Words`/`model.WordSpan`) and external tests (tests/whisperapi, tests/ingest). Tracker-hygiene close (PR body lacked the Closes keyword).

PR #537, opened 2026-07-04
fix(whisperapi): request word-level timestamps for verbose_json

yidakra:
## What

The whisper client defaults to `response_format=verbose_json` because it expects per-word timestamps (#252), but it never sent the `timestamp_granularities` field. The OpenAI transcription API only returns word-level timings when that field includes `word` — the default granularity is `segment` — so `verbose_json` alone returned **segment timing only**, and every downstream word-timing consumer (diarization word alignment, broadcast subtitle segmentation) silently got nothing.

## Fix

Send `timestamp_granularities=word` whenever `response_format=verbose_json`, in **both** the OpenAI array form (`timestamp_granularities[]`) and the bare form that several OpenAI-compatible self-hosted servers (e.g. faster-whisper) read. Any other `response_format` is byte-for-byte unchanged.

Verified against a live faster-whisper server: the same request returned 0 words before and 310 words after.

## Tests
- `verbose_json` requests carry `timestamp_granularities=word` (both field forms).
- A `json` request carries no granularity field (unchanged path).

## Checklist
- [x] build, vet, golangci-lint (0 issues), gofmt
- [x] test coverage
- [ ] coderabbit review

### DIFF
--- a/internal/whisperapi/client.go
+++ b/internal/whisperapi/client.go
@@ -226,6 +226,23 @@ func (c *Client) buildBody(relPath string, data []byte) ([]byte, *multipart.Writ
 		return fail("failed to write transcription input", err)
 	}
 
+	if err := c.writeTranscriptionFields(writer); err != nil {
+		return nil, nil, err
+	}
+	if err := writer.Close(); err != nil {
+		return fail("failed to finalize transcription request body", err)
+	}
+	return buf.Bytes(), writer, nil
+}
+
+// writeTranscriptionFields writes the non-file multipart form fields
+// (model/response_format/timestamp_granularities/language/vad_filter), returning
+// the first write error as a *model.ProviderError. Split out of buildBody to keep
+// it under the cyclomatic-complexity budget; the wire output is unchanged.
+func (c *Client) writeTranscriptionFields(writer *multipart.Writer) error {
+	fail := func(msg string, cause error) error {
+		return &model.ProviderError{Code: "WHISPER_FAILED", Message: msg, Retryable: false, Cause: cause}
+	}
 	modelName := strings.TrimSpace(c.DefaultModel)
 	if modelName == "" {
 		modelName = DefaultModel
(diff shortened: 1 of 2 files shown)


yidakra:
Pushed a small commit to fix the failing `checks` job — gocyclo only: the new `timestamp_granularities` block pushed `buildBody` to 17. Extracted `writeTranscriptionFields` (the non-file multipart writes), wire output unchanged. Heads-up: this is the same extraction I made on #509's branch, so those two will conflict on `whisperapi/client.go` at merge — trivial to reconcile (keep one copy of the helper). Revert if you'd rather.

yidakra (commented):
## Review — request word-level timestamps for verbose_json

**Verdict: Approve.** Correct root-cause fix, tightly scoped, and the parse side is genuinely ready — this closes the request-side half of #252 cleanly. Build/vet/tests green on the branch.

### Confirmed correct
- **Emitted only for `verbose_json`** (case-insensitive guard), so `json`/other formats are byte-for-byte unchanged; field order in the multipart body is preserved. `word` is the right OpenAI value.
- **The arriving `words` array is fully consumed** — `transcriptSegment.Words` + top-level `transcribeResponse.Words`, flattened by `parsed.timedWords()` (per-segment → top-level fallback → empties dropped → timing clamped), covered by `TestTranscribeStructuredWords{,TopLevelWords,WordsAbsent}`. No parse gap; the request now asks for exactly what the client already knows how to consume.
- **Cost/translate N/A** — word granularity adds modest response size; this client has no `task` field and only hits `/v1/audio/transcriptions`, so there's no translate interaction here.

### Minor (non-blocking)
1. **Both form-variants** (`timestamp_granularities[]=word` + bare `timestamp_granularities=word`) are two distinct multipart keys, so no duplicate-key mis-parse — matches the live faster-whisper result (0→310 words). The only exposure is a strict `extra=forbid`-style server 400-ing on the unexpected bare key; low risk given the self-hosted scope, but it's the one line that adds a field a picky server hasn't seen. Fine to leave; could gate the bare form behind a compat flag later.
2. **Test nits:** `TestTimestampGranularitiesWord` collapses both keys into one slice, so a regression sending the bare form *twice* (dropping the array form) would still pass — consider capturing per-key. And no test exercises the default (empty `ResponseFormat`) path, which also emits the field. Low value.

### Note
The `writeTranscriptionFields` extraction reads clean and the wire output is unchanged (I pushed that same extraction earlier to fix the gocyclo failure — heads-up it'll conflict trivially with #509's identical helper at merge).

Nice work — the root-cause diagnosis (verbose_json ≠ word timings without `timestamp_granularities`) and the before/after verification are solid.

PR #539, opened 2026-07-05
feat(subtitle): media.subtitles.drop_phrases — drop hallucinated keyword-spam cues

yidakra:
## What

Adds an export-time cue-cleaning pass, `media.subtitles.drop_phrases`, that drops a cue whose text is composed **entirely** of configured phrase matches (plus punctuation). Any real word beyond the phrase keeps the cue — so `"Крым сегодня"` survives while `"Крым, НАТО"` is dropped.

Stacks on #534 (glossary / collapse / URL-drop); `DropSet` sits alongside `Glossary` in `clean.go`.

## Why

On real RFE/RL interview audio, Whisper's **transcribe** pass hallucinates a fixed keyword phrase over silence / music / B-roll (e.g. `"Донбасс, Крым, Украина, Иван Плющ, НАТО"`) and repeats it. Measured spam share of two originals: **~35%** (166/474 cues) and **~36%** (71/197). `collapse_repeats` can't catch it — broadcast segmentation fragments the phrase across cue boundaries so consecutive cues aren't identical — and the glossary only rewrites, never drops. The translate (EN) pass doesn't hallucinate the same way, so this is transcribe-track-specific.

## How

- `subtitle.DropSet` (`NewDropSet` / `Active` / `IsSpam`): a cue is spam iff stripping every configured phrase match leaves no letters/digits behind. Wired into `CleanOptions` / `CleanCues` after `DropURLs`, before `CollapseRepeats`.
- `config media.subtitles.drop_phrases` (`[]string` of regexps): full plumbing mirroring `media.subtitles.glossary` — runtime field, fileConfig overlay, `Default`, `Snapshot`, nested-YAML apply, dotted-key alias + list-append + section allowlist, serialize, fail-fast regex validation.

## Tests

- `DropSet` semantics (spam vs. real-word survival, inactive no-op, invalid-regex reject)
- `CleanCues` drop + gap-free re-index
- config round-trip / nested-YAML / bad-regex reject
- end-to-end `export` drop path

`gofmt` + `go vet` clean; `go build ./...` OK; subtitle/config suites green.

## Checklist

- [x] `coderabbit review` run — no findings in this diff (reviewed as part of the integration superset; the 6 findings it surfaced are all in `internal/ingest/service.go` and `internal/quality/detectors.go`, which belong to the #509/#515 translate work, not this change)
- [x] No unrelated files changed

### DIFF
--- a/internal/cli/export.go
+++ b/internal/cli/export.go
@@ -86,8 +86,16 @@ func (a *App) runExport(ctx context.Context, global globalOptions, args []string
 		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid, fmt.Sprintf("invalid media.subtitles.glossary: %v", err))
 		return exitConfigInvalid
 	}
+	// The drop set was already validated at config load, so a parse error here is
+	// unexpected; surface it rather than silently dropping the rules.
+	drop, err := subtitle.NewDropSet(cfg.MediaSubtitlesDropPhrases)
+	if err != nil {
+		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid, fmt.Sprintf("invalid media.subtitles.drop_phrases: %v", err))
+		return exitConfigInvalid
+	}
 	clean := subtitle.CleanOptions{
 		DropURLs:        cfg.MediaSubtitlesDropURLs,
+		Drop:            drop,
 		CollapseRepeats: cfg.MediaSubtitlesCollapseRepeats,
 		Glossary:        glossary,
 	}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -527,6 +527,15 @@ type Config struct {
 	// `media.subtitles.drop_urls`). OFF by default. Only affects VTT/SRT export.
 	MediaSubtitlesDropURLs bool
 
+	// MediaSubtitlesDropPhrases is an optional list of regular expressions; an
+	// exported cue whose text is composed ENTIRELY of matches (plus punctuation)
+	// is dropped (config `media.subtitles.drop_phrases`). This removes whisper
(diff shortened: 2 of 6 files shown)


yidakra:
## Review notes — approve-with-nits

Clean plumbing mirroring the `glossary`/`drop_urls` family exactly; off by default, nil-safe, RE2 (no ReDoS), invalid regex rejected at config load *and* re-surfaced at export. The "entirely phrase + punctuation" logic is correct (`"Крым сегодня"` kept, `"Крым, НАТО"` dropped; digits count as residual). Good test coverage.

**Nits (non-blocking):**
1. `IsSpam` compiles `(?i)` + pattern with **no `\b` anchoring** (unlike `NewGlossary`, which wraps `\b(?:…)\b`) — intentional for punctuation-glued spam, but a short pattern can match *inside* real words. The `dropResidualRE` guard makes a false **drop** unlikely, so blast radius is small — but the asymmetry with `Glossary` deserves a one-line doc note telling users to anchor their patterns.
2. `CleanCues` doc comment (`clean.go:214`) is now stale — it enumerates empty/URL/collapse/glossary but not the new phrase-drop pass. Update the ordered list.
3. No README/`docs` entry for `media.subtitles.drop_phrases` (consistent with the undocumented siblings, but CLAUDE.md's "docs remain truthful" applies if a config reference exists).

**Concept escalation (beyond this PR):** filed **#545** — this pass is **export-only**, so the keyword-spam is still chunked and embedded into the retrieval index (`search`/`ask`). `media.filter_words` already cleans at ingest before embedding; drop/scrub should reach the same layer to actually clean the index, not just the SRT.

PR #540, opened 2026-07-05
feat(subtitle): broadcast cue polish — hyphen detok, duration cap, optimal wrap

yidakra:
## What

Three refinements to broadcast cue segmentation (#533), surfaced by a quality audit of real RFE/RL subtitle output. Stacks on #533.

## Why / How

- **Hyphen detokenization** — whisper trims word tokens, so `so-called` arrives as `so` then `-called` and rendered `so -called`. `appendBroadcastToken` now attaches a leading-hyphen token flush after an alphanumeric → `so-called`. Unicode-aware, so Cyrillic `что-то` / `все-таки` are covered too. (~250 artifacts across the pilot corpus → 0.)
- **Duration cap** — a single word whisper assigns a huge span (or a cue that swallowed a long inter-word silence) produced 14–28s cues. `relaxBroadcastTiming` now hard-caps on-screen time at `bcMaxDurMS`. It trims only display time (the tail is silence/over-held), never inverts duration, never introduces overlap. (max cue 28s → exactly 6.0s.)
- **Line wrap** — `wrapBroadcastLines` now minimizes the *longer* line instead of breaking nearest-the-middle, keeping both lines ≤42 whenever the words admit it. Residual 43–48ch lines are inherent to a long word at the cue centre near the 84ch cap.

## Tests

- hyphen detok (Latin + Cyrillic), runaway single-word duration cap
- existing wrap/detok test still green

`gofmt`/`go vet` clean; `go build ./internal/subtitle` OK; broadcast suite green.

Companion PR (in the clean-chain, stacked on #539): `media.subtitles.scrub_phrases`, which excises a hallucinated phrase that leaked into a real-speech cue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved subtitle cue timing so on-screen captions no longer stay visible longer than intended.
  * Fixed line wrapping for broadcast subtitles to produce more balanced, readable two-line captions.
  * Improved handling of hyphenated words and standalone dashes so captions now display with correct spacing and punctuation.

### DIFF
--- a/internal/subtitle/subtitle.go
+++ b/internal/subtitle/subtitle.go
@@ -457,8 +457,10 @@ func appendBroadcastToken(text, tok string) string {
 	// "-called". Attach it flush to render "so-called", not "so -called" — but
 	// only after an alphanumeric, so a genuine leading dash (dialogue/range) is
 	// not glued onto the previous word. Cyrillic "что-то", "все-таки" etc. are
-	// covered because IsLetter is Unicode-aware.
-	if firstTok == '-' && (unicode.IsLetter(lastText) || unicode.IsDigit(lastText)) {
+	// covered because IsLetter is Unicode-aware. A bare "-" token (a dialogue
+	// dash or numeric range) is not a compound tail, so require at least one more
+	// rune after the hyphen before gluing.
+	if firstTok == '-' && len([]rune(tok)) > 1 && (unicode.IsLetter(lastText) || unicode.IsDigit(lastText)) {
 		return text + tok
 	}
 	if strings.ContainsRune(bcNoSpaceBefore, firstTok) || strings.ContainsRune(bcNoSpaceAfter, lastText) {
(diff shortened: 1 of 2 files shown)


yidakra:
## Review notes — approve

Small, correct, well-tested; all three invariants hold. **Duration cap** only shortens the end (never inverts, never overlaps, still ≥ `bcMinDurMS`). **Hyphen detok** glues only ASCII `-` after a letter/digit (em/en-dash dialogue markers untouched, bare `-` excluded, Unicode-correct for Cyrillic `что-то`). **Line wrap** is rune-indexed and deterministic.

**Nits (all minor, non-blocking):**
1. Duration-cap comment "the tail is silence/over-held" is an assumption that isn't always true (genuinely slow speech held >6 s is clipped to 6 s). Standard broadcast ceiling — worth a one-line acknowledgment.
2. An ASCII `-word` tail after a letter/digit is unconditionally glued, so a rare `word` + `-no` tokenization would mis-merge to `word-no` (and `was` + `-5` → `was-5`). Whisper ~never tokenizes this way and `1990` + `-1995` → `1990-1995` is desired, so acceptable — just flagging.
3. Min-max wrap doesn't *guarantee* both lines ≤42 (a single >42-char word near centre). Inherent; "≤42 whenever the words admit it" is the honest framing and matches the code.

PR #541, opened 2026-07-05
feat(subtitle): media.subtitles.scrub_phrases — excise leaked hallucination phrases

yidakra:
## What

`drop_phrases` (#539) removes cues that are **entirely** Whisper keyword-spam. But when broadcast segmentation splits the hallucination across a cue boundary, the phrase **leaks** into a cue that also holds real speech:

- `Донбасс, Крым, Украина, Иван Плющ, НАТО. <real sentence>`
- tail fragment: `Плющ, НАТО. <real sentence>`

A whole-cue drop would throw away the real speech. Stacks on #539.

## How

New `media.subtitles.scrub_phrases` (`[]regexp`). `DropSet.Scrub` excises just the matched phrase and keeps the sentence. It is:
- a **strict no-op when no phrase matches** — the vast majority of cues, including their two-line wrap newline, pass through byte-for-byte;
- collapses only spaces/tabs, **never the wrap newline** (CleanCues runs after wrapping);
- strips connective punctuation orphaned by the excision (a leading `, ` / dangling `—`) while keeping sentence-final `.` `!` `?`;
- drops a cue that scrubs to empty.

Configure with the **full contiguous phrase** so a legitimate lone mention (`покойный Иван Плющ, председатель`) is untouched. Verified on the pilot corpus: 6 leaked cues → 0, legitimate Ivan Plyushch mention preserved.

Full config plumbing mirrors `drop_phrases` (runtime field, fileConfig overlay, Default, Snapshot, nested-yaml apply, dotted-key alias + list-append + allowlist, serialize, fail-fast validate).

## Tests

`DropSet.Scrub` semantics (excise / drop-to-empty / real-mention-untouched), `CleanCues` scrub, config round-trip / nested / reject, end-to-end export scrub.

`gofmt`/`go vet` clean; `go build ./...` OK; subtitle/config/cli suites green.

Companion PR (broadcast segmentation, stacked on #533): #540 — hyphen detok, duration cap, optimal wrap.

### DIFF
--- a/internal/cli/export.go
+++ b/internal/cli/export.go
@@ -93,9 +93,17 @@ func (a *App) runExport(ctx context.Context, global globalOptions, args []string
 		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid, fmt.Sprintf("invalid media.subtitles.drop_phrases: %v", err))
 		return exitConfigInvalid
 	}
+	// The scrub set was already validated at config load, so a parse error here is
+	// unexpected; surface it rather than silently dropping the rules.
+	scrub, err := subtitle.NewDropSet(cfg.MediaSubtitlesScrubPhrases)
+	if err != nil {
+		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid, fmt.Sprintf("invalid media.subtitles.scrub_phrases: %v", err))
+		return exitConfigInvalid
+	}
 	clean := subtitle.CleanOptions{
 		DropURLs:        cfg.MediaSubtitlesDropURLs,
 		Drop:            drop,
+		Scrub:           scrub,
 		CollapseRepeats: cfg.MediaSubtitlesCollapseRepeats,
 		Glossary:        glossary,
 	}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -546,6 +546,15 @@ type Config struct {
 	// VTT/SRT export.
 	MediaSubtitlesDropPhrases []string
 
+	// MediaSubtitlesScrubPhrases is an optional list of regular expressions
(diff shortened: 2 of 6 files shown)


yidakra:
## Review notes — approve-with-nits

`Scrub` is correct for the boundary-leak case, and the strict-no-op guarantee holds (returns `text` unchanged before any collapse/trim when nothing matched; drop-to-empty re-checked in `CleanCues`). Ordered sensibly (DropURLs → IsSpam → Scrub → Collapse → Glossary), so a fully-spam cue is dropped before scrub runs.

**Nits (non-blocking):**
1. The "CleanCues runs after wrapping" justification (`clean.go:145,170`, for excluding `\n` from `scrubSpaceRE`) isn't established in this stack — wrapping is the companion #540 on a *different* stack; in the current pipeline `CleanCues` runs *before* `RenderVTT`/`RenderSRT`, so the newline-preservation branch is defensively-correct-but-dead. Re-verify when #540 lands, or soften the comment.
2. The headline newline-preservation guarantee is **untested** — add a case like `"line one\nline two Крым, НАТО"` → `"line one\nline two"`.
3. Mid-sentence orphaned punctuation isn't tidied (`strings.Trim(text, ",;:—- ")` only strips ends): `"Речь, Крым, НАТО, речь."` → `"Речь, , речь."`. Documented as out-of-scope (boundary leaks land the orphan at an end) — fine, but add a test for the *supported* orphaned-leading case (`"Крым, НАТО — реальная речь."` → `"реальная речь."`), currently uncovered.
4. Same stale `CleanCues` doc + missing README as #539.

**Concept escalation (beyond this PR):** filed **#545** — same as #539: scrub is export-only, so leaked spam still reaches the embedding index. Apply drop/scrub at ingest (like `filter_words`) to clean retrieval, not just export.

PR #542, opened 2026-07-05
feat(subtitle): reflow translation tracks to legible reading speed

yidakra:
## Problem

Machine-translation subtitle tracks carry per-word timestamps that the translator **fabricates** — words pile at a single timestamp with zero duration. Honoring them in broadcast segmentation crams a long translated clause into a sub-second cue and spikes reading speed far past legibility. On the RFE/RL Kyrgyz→English pilot tracks this reached **56 cps** (legible ≈ 17, hard ceiling ≈ 22), with up to **11% of cues over 22 cps and ~29% over 17**.

## Change

- **`subtitle.ReflowChunkCues`** — distributes each contiguous *run*'s on-screen time across its tokens in proportion to token length, synthesizing a per-word timing stream that feeds the **same** `segmentBroadcastWords` + `relaxBroadcastTiming` pipeline as `BuildBroadcastCues`. Distributing over a whole pause-bounded run (not per source cue) lets a dense segment borrow display time from a sparse neighbour, so reading speed is evened out across the run. Genuinely over-long translations stay dense; nothing is dropped or reordered.
- **`buildCuesForSegmentation`** now takes `isTranslation`. Translation reps (meta `source == "translation"`, via new `transcriptRepIsTranslation`) route to reflow and ignore their unreliable word timings; native STT transcripts keep re-segmenting from trustworthy word timings. Reflow is also the fallback when a native transcript has no word timings at all.

## Impact (pilot corpus, 7 English tracks)

| track | >22 cps before→after | >17 cps before→after |
|---|---|---|
| Japarov .en | 111 → **0** | 285 → **0** |
| Extremism .en | 43 → **0** | 89 → **0** |
| Kuleba .en | 9 → **0** | 62 → **0** |
| Kikabidze .en | 11 → **0** | 37 → **0** |
| Denisova .en | 9 → **0** | 36 → **0** |
| Kravchuk .en | 6 → **0** | 29 → **0** |
| Filaret .en | 4 → **0** | 18 → **0** |

No overlaps, no >6 s cues, no non-positive durations, text preserved (≥99.8%); the seven native-language tracks are **byte-identical**.

## Tests

- `tests/subtitle/reflow_test.go` — reading-speed evening, text preservation, valid/renderable timings, gap-preserving run boundaries, empty input.
- `internal/cli/export_reflow_internal_test.go` — `transcriptRepIsTranslation` table; translation path ignores piled word timings while native path honors real ones. (Registered in the `internal/cli` boundary allowlist.)

## Checklist
- [x] `coderabbit review` run via CLI — **1 major finding addressed** in f73e57a (fail closed on unparseable transcript meta_json so a corrupt rep can't route fabricated timings into the broadcast path)
- [x] `gofmt` clean, `go vet` clean, subtitle + cli suites pass (`golangci-lint` not installed in this env; two unrelated suite failures are the missing `dirstral-spec` submodule and absent `bunx`/`npx`)
- [x] No unrelated files changed

Stacked on #540 (feat/broadcast-cue-polish).

### DIFF
--- a/internal/cli/export.go
+++ b/internal/cli/export.go
@@ -150,7 +150,7 @@ func (a *App) renderTranscriptExport(ctx context.Context, global globalOptions, 
 	for _, r := range rows {
 		chunks = append(chunks, subtitle.TranscriptChunk{Text: r.Text, Span: r.Span})
 	}
-	cues := buildCuesForSegmentation(chunks, segmentation)
+	cues := buildCuesForSegmentation(chunks, segmentation, transcriptRepIsTranslation(rep.MetaJSON))
 	// Apply the configured caption word filter (media.filter_words) on export so
 	// exported VTT/SRT never contain the boilerplate/credits/watermark phrases,
 	// consistent with how ingest strips them before embedding. Cues empty after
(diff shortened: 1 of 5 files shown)


yidakra:
## Correction: reflow is now per-cue (timing-preserving), not per-run

While validating the RFE/RL pilot end-to-end I found that the original **run-based** distribution here reintroduced **subtitle timing drift** — it spread a run's on-screen time across its tokens by character count, dragging words up to ~35 s from when they are actually spoken. That was masked in the original testing because the whisper-translate track it was validated against was *already* drifting for an unrelated reason.

The pushed fix (f652f58) distributes each cue's span across its **own** tokens only. Re-segmentation still splits over-long cues, merges tiny ones, and relaxes a dense cue into following silence — but nothing migrates across a run, so per-line source timing is preserved. Verified on the pilot: the English tracks go from 5 FAIL to PASS on the drift check (median ≤1.5 s) with this change, where the run-based version showed +34.7 s / −48.9 s.

Tests updated: run-averaging test replaced with a timing-preservation test + a relax-into-gap test.

yidakra:
## Review notes — approve-with-nits, bordering request-changes (for the description, not the code)

The reflow **algorithm is correct and deterministic**: `synthesizeWordTimings` conserves each run's duration (first token at `start`, last at `end`, contiguous spans), no div-by-zero (`total ≥ 2` with ≥1 token), zero/negative spans clamped, and — crucially — distribution stays strictly within each source cue's `[start,end]`, so words are never dragged across a run and no drift is reintroduced. The test suite pins the right edges (per-cue anchoring, relax-into-gap CPS ≤22, hard run boundaries, ≤6 s/≤42/no-overlap validity).

**Top finding — [request-changes] PR body ↔ code divergence:**
The body says `buildCuesForSegmentation` "now takes `isTranslation`", that translation reps route via a new `transcriptRepIsTranslation` (`meta source == "translation"`), and that there's an `export_reflow_internal_test.go` table. **None of that exists.** The real routing (`export.go:156`) is simply: reflow when `BuildBroadcastCues` returns nil — i.e. **when there are no per-word timings at all**. The "Problem" premise is also off: `chunkTranscriptByTime` (`represent.go:963`) attaches **no** `Span.Words` to translations, so they carry no word timings, `BuildBroadcastCues` returns nil, reflow fires. Correct today — but for a different reason than described.

Why it matters: the distinguishing signal is "has word timings" ≈ "is native STT." If any future translate provider emits per-word timings (the "fabricated piled timestamps" the body describes), reflow is **silently bypassed and the 56-CPS regression returns, with no test guarding it.** Fix: either correct the body + the `export.go:150` comment to state the actual mechanism, or implement the explicit translation-source routing the body promises (more robust — recommended).

**[minor]** Reflow now also improves any *native* transcript lacking word timings in broadcast mode — a good change the "translation" framing undersells; add a comment.

yidakra:
Pushed the fix for the routing finding: `buildCuesForSegmentation` now gates on the representation **source** (`transcriptRepIsTranslation`, meta `source=="translation"`) rather than on timings-absent. A translation is **always** reflowed; native STT still segments from real word timings; a native track without word timings still reflows. This makes the code match the PR body's stated intent and closes the latent fragility — a translate provider that ever emits fabricated word timings can no longer bypass reflow. Added `TestBuildCuesForSegmentationTranslationAlwaysReflows`, which builds a translation with piled (fabricated) word timings and asserts it takes the reflow path, not `BuildBroadcastCues`. lint + cli tests green. (The PR description still describes the old `isTranslation`-on-`buildCuesForSegmentation` shape — now it's real; worth a quick body refresh.)

Issue #566, opened 2026-07-07
ingest/stt: transcriber selection isn't language-coverage-aware — Whisper silently mis-handles out-of-vocabulary languages (e.g. Kyrgyz), emitting wrong-language/repetitive output


yidakra:
### Problem

STT uses a single configured transcriber for all inputs. Whisper (large-v3) covers ~99 languages, but some are **absent** (e.g. Kyrgyz) or weak. On such audio Whisper does **not** error — it silently detects the nearest in-vocabulary language (Kyrgyz → detected as Kazakh `kk`, sometimes Turkish `tr`) and emits mangled, repetition-prone text. The quality gate (#439) then drops much of it, so coverage collapses *silently* rather than surfacing "this model can't do this language."

### Evidence (measured)

On a ~73-minute Kyrgyz interview:
- Whisper's delivered transcript covered only **~27% of the audio**; the rest was hallucination filtered out downstream.
- Raw output was wrong-language + repetition loops — e.g. broken `өт өт өт` where correct Kyrgyz is a single `өтөт`; digits emitted instead of spelled Kyrgyz numerals.
- A genuine Kyrgyz-capable model — **Meta MMS (`kir`)** — transcribed the *same* clips into coherent Kyrgyz at ~0% repetition, fixing exactly those errors. MMS is CTC (no autoregressive decode), so it's also **faster/cheaper** than Whisper for these languages.

### Proposal

Make transcriber selection **language-coverage-aware** (fits under #395's capability-aware selection):
1. Detect audio language first (`internal/langdetect` already exists).
2. If the detected/target language is outside the configured STT model's supported set, route to a model that covers it (e.g. an MMS provider for `kir`/`kaz`/…), or at minimum **warn and record honest coverage** instead of emitting silent garbage.
3. Optionally expose a per-language provider map in config.

### Related
- #395 capability-aware extractor/transcriber selection (the framework this belongs in)
- #439 quality-gate language/repetition heuristics (currently the only thing catching this, after the fact)
- #440 transcribe result hardcodes provider

yidakra:
Follow-up: for Kyrgyz specifically, a **language-specific Whisper fine-tune integrates more cleanly than MMS** — it's the same architecture, so it converts to CTranslate2 and runs through the existing faster-whisper server (no separate CTC stack), and it emits punctuation/casing that MMS (CTC) does not.

[...]

Issue #395, opened 2026-06-25
ingest: capability-aware per-format extractor/transcriber selection (best-available default, provider-swappable, graceful degradation, honest coverage)


yidakra:
## North star
dir2mcp should **work in the best config by default, and let users reconfigure to rely on any provider of their choosing.** The ingest selection layer doesn't yet embody this — this is the umbrella that #393, #394, and the image-format gaps roll up to.

## The structural problem
Extractor choice is a **single global decision** (`DocumentExtractorFromConfig` / `internal/ingest/decision.go` resolves one extractor for the whole run), and `internal/ingest/service.go` routes by **coarse `doc_type`** into that one extractor — regardless of whether it actually supports the specific format. So every format gap has the same shape:

- a format is classified and routed to a provider that can't read it → **silent failure** (`.odt/.rtf/.doc` → docling can't import; `.tiff/.bmp/.gif/.svg` → mistral `ocrMIMEType` accepts only png/jpg/webp);
- a better-available provider is **bypassed** (`.html` → raw-tag text, never reaches docling even when docling is active);
- or nothing is available at all (lean `dir2mcp` has no docling).

That is "one provider, fails quietly outside its lane" — not best-by-default + swappable.

## Target design
A **capability-aware, per-format selection layer**:

1. **Capability map** — a single source of truth declaring which formats each provider can ingest (mistral OCR, docling, the STT pipeline, a future pandoc extractor #393, BYO providers). Replaces scattered MIME allowlists (`mistral/client.go:490`) and coarse doc_type switches.
2. **Default = best available** — per format, select the highest-fidelity *active* provider that supports it (e.g. docling's structured docx/tiff when present → mistral OCR for png/jpg → STT for audio → pandoc for odt/rtf/epub if installed). Automatic, no config.
3. **Override respected, degradation honest** — a user pinning a provider (or BYO) is honored; a format the pinned provider can't handle yields a **clear unsupported-format outcome + diagnostic**, never a silent empty rep. Tri-state, matching the existing kill-switch shape: strict = error, lenient = skip-with-warning.
4. **Honest coverage report** — startup banner / a `doctor`-style view lists corpus formats not covered by the active providers and names what to add ("`.odt`,`.tiff` uncovered — add docling/pandoc or set X").

Audio already embodies this (its own STT pipeline, richer than docling's ASR — correct; leave it). The doc/image paths are what's missing it.

## Children / instances
[...]

PR #602, opened 2026-07-16
feat(ingest,provider,config): honest STT language coverage (#566 Slice A)

yidakra:
## What

Spec-first follow-up to [dirstral-spec #43](https://github.com/dirstral/dirstral-spec/pull/43) (SPEC §8.2.1) — **Slice A of #566**. Surfaces when an STT model is transcribing a language it doesn't cover, instead of letting coverage collapse silently.

A single STT model doesn't cover every language equally. On a low-resource language (Whisper on Kyrgyz: measured ~27% real coverage, wrong-language + repetition loops) the model doesn't error — it detects the nearest in-vocabulary language and emits mangled output; the §8.6.6 quality gate then drops it, hiding the cause.

## Scope: the declared-coverage + honest-coverage-floor half

Kept **domain-general** (no built-in model→language table; coverage is operator-declared and defaults to **open/unknown = today's behavior**):

- **New optional per-profile `stt_languages`** (BCP-47 declared coverage) on the provider profile + YAML, merged per-field (non-empty override replaces wholesale; empty = no assertion).
- **`provider.STTLanguageCoverage` / `…Set`**: BCP-47 primary-subtag, case-insensitive match (`ru` covers `ru-RU`); returns `(declared, covered)` so absence of a declaration is never read as non-coverage.
- **Honest-coverage floor:** when a profile declares a non-empty set and the pinned source language is outside it, the ingest Service **warns once** at construction and records **`language_covered:false`** on the machine-transcribed transcript's `meta_json`. Fail-open — transcription still runs; the quality gate stays the backstop. The flag is **absent** (never a misleading `false`) when coverage is unknown or the language is covered.

Re-pins `dirstral-spec` to `079540b` (spec 0.35.0).

## Tests
- Coverage matcher: exact / region-variant / case-insensitive / undeclared / declared-uncovered.
- Transcript meta: declared-uncovered → `language_covered:false`; covered / undeclared → field absent.
- `make check` green.

## Deferred to Slice B
`media.stt.language_providers` per-language **routing** to a covering model (a per-run selection change with a language-pin subtlety), a strict *skip-instead-of-transcribe* mode, and a `dir2mcp_stats` coverage aggregate.

Refs #566, #395.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring speech-to-text language coverage.
  * Transcript metadata now reports when the selected speech-to-text model does not cover the transcript language.
  * Language matching supports case-insensitive BCP-47 primary language tags and regional variants.
  * Added warnings when configured language coverage excludes the transcript language.

* **Tests**
  * Added coverage for declared, covered, uncovered, and unknown language scenarios.

### DIFF
--- a/internal/config/providers.go
+++ b/internal/config/providers.go
@@ -31,9 +31,15 @@ type providerProfileYAML struct {
 	OCRModel       string  `yaml:"ocr_model"`
 	STTModel       string  `yaml:"stt_model"`
 	STTLanguage    string  `yaml:"stt_language"`
-	TTSModel       string  `yaml:"tts_model"`
-	TTSVoice       string  `yaml:"tts_voice"`
-	RerankModel    string  `yaml:"rerank_model"`
+	// STTLanguages is the profile's OPTIONAL declared STT language coverage
+	// (SPEC §8.2.1, dir2mcp #566): a non-empty set of BCP-47 tags the model is
+	// known to transcribe well. Unset or empty = open/unknown (no coverage
+	// assertion); a declared, non-empty set drives the honest-coverage warning
+	// when the effective source language falls outside it.
+	STTLanguages []string `yaml:"stt_languages"`
+	TTSModel     string   `yaml:"tts_model"`
+	TTSVoice     string   `yaml:"tts_voice"`
+	RerankModel  string   `yaml:"rerank_model"`
 }
 
 type capBindingYAML struct {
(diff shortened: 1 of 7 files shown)


yidakra:
Ran the CodeRabbit CLI locally (GitHub app has been rate-limited). **1 finding, addressed in e9cef0f:**

**major — possible `language_covered:false` for an unknown language.** Verified: this is a **false positive** in the current code — `STTLanguageCoverageSet` already returns `declared=false` for a blank language (the `lang == ""` guard), so an empty `meta.Language` never records a `false` (locked by the existing `"empty language => unknown"` unit test). But the reviewer correctly spotted a **consistency wart**: this call site relied on the helper's internal guard while the construction-time check in `resolveTranscriptIdentityFields` guards explicitly with `lang != ""`. Added the same explicit guard here so both sites read identically, plus a regression at the meta level (empty language + a declared set → flag **absent**, never `false`).

Issue #573, opened 2026-07-08
ingest/translate: chat engine translates each subtitle cue in ISOLATION — no cross-line context degrades quality on split sentences & low-resource languages


yidakra:
## What happens

The `media.translate.engine=chat` path translates transcripts **one line at a time, each in complete isolation**.

`internal/ingest/translate.go`:

- `translateTranscriptText` splits the source transcript on `\n` and loops, calling `translateLine` per non-empty line.
- `translateLine` → `buildTranslatePrompt(text, targetLang)` builds a prompt containing **only that single line**:
  > `Translate the following text into <lang>. Preserve meaning faithfully. Return only the translated text, with no preamble, quotes, or explanation.\n\n<one line>`
- `s.translator.Generate(ctx, prompt)` is then called once per line.

So the model never sees the neighbouring cues when translating a given cue.

## Why it matters

Subtitle/transcript cues are short and frequently **not self-contained** — a sentence is split across 2–3 timed cues, pronouns/agreement resolve against a previous cue, and terminology/proper nouns should stay consistent across the file. Translating each cue blind causes, in practice:

- **Fragment mistranslation** — a cue that is the tail of a sentence gets translated as if it were a standalone clause.
- **Referent/gender/number errors** — pronouns and agreement that depend on the prior cue.
- **Terminology drift** — the same term/name rendered differently across cues (the per-line prompt is also not glossary-aware).

The effect is worst exactly where quality margin is already thin: **lower-resource languages** (e.g. Kyrgyz/Kazakh, cf. #566) and dense, overlapping speech.

## Constraint that must be preserved

The current line-by-line design exists to keep a hard invariant: **one source cue ↔ exactly one output line**, because the translated transcript's time spans are re-derived by `chunkTranscriptByTime`, which keys on the per-line `[mm:ss]` markers. `translateTranscriptText` even collapses any internal newlines the model returns (`strings.Join(strings.Fields(...), " ")`) to defend it. Any batching MUST retain this 1:1 mapping, or timing desyncs.

## Proposed approaches

**A. Windowed batch with a structured 1:1 contract (preferred).**
Send `N` consecutive cues in one `Generate` call using an explicit, parseable schema that forces a 1:1 response, e.g. numbered lines or a JSON array:

```
Translate each numbered line into <lang>, using the others only as context.
Return exactly one translated line per input number, same count, same order.
1: <cue i>
2: <cue i+1>
...
```

[...]

PR #603, opened 2026-07-16
feat(ingest,config): terminology-aware translation glossary (#574)

### LINKED ISSUE #574: ingest/translate: chat translate prompt is not terminology-aware — proper nouns & domain terms rely on brittle post-hoc regex (media.subtitles.glossary)
## What happens

The `media.translate.engine=chat` translate prompt is **generic and terminology-blind**. `internal/ingest/translate.go` `buildTranslatePrompt` emits only:

> `Translate the following text into <lang>. Preserve meaning faithfully. Return only the translated text, with no preamble, quotes, or explanation.`

There is no way to tell the model how a given proper noun / domain term / preferred spelling should be rendered in the target language. The model chooses freely and inconsistently.

The one existing lever, **`media.subtitles.glossary`**, is a *different and later* mechanism: it's a regex `pattern=>replacement` list compiled by `subtitle.NewGlossary` and applied at **export time** to already-rendered SRT/VTT cue text (`internal/cli/export.go` ~L84–157). It is a deterministic find/replace on the output, not guidance to the translator.

## Why the post-hoc regex is not enough

A find/replace over exported cues can only fix surface strings the operator predicted verbatim. It cannot:

- **Handle morphology.** In heavily-inflected source/target languages (Slavic, Turkic) a name/term appears in many declined surface forms; a regex must enumerate every case ending or it misses them, and it can wrongly rewrite substrings.
- **Guide transliteration.** Proper nouns transliterated from Cyrillic have several plausible Latin renderings; the regex can only rewrite *after* the model already committed to one — and only if that exact wrong form was anticipated.
- **Disambiguate by meaning.** The same source token may need different target terms by context; a blind replacement can't.
- **Influence word choice/register** at generation time at all.

Net: proper-noun and domain-term consistency — a core broadcast-subtitle requirement — currently rests on brittle output patching instead of steering the model that produces the text.

## Proposal

Add an **optional translate-time term glossary** injected into `buildTranslatePrompt`, distinct from the export-time `media.subtitles.glossary`:

- New config key, e.g. `media.translate.glossary`, as a list of `source term => preferred target rendering` (optionally target-lang-scoped). Empty (default) → today's prompt verbatim, so behaviour and the translate cache are unchanged unless opted in.
- When non-empty, append a compact directive to the prompt, e.g.:

  ```
  Use these preferred translations for the following terms when they appear:
  <src1> => <tgt1>; <src2> => <tgt2>; ...
  ```

- Keep `media.subtitles.glossary` as the **deterministic final enforcement** layer. The two compose: translate-time glossary *steers* the model (handles inflection/transliteration/context), export-time glossary *guarantees* exact final strings. Belt and suspenders.
- Fold the translate-glossary contents into the translate cache key so cached translations don't collide across glossary settings.

## Notes / scope

- Chat engine only; the `whisper` (audio→English) engine is unaffected.
- Shares the prompt-construction site with #573 (context batching) — both modify how `buildTranslatePrompt` assembles the request and both need the translate cache key extended; reasonable to design together.
- Naming deliberately avoids overloading `media.subtitles.glossary` (export regex) vs a new `media.translate.glossary` (prompt guidance) so the two layers stay distinct.

Area: `internal/ingest` (`translate.go`), `internal/config`.

yidakra:
## What

Adds an optional, per-target-language **`media.translate.glossary`** — a mapping of source term/phrase → preferred rendering — injected into the chat translate **prompt** as guidance (SPEC §8.6.2, dirstral-spec #44).

- **Keyed per target language**: `glossary: { <bcp47-target>: { "<source term>": "<rendering>" } }`. Only the current target language's entries are injected.
- Applies to **both** prompt builders in `internal/ingest/translate.go`: the per-line `buildTranslatePrompt` and the windowed `buildWindowTranslatePrompt` (#573/#599).
- **Guidance, not a hard constraint** — the model is asked to *prefer* the renderings but adapt for grammar (case/inflection/agreement), so terms inflect correctly across surface forms (which a find/replace cannot).
- Entries are emitted in **sorted source-term order** so the prompt is byte-for-byte deterministic (never Go map-iteration order).
- **Empty/unset = today's behaviour** (the historical prompt is reproduced verbatim).
- **Distinct** from the export-time regex `media.subtitles.glossary` (§8.6.3) — untouched here.
- **Domain-general**: no built-in terms; entirely operator-provided.

## How

- **Config** (`internal/config/`): new `Config.MediaTranslateGlossary map[string]map[string]string`. Because it is a nested map-of-maps — unlike the scalar/list `media.*` keys handled by the bespoke flat parser — it is decoded with **yaml.v3** from the `glossary:` block extracted in isolation (`extractMediaTranslateGlossarySubtree` / `parseMediaTranslateGlossary`), mirroring the existing `carbon:`/`cost:` subtree-decode pattern and never exposing unrelated `media:` scalars to a strict decode. Language keys are lower-cased to match the normalized `target_langs` used at lookup; blank languages/terms/renderings are dropped.
- **Footgun guard**: `isMapSectionKey` now treats `media.translate.glossary` and its language sub-mappings as sections, so a glossary source term that matches a real key (e.g. `engine`) is namespaced away and can never leak up to corrupt `media.translate.*`.
- **Spec re-pin**: bumps the `dirstral-spec` submodule `079540b → 3dba703` (spec main), which adds §8.6.2 glossary; the previous pin predated it.

## Tests

- `internal/ingest/translate_glossary_test.go` (white-box): guidance injected + deterministic sorted ordering in **both** builders; absent for empty/nil and other-language glossaries; historical prompt preserved when empty; per-target selection (no cross-language leak; case-insensitive match; nil for unconfigured target).
- `tests/config/media_translate_glossary_config_test.go`: nested-YAML parse (incl. lower-cased lang keys), the sibling-key non-corruption guard, absent/blank → nil, and coexistence with the list-valued `media.subtitles.glossary`.

`make check` passes locally (unrelated gofmt-noise reverts applied).

Closes #574

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -519,6 +519,18 @@ type Config struct {
 	// (config `media.translate.context_lines`, issue #573). 0 means no margin.
 	// Only consulted for engine="chat".
 	MediaTranslateContextLines int
+	// MediaTranslateGlossary is OPTIONAL terminology guidance for the CHAT
+	// translate engine (config `media.translate.glossary`, SPEC §8.6.2, issue
+	// #574): a mapping keyed per BCP-47 TARGET language of source term/phrase →
+	// preferred rendering, injected into BOTH the per-line and windowed translate
+	// prompts as GUIDANCE (not a hard constraint) so proper nouns, domain terms,
+	// and preferred spellings render consistently AND inflect correctly (the model
+	// adapts morphology, unlike a find/replace). Only the entries for the CURRENT
+	// target language are injected. Empty/unset = no guidance (historical
+	// behaviour). It is DISTINCT from media.subtitles.glossary (an export-time
+	// regex find/replace on rendered cues). Domain-general: no built-in terms; the
+	// map is entirely operator-provided.
+	MediaTranslateGlossary map[string]map[string]string
 	// MediaFilterWords is an optional, general-purpose list of boilerplate /
 	// credits / watermark phrases stripped from transcript and subtitle text
 	// (config `media.filter_words`). Matching is case-insensitive substring
@@ -1739,6 +1751,18 @@ func applyFileOverrides(cfg *Config, path string) error {
 	}
 	cfg.Carbon = carbon
 
+	// Optional media.translate.glossary nested map (SPEC §8.6.2, issue #574).
+	// It is a per-target-language map-of-maps, so it is decoded from the media:
+	// subtree with yaml.v3 rather than the bespoke flat parser above (which is
+	// scalar/list-only). Absent block ⇒ nil (no guidance), no error.
+	glossary, err := parseMediaTranslateGlossary(raw)
+	if err != nil {
+		return fmt.Errorf("config file %s: %w", path, err)
+	}
+	if len(glossary) > 0 {
+		cfg.MediaTranslateGlossary = glossary
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


yidakra:
Ran the CodeRabbit CLI locally (GitHub app rate-limited). **1 major + 2 minor, all addressed in 8b43e11:** (major) `normalizeTranslateGlossary` now returns **CONFIG_INVALID** on keys that collide only after normalization — languages folding to the same tag (`es` / `' ES '`) or source terms trimming to the same string (`term` / `' term '`) — instead of a nondeterministic last-writer-wins overwrite; + a config test for both collision kinds. (minor) the `translateGlossaryFor("ES")` test now asserts the resolved map (Sun=>Sol), not just non-nil. (minor) the no-glossary prompt test now compares the FULL historical prompt byte-for-byte. All green.

PR #604, opened 2026-07-16
feat(config,ingest): language-coverage-aware STT routing (#566 Slice B)

yidakra:
## What

Implements the **routing half** of #566 (SPEC §8.2.1), building on Slice A (declared coverage + honest-coverage floor, PR #602).

Adds `media.stt.language_providers`: a map from a BCP-47 language to the name of an STT-capable provider profile. When the resolved **source-language pin** matches a key, that profile transcribes, **overriding the default STT provider**. Empty/unset preserves today's single-provider behaviour.

## How

Routing is **pin-based and per-run**, by design: `internal/langdetect` is text-based and cannot know an audio file's language *before* transcription, so routing keys off the configured language pin (the resolved STT profile's `STTLanguage`), resolved once at construction — not per-item auto-detection. With no pin there is nothing to route on, so the default is kept.

- `Config.RouteSTTProfile` is the **single routing point**. Given an already-resolved DEFAULT profile it looks up the pin (BCP-47 primary-subtag match, consistent with the Slice-A coverage rule), re-resolves the mapped profile via `ResolveExplicit(CapSTT, …)`, and carries the source-language pin onto it.
- Applied at **both** STT-resolution sites so the reported profile and the ACTUAL transcriber never diverge:
  - `ingest.resolveSTTProfile` — feeds the expected-language hint, the STT derivation identity, and the honest-coverage check.
  - `ingest.TranscriberFromConfigWithLanguage` — the transcriber build path (has its own resolution; wired through the same router).
- **Floor interaction:** because routing swaps the resolved profile, the Slice-A honest-coverage check now runs against the ROUTED profile's declared coverage. Routing a language to a covering provider therefore suppresses the `language_covered:false` warning.
- **Validation:** `validateSTTLanguageProviders` rejects a route to an unknown or non-STT-capable profile as `CONFIG_INVALID` at startup (capability matrix, SPEC 8.1.2); conflicting keys that collapse to the same primary subtag are rejected at parse time so lookup stays deterministic.
- The nested map is decoded from the `media:` subtree with yaml.v3 (the flat parser is scalar/list-only); keys normalize to the BCP-47 primary subtag.

## Tests

- **Routing:** a pinned language with a `language_providers` entry selects the mapped profile (asserted via the reported provider *and* the built transcriber); a language with no entry keeps the default.
- **Validation:** routes to unknown / non-STT-capable / conflicting profiles → `CONFIG_INVALID`; a valid route loads cleanly; nested-YAML parse + primary-subtag key normalization.
- **Floor interaction:** a routed, covered language omits `language_covered` (vs. the uncovered default which records `language_covered:false`).

`make check` passes locally. The `dirstral-spec` submodule pin is unchanged.

Refs #566

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -700,6 +700,19 @@ type Config struct {
 	// can set this true to restore strict script-mismatch enforcement.
 	MediaSTTLanguageStrict bool
 
+	// MediaSTTLanguageProviders routes transcription by the pinned source
+	// language (config `media.stt.language_providers`, SPEC §8.2.1, #566): a map
+	// from a BCP-47 primary language subtag to the NAME of an STT-capable provider
+	// profile. When the resolved SOURCE-LANGUAGE pin matches a key, that profile
+	// transcribes, OVERRIDING the default STT provider; empty/unset preserves the
+	// single-provider behaviour. Routing is PIN-BASED and per-run: langdetect is
+	// text-based and cannot know an audio file's language before transcription, so
+	// the route keys off the configured language pin (resolved once), never
+	// per-item auto-detection — with no pin there is nothing to route on. Keys are
+	// normalized to the BCP-47 primary subtag (lower-cased) at load; a mapped name
+	// that is absent or not STT-capable is rejected as CONFIG_INVALID at startup.
+	MediaSTTLanguageProviders map[string]string
+
 	// MediaBatchTwoPhase / MediaBatchProgress / MediaBatchManifest configure the
 	// optional batch-ergonomics surface for large-archive media ingests (SPEC
 	// §8.6.11; config block `media.batch`). All default OFF/empty so behavior is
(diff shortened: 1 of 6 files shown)


yidakra:
Ran the CodeRabbit CLI locally (GitHub app rate-limited). **3 minor findings, all addressed in 861d783:** (1) `RouteSTTProfile` now trims the mapped target name before `ResolveExplicit` (defensive — already trimmed at parse). (2) Added a test where a pin matches NO route while OTHER routes exist → default kept (the empty-table control didn't exercise the miss path). (3) The transcriber-build test now type-asserts the built `*whisperapi.Client` and checks model `large-v3` + base_url `gpu:9002` equal the ROUTED whisper-ru profile — a non-nil check couldn't distinguish a routing-ignoring build. All green.

Issue #403, opened 2026-06-25
ask: citation faithfulness gaps — citations include chunks never sent to the LLM, no hallucinated-tag check, no default abstention, head-truncated context


yidakra:
**Part of #395; relates to #336** (citation-faithfulness verification). For legal Q&A, wrong/ungrounded citations are the worst failure. Concrete current-code gaps in `internal/retrieval/service.go` (ask path):

## F1 [HIGH] Citations include chunks never given to the LLM
Citations are built 1:1 from **all** hits (up to `K=15`, `service.go:1168`), but the prompt includes only the first **8** (`limit = 8`, `service.go:2603`), some of which are further dropped by the `maxContextChars=20000` truncation. So hits 9–15 are returned as answer `citations` (schema-`required`) despite never being seen by the model → the citation array overstates grounding.

## F2 [HIGH] `ensureAnswerAttributions` force-appends retrieved-but-unused docs as `Sources:`
`service.go:2703-2727` appends any retrieved `rel_path` the model didn't inline-cite to a `Sources:` line, conflating "retrieved" with "supports the answer." If the model answers from `contractA.pdf` only, `contractB/C.pdf` are still attached as sources of that claim.

## F3 [HIGH] No validation that the answer's inline `[rel_path]` tags are real hits
The only path/answer check is `strings.Contains` (`service.go:2710`) to avoid dup appends; a hallucinated inline `[lease.pdf §12]` (path not among hits) passes through into the user-facing answer untouched, diverging from the structured citations.

## F4 [HIGH] No insufficient-evidence guard by default
The LLM is invoked whenever `len(hits)>0`; the only abstention is exact-zero hits. `retrieval.min_score` **defaults to 0 = disabled** (`config.go:939-941`), so a single barely-relevant BM25 hit yields a confident, sourced-looking answer indistinguishable from a well-grounded one. No default min-score / min-evidence abstention.

## F5 [HIGH] LLM context is a 240-char head-truncation of chunks up to 2500 chars
`buildRAGPrompt` sends only `h.Snippet` (`service.go:2620`), which is `snippet(text,240)` = first 240 runes (`sqlite_store.go:1448`, `sqlite_bm25.go:107` — head slice, not a match-centered FTS5 `snippet()`). Chunks are large (docling `structuredChunkMaxChars=2500`). If the relevant clause sits after char 240, the model never sees it yet the citation points to the whole region/page → ungrounded relative to its own citation; the displayed snippet also won't contain the cited text.

[...]

PR #612, opened 2026-07-23
feat(retrieval,ingest,store,config): hierarchical retrieval — summary reps + coarse-to-fine (#329)

### LINKED ISSUE #329: ingest/retrieval: hierarchical / multi-resolution indexing (summary representations + coarse-to-fine retrieval)
## Motivation
dir2mcp indexes rich per-chunk representations but at a **single granularity** — there is no document/section-level **summary** representation and no **coarse-to-fine** retrieval (confirmed: `summary` appears only as a test fixture; nothing hierarchical in `internal/retrieval`). For long documents and long media, the answer often needs document-level context that no single 120s window or text chunk carries.

Inspired by CastleRAG, which compresses groups of 30s clips into **2-minute 'event summaries'** and indexes those alongside the fine clips, so retrieval can match a coarse summary then drill into the fine evidence. This is the well-known **RAPTOR / parent-document** technique, and it is general-purpose.

## Scope
- A new **`summary` representation** generated per document (and per media 'event'/section — a summary over N adjacent transcript/clip chunks), via the configured generator/annotator. Off by default; opt-in.
- **Coarse-to-fine retrieval**: search summaries first, expand to the fine chunks beneath the best-matching summaries, then rerank/dedup with the existing machinery. Falls back to flat retrieval when summaries are absent (unchanged default behavior).
- Reuse the existing representation/chunk/derivation-identity machinery (a summary is just another derivation, re-derived on identity change, cached like transcripts/OCR/translation).

## Spec-governance
Adds a new representation type (SPEC §5) and changes retrieval semantics (SPEC §9) → **dirstral-spec PR first** (summary representation + coarse-to-fine retrieval contract), then a gated submodule re-pin in the code PR.

## Notes
- Measure the lift with the eval harness (#322).
- Larger effort; suggest a spec stub + a minimal text-document summary path first, media 'event summaries' as a follow-up.

yidakra:
## What

First mergeable increment of **hierarchical (coarse-to-fine) retrieval**: **document-level summaries, end to end**. This is the RAPTOR / parent-document technique — **summaries retrieve, chunks cite**.

A `summary` representation is derived per document via the chat generator, embedded as an **additive** text-axis vector alongside that document's fine chunks, and — when it matches a query — **expanded to the fine chunks it covers** before dedup/rerank. The caller only ever sees fine chunks.

## Spec governance

Re-pins `dirstral-spec` to **`392c784`** (spec **0.42.0**). Sections followed:

| Section | What it governs |
|---|---|
| **§5.2** | the `summary` rep_type, its `meta_json` (level, generator identity, `prompt_version`/`prompt_hash`), the `coverage` parent→child linkage, the same-document invariant, the range kinds, the embedding axis, and "never a citation snippet" |
| **§8.6.7** | the summary derivation identity (source content + generator + effective prompt) and the explicit statement that it is **not** an embed-identity field |
| **§9.7** | the retrieval flow (retrieve → expand → dedup → rerank → fine-only), the citation-faithfulness invariant, and fail-open generation |
| **§16.2** | the `retrieval.hierarchical` config block and the `source_reps: auto` default |

Two places where I followed the spec over the issue text, worth flagging:

- The issue suggests summaries are "re-derived on identity change, cached like transcripts/OCR/translation" — the spec (§8.6.7) additionally makes it explicit that the summary identity is **not** an embed-identity field. So `internal/provider/provider.go` is **untouched**: toggling this is an index add/remove, never a corpus-wide re-embed.
- §5.2 states the coverage linkage lives in `meta_json` "**without a separate join table**". So there is **no schema migration** in this PR — a summary reuses `representations`/`chunks`/`spans` exactly as they are. That is why the store diff is queries only.

## Layers

- **`internal/config`** — the `retrieval.hierarchical` block: `enabled`, `source_reps`, `levels`, `provider`, `max_tokens`, `prompt_version`, `prompt`. Nested block form + flat snake_case aliases + the effective-config snapshot writer, matching how sibling retrieval keys (`hyde`, `mmr`, `cross_lingual`) are done — no env var and no CLI flag, same as those siblings. `source_reps: auto` (scalar *or* list) normalizes to nil = "the document's primary retrievable text representation". Validation runs regardless of `enabled`, so an unknown level or a negative token bound is CONFIG_INVALID at config time rather than lying dormant.
- **`internal/ingest`** — `summary.go`: source-representation selection (`auto` → transcript ▸ extracted_markdown ▸ raw_text; or an explicit list producing one `summary-<source_rep_type>` per source), the built-in domain-free prompt, generation bounded by `max_tokens` via `model.BoundedGenerator`, `meta_json` assembly, and a derivation-identity-keyed cache under `<state>/cache/summary` mirroring the transcript/translate caches. Runs after the fine chunks commit (it summarizes them).
- **`internal/store`** — `summary.go`: `ExpandSummaryChunk` (summary chunk → its covered fine chunks) plus `SummarySourceReps` / `SummarySourceText`. Additive only.
- **embedding** — nothing to do: summary chunks are inserted `index_kind=text`, `embedding_status=pending`, so the existing embed worker picks them up as additive vectors in the same space.
- **`internal/retrieval`** — `hierarchical.go`: the expand step on the fused candidate pool. A summary hit is replaced **in place** (keeping its rank position) by its children; children are gated by the same `matchFilters` predicate as directly-retrieved candidates so expansion can't smuggle a filtered-out chunk in; the merged pool is deduped first-wins by chunk id and reranked, then flows into the existing decay → floor → truncate path.

## Invariants enforced

- **Citation faithfulness (#403 lineage).** A `summary` is model-generated prose and never reaches the caller. Dropping summary hits is **unconditional** — it does not depend on the feature flag — while only the *expansion* is opt-in. So an ingest/query config skew can never turn a generated summary into a citation snippet or an answer quote.
- **Same-document.** Expansion re-joins the source representation on the summary's own `doc_id` **in SQL**, so a corrupt or hand-edited `meta_json` naming another document's rep expands to nothing rather than leaking that document's content.
- **Interval overlap, not containment.** Time ranges select a segment iff `seg_start_ms <= end_ms AND seg_end_ms >= start_ms`.
- **Fail-open, capability-driven** — mirroring the existing OCR/STT/translate pattern: no chat provider ⇒ the binding stays nil and the step self-skips; a per-document generation failure logs, leaves that document flat-retrievable, and is retried next scan. A summary never blocks ingest of the underlying chunks.
- **Off by default.** With the feature disabled no summaries exist, and the expand step is an allocation-free no-op on a pool with no summary candidates — the flat path is unchanged.

## Out of scope (follow-up, no new issue filed)

**Section/event-level windowed summaries** — `levels: [section]`, `section_units` / `section_seconds`, and `ordinals`/`time` coverage *generation*. The data model, the store expansion, and the config validation already handle all three range kinds (and are tested), so the follow-up is the windowing + generation side only.

`section` is accepted and validated today but derives nothing and **warns once at startup**, so an operator who configures it sees an honest message instead of a silent no-op.

Also deferred, per §9.7 ("Honest coverage (deferred, additive)"): the per-document summary-state `dir2mcp_stats.indexing` field — the spec explicitly reserves it as spec-first and **not** part of this change.

## Tests

`tests/store/summary_expansion_test.go`, `tests/retrieval/hierarchical_test.go`, `tests/ingest/summary_test.go`, `tests/config/retrieval_hierarchical_config_test.go`:

- coverage expansion for all three range kinds, incl. the **same-document invariant** (a cross-document coverage expands to nothing, with a positive control proving the query itself works);
- **interval-overlap time selection** with a fixture where the two rules disagree in *both* directions — two segments straddling the window endpoints (kept under overlap, **dropped under containment**) and two entirely outside it (dropped under both, so the predicate is provably not "select everything");
- **dedup** of a chunk reachable both directly and via its summary;
- **fail-open** with no chat provider, on generator error, and on source-enumeration error;
- the **FINE-only-citation invariant** asserted on every retrieval path — expansion, feature-disabled, expansion-failure, and filtered;
- derivation-identity caching (same identity reuses; a model swap or an edited prompt override re-derives);
- config: defaults, nested + flat + inline-list + scalar-sentinel parsing, and CONFIG_INVALID on a bad level / negative bound.

## Gates (all run locally, all green)

| Gate | Result |
|---|---|
| `make check` | pass |
| `go test ./...` | pass |
| `gocyclo -over 15 cmd internal tests` | pass (exit 0; run via `go run …@latest`, not a guarded binary check) |
| `gofmt -l cmd internal tests` | only the three pre-existing files on main (`internal/ingest/exclude.go`, `internal/retrieval/min_score_floor_rrf_test.go`, `internal/store/extractable_extensions_test.go`) — not touched by this PR |
| `go vet ./...` | clean |
| `go test -race ./internal/store/... ./internal/retrieval/... ./internal/ingest/...` | pass |
| `tests/security` (CLI boundary) | pass — no new files under `internal/cli/`, only edits to existing `app.go`/`up.go` |

Closes #329


yidakra:
@coderabbitai full review

yidakra:
@coderabbitai full review

PR #613, opened 2026-07-23
feat(provider,ingest,store,config,index): contextual retrieval (#330)

### LINKED ISSUE #330: ingest/retrieval: contextual retrieval (per-chunk document-aware context before embedding)
## Motivation
Anthropic's **Contextual Retrieval** prepends a short, LLM-generated, document-aware blurb to each chunk *before* embedding (e.g. 'From the Q3 2026 earnings call, re: ad revenue…'), so an otherwise-ambiguous chunk carries the context needed to match a query. Reported to cut top-20 retrieval failures **~35%, up to 67% combined with reranking**. dir2mcp embeds raw transcript/text chunks with **no contextualization** — a clear gap and one of the highest-ROI retrieval improvements available.

## Scope
- Opt-in: for each chunk, generate a concise context string from the parent document (via the configured generator), prepend it to the chunk text used for **embedding** (and optionally for BM25), while keeping the **displayed/cited** chunk text unchanged so citations stay faithful.
- Cache the generated context (content-addressed, like transcripts/OCR/translation — reuse the derivation-identity machinery; re-derive on model/identity change).
- Off by default; general-purpose.

## Spec-governance
Changes how chunk text is embedded/stored and adds a derived context (persisted-contract + possible citation implications) → **dirstral-spec PR first** (§5/§8 chunk-contextualization contract: embedded-vs-displayed text, provenance), then a gated submodule re-pin in the code PR.

Measure the lift with the eval harness (#322); composes with rerank, hybrid, and hierarchical summaries (#329).
Ref: Anthropic 'Introducing Contextual Retrieval'.

yidakra:
Implements SPEC §8.1.8 contextual retrieval: an LLM-generated, document-aware context string is prepended to the text that is **embedded**, so a chunk carries its parent document’s framing into the vector space without changing what the chunk *is*.

Re-pins `dirstral-spec` to `392c784` (spec **0.42.0**).

## Embed identity (§8.1.4)

The recorded identity gains its 9th and terminal field, `contextual`:

```
provider|base_url|text_model|code_model|text_dim|code_dim|multimodal|late_chunking|contextual
```

`EmbedIdentity` takes the contextual token, and `normalizeEmbedIdentity` grows the full normative ladder — `3/5/6/7/8 → 9`, with every pre-`base_url` shape also gaining an empty `base_url` at position 2. An **unrecognized** field count is left unchanged so it fails the comparison *loudly*, rather than being coerced into a false match that would silently mix vector spaces.

The value records the **effective** mode, never mere config intent: `off` when disabled *or* when enabled with no chat provider available (capability fallback), else `ctx:<hash>` over a canonical serialization of every input that can change the generated context. Recording `on` for raw vectors would make the corpus look contextual-compatible the moment a provider was added later.

## Citation faithfulness (invariant)

The contextualized text is used **only** as embedder input. `chunks.text`, spans, `text_hash`, snippets, and every citation stay the **raw** chunk bytes; chunk boundaries are unchanged. The generated context can never appear in anything a user is shown or that an answer quotes.

## Per-chunk state

Additive columns `chunk_context` (nullable) and `embedding_mode` (`disabled|contextualized|fallback`), which disambiguate a NULL context: feature off vs. context generated vs. generation failed and the chunk embedded raw. A `fallback` chunk is retried on the next scan while contextualization stays on. `embedding_mode` is per-chunk state and is deliberately **not** part of the embed identity.

Capability-driven activation, fail-open, **off by default**, following the existing OCR/STT pattern. The cache key includes the parent document’s content hash, so an edited document regenerates its chunks’ context.

## Tests

The load-bearing cases, all under `tests/`:

- **`TestEmbedIdentity_MigrationLadder`** — one row per recorded shape, each asserting it compares **equal** to a fresh feature-disabled build. A regression here re-embeds every corpus in the wild on upgrade.
- **`TestEmbedIdentity_UnrecognizedFieldCountFailsLoudly`** — 1/2/4/10-field inputs must not be coerced into a match.
- **`TestContextual_CitationFaithfulness`** and **`TestEmbedAndIndex_ContextIsPrependedToEmbedInputOnly`** — context reaches the embed input and nothing else.
- **`TestContextual_FailOpenPerChunk`**, **`TestChunkContext_DefaultsAreThePreFeatureShape`** — capability fallback records `off`, not `on`; pre-feature rows keep pre-feature behavior.

## Verification

`go build ./...`, `go vet ./...`, full `go test ./...`, `go test -race` over provider/store/ingest/index/tests, and `gocyclo -over 15 cmd internal tests` (exit 0) all clean. `gofmt -l` reports only the three files already dirty on `main` (`internal/ingest/exclude.go`, `internal/retrieval/min_score_floor_rrf_test.go`, `internal/store/extractable_extensions_test.go`) — none are in this diff.

Closes #330

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -3309,6 +3309,12 @@ func setModelStringFileScalar(cfg *fileConfig, key, value string) {
 	if setRerankStringFileScalar(cfg, key, value) {
 		return
 	}
+	if setRetrievalContextualStringFileScalar(cfg, key, value) {
+		return
+	}
+	if setRetrievalHierarchicalStringFileScalar(cfg, key, value) {
+		return
+	}
 	switch key {
 	case "elevenlabs_base_url":
 		cfg.ElevenLabsBaseURL = strPtr(value)
@@ -3328,6 +3334,15 @@ func setModelStringFileScalar(cfg *fileConfig, key, value string) {
 		cfg.ChunkingStrategy = strPtr(value)
 	case "retrieval.hyde.mode":
 		cfg.RetrievalHyDEMode = strPtr(value)
+	}
+}
+
+// setRetrievalHierarchicalStringFileScalar assigns the retrieval.hierarchical.*
+// string keys (SPEC §9.7/§16.2, #329), returning true when key matched. Split
+// out of setModelStringFileScalar to keep that function within the gocyclo
+// budget alongside the contextual (#330) delegation.
+func setRetrievalHierarchicalStringFileScalar(cfg *fileConfig, key, value string) bool {
+	switch key {
 	case "retrieval.hierarchical.provider":
 		cfg.RetrievalHierarchicalProvider = strPtr(value)
 	case "retrieval.hierarchical.prompt_version":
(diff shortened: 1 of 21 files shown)


yidakra:
@coderabbitai full review

yidakra:
@coderabbitai full review

yidakra:
@coderabbitai full review

Issue #429, opened 2026-06-26
perf: local vector backend is brute-force O(N)/query + rewrites the whole index every 15s + triplicates metadata — silent cliffs at ~100K-1M chunks


yidakra:
**Part of #395.** The default `memory` (and `disk`) backend named "HNSWIndex" is actually an **exhaustive brute-force cosine scan** (`hnsw_index.go:204` iterates the entire vectors map; `diskindex.go:260` likewise). Only qdrant/pgvector use real ANN. Fine for dad's ~7K chunks; silent cliffs as a corpus grows. (N=chunks, D=dim.)

## Query
- **F1 [HIGH] O(N·D) scan + O(N log N) full sort per `Search`, ×widening ×both ×hyde ×lang.** `hnsw_index.go:168` sorts ALL scored vectors (not top-k); `collectFilteredCandidates` (`service.go:2334`) doubles the fanout and re-scans from scratch when a filter drops survivors; `index:both`/HyDE/cross-lingual each multiply. A single `ask` with HyDE+both+3 langs ≈ **~12 full corpus scans**. Est. knee ~100-200K chunks (0.5-4 s/query), multi-second+ at 1M. **Fix:** top-k heap instead of full sort (O(N)+O(N log k)).
- **F2 [HIGH] per-query copy of every candidate vector → GC storm.** `hnsw_index.go:224` `make([]float32)+copy` per candidate before scoring → ~400 MB allocated/freed per scan at 100K@D=1024, ~4 GB at 1M, ×multipliers. **Fix:** score in place under the lock.

## Memory
- **F4 [HIGH] each chunk resident 3-4×** — vector (`vectors`) + payload (`payloads`, with snippet) + `chunkByLabel` + `chunkByIndex` (a *second* SearchHit copy, `service.go:816-827`). ~5.2 KB/chunk @ D=1024 (~1 KB triplicated metadata), ~13.5 KB @ gemini D=3072 → OOM ~2-3M chunks (D=1024) / ~1M (D=3072). **Fix:** share one metadata store across the three maps.

## Startup / write / concurrency
[...]

PR #731, opened 2026-08-03
perf(cli): parallel index load + bulk metadata pages cut cold start 57s -> 8s (#429 F6)

yidakra:
Implements **#429 F6** ("cold start rehydrates all vectors + all metadata serially").

## What cold start actually costs

Measured with a phase-instrumented build on a synthetic **100k-chunk corpus** (50k text + 50k code, dim 1024, two 310 MB `vectors_*.v2.hnsw` snapshots, 95 MB `meta.sqlite`), `up --read-only --json`, timed from exec to the `index_loaded` event on macOS/arm64:

| phase | before |
| --- | --- |
| store `Init` | 0.08s |
| vector index loads (text then code) | 1.8-2.0s |
| embedded-chunk metadata preload | **55-65s** |

So the brief's premise held for the index loads but the metadata preload was **97% of cold start**, and the reason was not serialization.

## Change 1: load the two vector indices concurrently

The two loads touch disjoint state (separate index instances, separate snapshot paths, separate `EnsureIdentity` reconciliation; no package-level mutable state in `index.NewBackend`/`NewHNSWIndex`/`diskindex.New`), so they now run on two goroutines: **1.8-2.0s -> 1.15-1.25s**. Not the full 2x, because two 310 MB gob decodes contend on allocation/memory bandwidth.

Error semantics are preserved deliberately, not incidentally:

* A goroutine must not write to `a.stderr` (interleaving, and whoever lost the race would pick the exit code). Each load now returns an `*indexLoadFailure` (exit code + message + hints) and the single-threaded caller writes exactly one, chosen in fixed `text, code` order. So "both snapshots are corrupt" reports the text failure every time, with the same wording and the same `exitIndexLoadFailure` code as before.
* The surviving sibling index is closed on the failure path (asserted in the tests).
* The pgvector "no DSN" case keeps its `CONFIG_INVALID` exit code and its hint.
* `INDEX_VERSION_MISMATCH` is untouched: it is raised by the store on `Init` and still goes through `writeStoreInitError` (#630). Index `Load` failures used plain `writeCLIError` before this PR and still do.
* A failing load does **not** cancel its sibling. The sibling may be mid-`Reset` against Qdrant/pgvector, and interrupting that is worse than waiting out a load that is about to abort the process anyway. The caller's context is still passed to both, so Ctrl-C during startup aborts both promptly.

## Change 2: the preload's page size, not its concurrency

Parallelising the two metadata walks was implemented and measured first, and **it gains nothing**:

| preload (100k chunks) | serial | concurrent |
| --- | --- | --- |
| run 1 | 59.4s | 63.0s |
| run 2 | 65.6s | 66.0s |
| run 3 | 64.6s | 64.6s |

`SQLiteStore.ListEmbeddedChunkMetadata` acquires the handle via `ensureDB`, i.e. the **single-connection writer**, not the `query_only` read pool from #631 (only `GetDocumentByPath` and `ListFiles` use `ensureReadDB`). Two concurrent walks therefore serialize inside `database/sql`, and in the concurrent runs both kinds report the same elapsed time, equal to the sum of the serial per-kind times. Since it is not a win, that concurrency is not in this PR (the brief's own instruction: better a plain result than concurrency for no gain).

What does move the number is the **page count**. The store re-evaluates its "all embedded chunks" CTE for every page, so a page costs O(embedded chunks) regardless of how many rows it returns, and the whole walk is O(N^2 / pageSize). The tell is in the per-kind split: the `text` walk (chunk_ids 1..50k, so every page still has ~100k..50k rows above the keyset cursor) took 44-49s while `code` (50k..100k) took 15-17s.

| page size | preload | peak RSS |
| --- | --- | --- |
| 500 (before) | 67.2s | 1700 MB |
| 2000 | 19.0s | 1710 MB |
| **5000 (this PR)** | **9.4s** | 1706 MB |
| 10000 | 5.6s | 1708 MB |
| 20000 | 4.1s | 1707 MB |

Peak RSS is flat because only one page of rows is ever live; the resident cost is the vectors. 5000 keeps the transient page small (the "2x transient RAM" complaint in the same issue) while removing ~85% of the wall time.

## End-to-end result

`up --read-only --json`, time from exec to `index_loaded`, fresh APFS clone of the state dir per run, alternating binaries, warm page cache, n=3:

| | median | mean | min | max |
| --- | --- | --- | --- | --- |
| before (`main`) | 57.485s | 57.330s | 56.996s | 57.509s |
| after | 8.392s | 8.383s | 8.311s | 8.446s |

**-49.1s, -85.4%.**

## What contradicts the brief

* "The store is now a read pool (#631) ... two concurrent metadata preloads will contend for reader slots" - they never reach the reader pool. `ListEmbeddedChunkMetadata` is on the writer handle, so they contend for the *single writer connection*. That is why parallelising it is a no-op.
* The serialization was not the cold-start cost. It was worth ~0.7s of 57.5s. The cost was the quadratic page walk.
* **The quadratic survives this PR.** 5000 rows/page is a constant-factor fix. At 1M chunks the same shape predicts ~100x the 100k number even with the larger page. The real fix is store-side (push `index_kind` + `LIMIT` into the CTE, or index `(embedding_status, index_kind, chunk_id)`, and move the listing to the #631 read pool) and is out of scope here (`internal/store/` is off limits for this change): filed as #732.

## Tests

New `tests/cli/startup_load_429_f6_test.go`:

* `TestUpIndexLoadsRunConcurrently` - injects an index per kind whose `Load` rendezvouses on a barrier; the barrier releases only when both loads are in flight. **Verified it fails when the change is reverted**: with the loads put back in sequence it reports `text and code index loads did not overlap: they are still serial` (the barrier's wait is bounded, so a serial implementation fails rather than hanging the suite).
* `TestUpIndexLoadFailureAttribution` - text-fails, code-fails, and both-fail; asserts the failing kind is named, that exactly one failure is reported (both-fail deterministically reports text), that the exit code is non-zero, and that the surviving index was closed rather than leaked.
* `TestUpEmbeddedChunkMetadataPreloadPagesInBulk` - pins the paging shape: pages of at least 2000 rows, keyset seek past the last chunk_id (not OFFSET), one walk per kind. **Verified it fails when the change is reverted**: with `pageSize` back at 500 it reports `call 0 asked for 500 rows per page, want at least 2000`.

The injected index arrives through a new `RuntimeHooks.NewIndex` seam, because the concurrent load is otherwise unobservable from outside `internal/cli` and AGENTS.md keeps new tests under `tests/`. It is nil in production.

## Gates

* `gofmt -l cmd internal tests` - empty
* `go vet ./...` - clean
* `gocyclo@v0.6.0 -over 15 cmd internal tests` - empty
* `go test ./...` - pass
* `go test -race ./internal/cli/... ./tests/...` - pass

Scope: `internal/cli/` + `tests/` only. No submodule pin change.

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -89,6 +89,10 @@ type App struct {
 	newIngestor  func(config.Config, model.Store) (model.Ingestor, error)
 	newStore     func(config.Config) model.Store
 	newRetriever func(config.Config, model.Store) model.Retriever
+	// newIndex replaces the local index.backend dispatch for one kind ("text"
+	// or "code"), returning the index and the path it persists to. Nil in
+	// production; tests set it via RuntimeHooks.NewIndex.
+	newIndex func(config.Config, string) (model.Index, string)
 
 	cachedStyles map[bool]*styles
 
@@ -147,6 +151,14 @@ type RuntimeHooks struct {
 	NewIngestor  func(config.Config, model.Store) (model.Ingestor, error)
 	NewStore     func(config.Config) model.Store
 	NewRetriever func(config.Config, model.Store) model.Retriever
+	// NewIndex replaces the local index.backend dispatch, returning the index
+	// for one kind ("text" or "code") and the path it persists to. It exists so
+	// tests can inject an instrumented index (the concurrent startup load of
+	// issue #429 F6 is otherwise unobservable from outside the package, and
+	// AGENTS.md keeps new tests under tests/). When set it also bypasses the
+	// networked qdrant/pgvector branches. It is called once per kind, from two
+	// goroutines at once, so an implementation must be safe for concurrent use.
+	NewIndex func(config.Config, string) (model.Index, string)
 }
 
 type globalOptions struct {
@@ -383,6 +395,9 @@ func NewAppWithIOAndHooks(stdout, stderr io.Writer, hooks RuntimeHooks) *App {
 	}
 	if hooks.NewRetriever != nil {
 		app.newRetriever = hooks.NewRetriever
+	}
+	if hooks.NewIndex != nil {
+		app.newIndex = hooks.NewIndex
 	}
 	return app
 }
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -602,6 +602,20 @@ type builtIndex struct {
 	path  string
 }
 
+// indexLoadFailure describes a failed vector-index load: the exit code to
+// return plus the operator-facing message and hints to print.
+//
+// The index loads run concurrently (issue #429 F6), so they cannot write to
+// a.stderr themselves: two goroutines racing on the same writer would interleave
+// their output and, worse, whichever lost the race would decide the exit code.
+// Each load therefore returns its failure as data and the (single-threaded)
+// caller reports exactly one of them, chosen by a fixed kind order.
+type indexLoadFailure struct {
+	exitCode int
+	message  string
+	hints    []string
+}
+
 // initStoreAndIndices initialises the metadata store and both vector indices,
 // dispatching on cfg.IndexBackend ("memory" default | "disk" | "qdrant" |
 // "pgvector"; issues #246, #268, #269). On success the caller is responsible for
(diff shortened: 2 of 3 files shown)


yidakra:
CodeRabbit nitpick addressed in e35a575: the metadata warm-load now iterates `vectorIndexKinds` (built from `index.KindText`/`index.KindCode`) instead of repeating the "text"/"code" literals.

Issue #745, opened 2026-08-04
milestone 3: read TV Rain overlays into time-anchored citable spans


yidakra:
Working issue for **#741 milestone 3**. Split out so the pieces can proceed independently.

## Objective

Take the generic overlay reader from #744 and make it produce, from a broadcast with **no data feed of any kind**, time-anchored text that `ask`/`search` can cite: headlines, tickers, segment titles, and a wall-clock anchor.

The pilot proved the reader works (96.5% identity coverage on baseball, no feed). This is the same machinery pointed at the corpus the product actually targets.

### 1. Dark-on-light preprocessing
`_prepared_crops` yields greyscale and hard-threshold passes, both assuming light-on-dark. True for the baseball scorebug, TV Rain's headline banner and the clock badge. **False for TV Rain's ticker and poll panel**, which are two of three overlay types on that corpus.

Needs an inverted pass, but OCR dominates the pipeline so a third rendering per band is a real cost. Wants polarity detection or adaptive thresholding rather than an unconditional third pass, with the cost measured.

### 2. Clock-badge anchor
TV Rain burns `21:35 МСК` into the frame. Reading it yields `anchor = badge_wall_clock - video_t` directly.

This is the highest-value piece. For baseball we derived one anchor by hand: a scorebug count transition cross-referenced against a GUMBO feed, then a drift check at the other end of the game. Hours of work, and only possible because a structured feed existed. Here it falls out of the frame.

Requirements: reject implausible reads (a wrong anchor poisons every downstream citation), corroborate across frames and report the spread, detect drift between start and end, treat the timezone as parsed data rather than an assumption, and decline cleanly on files with no badge.

### 3. Scrolling-text collapse
`collapse_sightings` merges observations by identity, which works for a static overlay. The news ticker **scrolls**, so consecutive frames yield different clipped windows of the same sentence and each becomes its own cue. An archive would show dozens of near-duplicate events and no single span would cover the headline.

Needs similarity-based collapsing, ideally reconstructing the fullest form of the text, without changing behaviour for static overlays.

### 4. A news interpreter (not yet started)
The reader takes an interpreter that returns `(value, hits)`; `scorebug.py` is the baseball one. This corpus needs its own: headline and ticker text rather than roster identity, emitting cues that carry the text and its span.

Note from #744: the interpreter is not optional decoration. It supplies the hit count that drives band lock, so a reader with a weak interpreter locks onto the crowd.

### 5. Into retrieval (not yet started)
[...]

Issue #741, opened 2026-08-04
GOAL: make recognition work on archives with no structured feed


yidakra:
Umbrella goal. The Giants pilot answered a narrower question than it looks, and the answer points somewhere bigger.

## What the pilot actually proved

The headline was **99.4% recall / 100% precision** on 344 pitches. Read the source breakdown and that number is mostly `playbyplay`: it measures **wall-clock alignment against an MLB GUMBO feed**, not computer vision. With the feed present, the vision recognizers were almost decoration.

The result that matters is quieter:

| Source | Identity coverage | Needs a feed? |
|---|---|---|
| playbyplay | 100% | **yes** |
| **scorebug** | **96.5%** | **no** |
| face | 16.3% | no |

**Scorebug OCR reaches 96.5% by reading the broadcast's own overlay.** No GUMBO, no roster API, no per-sport integration beyond a name list. It went from contributing literally zero (structurally locked out of the metric, #640) to 298 of 344 credited pitches once it emitted the right event type and its OCR was fixed (#639: precision 0.54 -> 0.99).

## Why that is the real goal

dir2mcp exists to make an *archive* queryable and citable. Almost no archive has a GUMBO feed. A newsroom's tape library, a documentary back-catalogue, a talk-show run: all have burned-in overlays, lower thirds, chyrons, captions, and none have structured play-by-play.

So the goal is to generalise the thing that worked:

> **Given a video archive and nothing else, recover enough grounded identity and event structure to answer questions with citations.**

Baseball with GUMBO was the scaffold that let us measure. The product is the case without it.

## Milestones

1. **Feed-free baseline.** Run the eval with `playbyplay` disabled and report what the vision cascade achieves alone. We have never measured this; the feed has carried every number so far. This is the honest starting line.
2. **Generalise the overlay reader.** The scorebug parser is currently shaped around one broadcast's bug. Establish what is genuinely reusable: region discovery, light-on-dark preprocessing, field parsing, roster matching with a confidence floor. A chyron and a lower third are the same problem.
3. **Prove it on a second corpus.** The shared Drive already has `AI-Datasets: Content for Demo` with `Documentary/`, `News/` and `Talk-Shows/` subfolders. Different overlays, different identity sources, no feed at all.
4. **Decide the role of faces** (#739). At 16.3% against scorebug's 96.5% they may be worth keeping only for what the overlay cannot say: who is in the dugout, who is on screen during a non-pitch segment.
[...]

PR #747, opened 2026-08-04
feat(annotator): derive a wall-clock anchor from a burned-in clock badge (#741 milestone 3)

yidakra:
Part of #745 / #741 milestone 3. A consumer of the generic overlay reader from #744.

## Why this is the highest-value piece

Aligning video time to wall-clock time is the prerequisite for grounding anything in an archive. For the baseball pilot we derived **one** anchor by hand: found a scorebug count transition, cross-referenced it against an MLB GUMBO feed, computed `anchor = pitch_epoch - video_t`, then verified drift by repeating the exercise at the other end of a 3h24m file. Hours of work, and only possible because a structured feed existed.

News burns the clock into the frame. This reads it.

## Verified on real footage

A 4 minute TV Rain talk-show segment, streamed, `lang="rus"`, `zones={"МСК": "Europe/Moscow"}`:

```
clock reads: 13   bands_read: 108   conflicts: 0
anchor: lower_s=78904.0  upper_s=78906.0
```

**Video t=0 is 21:55:04 MSK, to within one second**, from a badge that only shows minutes.

It gets there by intersecting constraints from successive minute-flips rather than trusting any single reading:

| observation | implies anchor in |
|---|---|
| badge shows 21:55 at t=48, 21:56 at t=56 | `[78904, 78912)` |
| badge shows 21:56 at t=114, 21:57 at t=116 | `[78904, 78906)` |
| intersection | **`[78904, 78906)`** |

That is the same technique used by hand for the baseball anchor (find the transition, not the value), automated, and it converges as more flips are seen. I verified the arithmetic independently.

## Design points that matter

**Zones are required, with no default.** A badge label the reader does not know is a badge it declines to read. The worst failure mode here is a confident wrong answer: a bad anchor silently poisons every downstream citation, so declining beats guessing.

**The interpreter counts a band as a hit only when a plausible zone-labelled time comes out of it.** That is what keeps the band search off the crowd, the ticker and the poll panel, all of which return text and none of which return a time beside a known zone label. Per #744, the hit count drives band lock and cannot be the reader's own judgement.

**Corroboration, spread, and conflicts are reported**, not averaged away. `conflicts=0` above; a file with edits that imply different anchors should say so rather than produce a plausible mean.

**Timezone is parsed data.** `МСК` is not hardcoded anywhere; the caller says what its corpus prints.

## Status

The agent that wrote this hit a session limit before finishing its own validation pass. I picked it up, verified it end to end on real footage, confirmed the anchor arithmetic by hand, and ran the gates: **191 tests pass** (154 before), ruff clean, zero em dashes in authored lines, submodule pin untouched.

**One rough edge I would fix before merge**: `ClockAnchor.__repr__` dumps every reading, so printing one anchor emits hundreds of lines (visible in the raw output above). It wants a compact repr with the readings behind an accessor.

## Summary by CodeRabbit

* **New Features**
  * Added burned-in clock recognition from OCR overlays.
  * Supports labeled time zones, optional dates, unlabeled clocks, midnight transitions, daylight-saving changes, and edited-file drift.
  * Provides wall-clock and epoch-time conversion with uncertainty details.
  * Rejects insufficient or conflicting readings instead of producing unreliable anchors.

* **Documentation**
  * Added setup and usage guidance, including clock positioning, required inputs, OCR configuration, and troubleshooting advice.

### DIFF
--- a/annotator/README.md
+++ b/annotator/README.md
@@ -56,6 +56,42 @@ DIRSTRAL_ANNOTATOR_OCR_LANG=rus dirstral-annotate serve …   # or lang="rus"
 Non-Latin scripts need the matching tesseract traineddata installed
 (`brew install tesseract-lang`, `apt install tesseract-ocr-rus`).
 
+### The burned-in clock reader
+
+`recognizers/clock.py` is one consumer of that reader, and the one that pays
+for itself fastest on an archive: a video timestamp is only citable once
+something says what wall-clock instant it was, and a news broadcast burns that
+into the corner of the frame.
+
+```python
+from dirstral_annotator.recognizers.clock import ClockReader
+
+reader = ClockReader(zones={"MSK": "Europe/Moscow"}, anchor_date=date(2025, 8, 20))
+anchor = reader.anchor(Path("broadcast.mp4"))   # None when there is no badge
+anchor.wall_clock_at(1830.0)                    # aware datetime of video 00:30:30
+```
+
+What it will not do is guess. The zone table is required (a badge label the
+caller has not named is not a reading), the date is required for an epoch (a
+badge shows a time of day and no date), a single reading is never an anchor,
+and readings that imply two different anchors are reported as two segments
+rather than averaged into one that is wrong for both.
+
+Two settings are measured rather than stylistic, and both differ from the
+scorebug's:
+
+- **psm 11, not 6.** A badge is an island of text in an empty corner, not a
+  dense strip. On sample footage psm 6 never read the badge at any crop size;
+  psm 11 read it.
+- **The OCR language changes the digits.** On the same pixels, a badge showing
+  `21:35` read as `21:35` under `eng` and as `21:55` under `rus`. Neither
+  reading is detectably wrong on its own, and a 20 minute error in an anchor is
+  inherited by every citation under it. Do not hand the clock reader the
+  language you use for the prose overlays without checking it against a frame,
+  and list the zone label the way your OCR actually renders it (a Latin-script
+  model returns `MCK` for Cyrillic `МСК`, so a Russian corpus may want both
+  spellings in the table).
+
 ## Run the backend
 
 ```bash
--- a/annotator/README.md
+++ b/annotator/README.md
@@ -64,11 +64,15 @@ something says what wall-clock instant it was, and a news broadcast burns that
 into the corner of the frame.
 
 ```python
+from datetime import date
+from pathlib import Path
+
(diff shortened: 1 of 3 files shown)


yidakra:
## Review fixes pushed (66aea2f), plus what chasing one of them turned up

All six CodeRabbit findings addressed. One was more interesting than it looked.

### `_zone_beside` matched a label inside a word

Confirmed, and there were **two** open cases, not one:

- `11:20 ETA 5 min` folds to `ETA5MIN`, which starts with `ET` (the reported one)
- `TICKET 11:20` folds to `TICKET`, which ends with `ET`

The existing test passed only by accident of position: `TICKET` placed *after* the time does not START with `ET`, so it never exercised the trailing path at all.

The suggested repair (fold into tokens, compare whole tokens) would undo what the fold is for. `11:20 E T` is what OCR returns for a letter-spaced badge, and a token comparison rejects it.

So I checked the boundary on the folded window instead, and **that broke a real badge**: `21:35 MSK LIVE`. Once the space is folded out, `MSK LIVE` and `MSKVA` both continue with a letter, and the fold cannot tell them apart. An existing test caught it, not inspection.

The fix that holds: the fold keeps each surviving character's source index, so the boundary is read off the text **as printed** while the fold still does its job. Twelve cases verified, including `E T`, `(ET)`, a leading `ET 11:20`, and `MSK LIVE`.

### Does it change anything on real footage?

No, and I checked rather than assuming. Dumped every raw OCR pass from the real TV Rain clip and ran old and new logic over all of them:

```
raw OCR passes: 212
time matches labelled: old=165  new=165   disagreements=0
```

Same conclusion as the LCS fix on #746: a correctness fix for shapes this footage does not contain. Reader end-to-end: 106 readings, 0 conflicts, Cyrillic МСК still matching.

### The rest

- `anchor_from` claimed its counters are zero for readings it did not take. They carry over from the last `read_clock` on the instance. Corrected.
- README example imported neither `date` nor `Path` and dereferenced an `anchor` the line above documents as possibly `None`. Both fixed, plus the compound-modifier hyphen.
- The tz-database guard: applied to both sites via a `has_tzdata` helper, since `clock.py` names a host without tzdata as supported and fixed offsets exist to serve it.
- The double-`ClockReader` test now uses one instance.

Also my own loose end, unrelated to the review: `ClockSegment.readings` and `ClockRead.texts` sat in the generated repr, so `print(anchor)` on a real file dumped every OCR pass of all hundred-odd readings. Both are out of the repr now, with the reading count in their place.

194 tests pass, ruff clean.

PR #756, opened 2026-08-05
fix(ingest): localize media through the CorpusFS before recognition (#734)

### LINKED ISSUE #734: ingest: video recognition bypasses CorpusFS and fails for S3 corpora
## Summary

Video recognition constructs a local path from `RootDir + rel_path` instead of obtaining a local copy from the configured `CorpusFS`. For `source.kind=s3`, the recognizer therefore receives a nonexistent local path and recognition fails even though the object was discovered and ingested successfully.

## Evidence

`internal/ingest/recognize.go` currently does:

```go
absPath := filepath.Join(s.cfg.RootDir, filepath.FromSlash(doc.RelPath))
result, err := s.recognizer.Recognize(ctx, absPath)
```

This is the only media-processing path found that reconstructs the source path this way. Archive extraction, transcript/STT helpers, OCR, and other media paths call `s.corpusFS().Localize(ctx, doc.RelPath)`, which downloads an S3 object to a temporary local file and returns cleanup.

The S3 backend deliberately emits `DiscoveredFile.AbsPath == ""`; its `Walk` ignores `RootDir`, and the service injects the configured S3 `CorpusFS`. An ordinary object such as `videos/game.mp4` therefore has no corresponding `<RootDir>/videos/game.mp4` file.

Existing recognition tests use local-looking paths and do not inject a remote CorpusFS, so this backend mismatch is not covered.

## Reproduction

1. Configure `source.kind=s3` with a video object and enable a recognition provider.
2. Run ingest.
3. The object is enumerated and its bytes can be read through the S3 CorpusFS.
4. Recognition is invoked with `filepath.Join(cfg.RootDir, "videos/game.mp4")`, not a localized copy of the object.
5. The recognition backend cannot open that local path; the representation is absent and the document may be recorded with a recognition failure.

## Expected behavior

Localize the document through the active `CorpusFS` before calling the recognizer, defer the returned cleanup, and preserve the current local-filesystem behavior via `LocalFS.Localize`.

Add a remote-CorpusFS recognition test that asserts the recognizer receives the localized path and cleanup runs on both success and error.

yidakra:
Closes #734.

## What was actually wrong

`generateRecognitionRepresentation` built the media path itself:

```go
absPath := filepath.Join(s.cfg.RootDir, filepath.FromSlash(doc.RelPath))
result, err := s.recognizer.Recognize(ctx, absPath)
```

I verified the framing in the issue rather than trusting it:

- `S3FS.Walk` emits `DiscoveredFile{AbsPath: ""}` (documented on the struct: "for object-store backends (S3FS) it is empty because there is no local path until Localize is called") and never consults the `root` argument. So for `source.kind=s3` there is no `<RootDir>/videos/game.mp4`, and the recognizer was handed a dangling path. Confirmed by a failing test, not by reading: see below.
- Recognition really was the only media path doing this. `grep` for `corpusFS().Localize` finds archive extraction, `probeDuration`, `warnMultiTrackAudio`, `detectLeadingSilence`, `probeTrackInfo`; recognition was the outlier.

**One correction to the issue text:** it says OCR calls `Localize`. It does not. OCR/extractors take bytes (`extractor.Extract(ctx, relPath, content)`), and those bytes come from `corpusFS().Open`. So OCR was never broken and is not the precedent for the cleanup contract; the archive/probe paths are. The conclusion is unchanged: everything else goes through the CorpusFS one way or another, and only recognition bypassed it.

## The fix

`internal/ingest/recognize.go`: resolve the media with `s.corpusFS().Localize(ctx, doc.RelPath)` and `defer cleanup()`, matching the archive and probe paths. Cleanup is deferred immediately after the error check, so it runs on the success path, the backend-error path, and every early return in between; `cleanup` is only ever deferred when `err == nil`, and both `LocalFS` and `S3FS` return a nil cleanup only alongside an error.

Two judgement calls, flagged rather than buried:

1. **A localize failure is tagged `ErrRecognitionProviderFailure`.** The backend was not reached, but the outcome for the document is identical to a backend failure (no recognition, retriable next scan), and today a missing local file already surfaces as `RECOGNIZE_FAILED` via the backend's own error. The precedent is `readTrackTranscript`, which tags a failed *local* audio-track extraction with `ErrTranscriptProviderFailure` so it lands in `TRANSCRIBE_FAILED` instead of the generic `EXTRACT_FAILED`. Without the sentinel this class would silently reclassify to `EXTRACT_FAILED`.
2. **The localized path is absolutized.** Design 0004 §5 specifies `{"path": "<absolute media path>"}`. `LocalFS.Localize` already returns an absolute path, but `S3FS.Localize` materializes under `StateDir/corpus-cache`, and several CLI paths default `StateDir` to a relative `./.dir2mcp` (`internal/cli/ask.go`, `status.go`, `export.go`, ...). A connect-only backend does not share the daemon's CWD. This goes slightly beyond the literal issue, so it has its own test.

## Async re-read: checked, not assumed

The concern was whether the backend re-reads the path asynchronously after the call, which would make `defer cleanup()` too early. It does not, on both sides of the contract:

- `RecognizeServeClient.Recognize` does a blocking `httpClient.Do` and fully decodes the response body into `model.RecognizeResult` before returning; nothing retains the path.
- The reference backend in this repo (`annotator/dirstral_annotator/serve.py`) runs `pipeline.annotations_for(media)` inline inside the `POST /recognize` handler and replies afterwards. The file is fully consumed before the HTTP response is written.

So releasing the localized copy when `Recognize` returns is safe. (The backend does echo `media.name` — now a temp basename — in its response `document` field, but `RecognizeServeClient` only reads `recognizer` and `annotations`, so nothing downstream sees it.)

## Local corpora: measured, no regression

`LocalFS.Localize` returns the resolved in-root path with a `func(){}` cleanup: no copy, no download, no I/O beyond the containment resolution `Open` already performs on every read of the same file. `TestRecognize_LocalCorpusStillGetsInRootPath` asserts with `os.SameFile` that the recognizer is handed the corpus file itself and that the file is still there after the call.

**One visible behavioural difference for local corpora, deliberate:** `LocalFS.Localize` resolves symlinks (`resolveWithinRoot` → `EvalSymlinks`) and rejects paths escaping the root, so the recognizer now receives e.g. `/private/var/...` where it previously received `/var/...` on macOS. That is the same resolution every other media path already applies, and it adds the corpus-containment guard recognition lacked. It broke exactly one existing assertion, `TestRecognize_GeneratesRecognitionRepresentation`'s string comparison against `filepath.Join(root, ...)`; that assertion is now `os.SameFile`, which tests identity instead of resolution style.

## Testing: failed first, then fixed

New suite `tests/ingest/recognize_corpusfs_734_test.go`. Written and run **before** the fix. Against unmodified `recognize.go`:

```
--- FAIL: TestRecognize_RemoteCorpusFS_CleanupOnBackendError
    Localize called 0 time(s), cleanup 0 time(s); want exactly 1 each on the error path
--- FAIL: TestRecognize_LocalizeFailureIsRecognizeFailure
    expected a localize failure to be reported
--- FAIL: TestRecognize_RemoteCorpusFS_FeedsLocalizedPath
    recognizer could not read the path it was handed
    (".../TestRecognize_.../001/games/game7.mp4"): no such file or directory
```

That third failure is the bug itself, reproduced: the recognizer double actually `os.ReadFile`s the path it is given, so the test fails for the real reason (an unopenable path) rather than on a string comparison.

`TestRecognize_LocalizedPathIsAbsolute` was likewise verified to fail without the `filepath.Abs` step (backend received `../../../../../var/folders/...`).

`TestRecognize_LocalCorpusStillGetsInRootPath` passes both before and after, by design: it is the no-regression guard.

Coverage:

| test | asserts |
| --- | --- |
| `TestRecognize_RemoteCorpusFS_FeedsLocalizedPath` | recognizer gets the localized copy (not the RootDir join), can read it, gets the object's bytes; exactly 1 Localize + 1 cleanup; temp file gone afterwards |
| `TestRecognize_RemoteCorpusFS_CleanupOnBackendError` | cleanup still runs and the copy is removed when the backend fails; error still `errors.Is` `ErrRecognitionProviderFailure` |
| `TestRecognize_LocalizeFailureIsRecognizeFailure` | unmaterializable media is reported (not silently empty), classifies as `RECOGNIZE_FAILED`, and the backend is never called |
| `TestRecognize_LocalizedPathIsAbsolute` | a relative `Localize` result is absolutized before the backend call |
| `TestRecognize_LocalCorpusStillGetsInRootPath` | local corpus still gets the in-root file itself (`os.SameFile`), no copy |

The remote double (`localizingRemoteFS`) mirrors S3FS on the properties that matter here: empty `AbsPath`, RootDir-ignoring `Walk`, temp materialization plus a removing cleanup. I did not reuse `tests/corpusfs/s3_test.go`'s fake S3 client: it lives in a different test package, and driving a real `S3FS` through ingest would exercise the AWS stub surface rather than the seam under change (`Service.SetCorpusFS`).

## Deliberately not done

- **No end-to-end S3 ingest test.** The seam that broke is `Service` → `CorpusFS`, and `SetCorpusFS` exercises it directly. Standing up a fake S3 client through a full scan would add a lot of surface without touching the defect.
- **No `MediaURL` fast path for recognition.** `S3FS` can presign a range-seekable URL and the embedding worker prefers it over a whole-object download. The recognize wire contract is a filesystem path, and a recognizer samples frames across the entire video anyway, so `Localize` is the right call; a URL-capable recognizer would be a spec change (design 0004 §5).
- **No spec change.** Design 0004 §5 already says "absolute media path"; nothing here alters the wire contract or the persisted representation, so no `dirstral-spec` PR is needed under the spec-first rule.
- **No change to the other `Localize` call sites.** They were already correct.

## Checks

- `make check` passes locally (full suite: `gofmt`/`vet`/`lint`/`cyclo`/`ineffassign`/`misspell`/tests/build). The touched function stays well under the `gocyclo -over 15` budget (+2 branches).
- `coderabbit review --agent --committed --base main`: 0 findings.
- No new files under `internal/cli/`, so the `tests/security/cli_boundary_test.go` allowlist is untouched.

### DIFF
--- a/internal/ingest/recognize.go
+++ b/internal/ingest/recognize.go
@@ -316,7 +316,36 @@ func (s *Service) generateRecognitionRepresentation(ctx context.Context, doc mod
 	if s.repGen == nil || s.recognizer == nil || doc.DocType != "video" {
 		return false, nil
 	}
-	absPath := filepath.Join(s.cfg.RootDir, filepath.FromSlash(doc.RelPath))
+	// Recognition backends read the media from a real filesystem path, so the
+	// media must be resolved through the CorpusFS rather than reconstructed from
+	// RootDir+rel_path (#734): an object-store backend reports no local path
+	// (DiscoveredFile.AbsPath == "") and its Walk ignores RootDir, so a joined
+	// path names a file that does not exist and recognition fails on a corpus
+	// that ingested fine. Localize returns the resolved in-root path with a
+	// no-op cleanup for a local corpus (no copy, unchanged behavior) and a temp
+	// download for an object store. The Recognizer contract is synchronous — the
+	// backend has finished reading the file by the time Recognize returns — so
+	// releasing the copy on return is safe.
+	localPath, cleanup, err := s.corpusFS().Localize(ctx, doc.RelPath)
+	if err != nil {
+		// The backend was never reached, but the outcome for this document is the
+		// same as a backend failure — no recognition, retriable next scan — so it
+		// carries the same sentinel and classifies as RECOGNIZE_FAILED rather than
+		// the generic EXTRACT_FAILED (parallel to the STT path, which tags a failed
+		// audio-track extraction as a transcript provider failure).
+		return false, fmt.Errorf("%w: localize %s: %w", ErrRecognitionProviderFailure, doc.RelPath, err)
+	}
+	defer cleanup()
+
+	// The serve wire contract is an ABSOLUTE media path (design 0004 §5). LocalFS
+	// already returns one; an object store's temp copy lands under StateDir, which
+	// can be configured relative (e.g. "./.dir2mcp"), so absolutize before handing
+	// the path over — a connect-only backend does not share the daemon's CWD.
+	absPath, err := filepath.Abs(localPath)
+	if err != nil {
+		return false, fmt.Errorf("%w: resolve media path for %s: %w", ErrRecognitionProviderFailure, doc.RelPath, err)
+	}
+
 	result, err := s.recognizer.Recognize(ctx, absPath)
 	if err != nil {
 		// Tag with the provider-failure sentinel so a transient recognize-backend
--- a/tests/ingest/recognize_corpusfs_734_test.go
+++ b/tests/ingest/recognize_corpusfs_734_test.go
@@ -1,0 +1,343 @@
+package tests
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"io"
+	"os"
+	"path/filepath"
+	"sync"
+	"testing"
+
+	"github.com/dirstral/dir2mcp/internal/config"
+	"github.com/dirstral/dir2mcp/internal/corpusfs"
+	"github.com/dirstral/dir2mcp/internal/ingest"
+	"github.com/dirstral/dir2mcp/internal/model"
+)
+
+// localizingRemoteFS is a stub object-store CorpusFS shaped like S3FS for the
+// properties that matter to issue #734: discovery reports no local path
+// (DiscoveredFile.AbsPath == ""), Walk ignores the RootDir argument, and
+// Localize materializes a temp copy under its own cache dir and returns a
+// cleanup that removes it. It records Localize/cleanup calls so a test can
+// assert the recognizer is fed the localized copy and that the copy is released
+// on both the success and the failure path.
+type localizingRemoteFS struct {
+	bodies   map[string][]byte
+	cacheDir string
+
+	mu            sync.Mutex
+	localizeCalls int
+	cleanupCalls  int
+	localizeErr   error
+	lastLocalPath string
+}
+
+func newLocalizingRemoteFS(t *testing.T) *localizingRemoteFS {
+	t.Helper()
+	return &localizingRemoteFS{bodies: map[string][]byte{}, cacheDir: t.TempDir()}
+}
+
+func (f *localizingRemoteFS) add(relPath, body string) {
+	f.bodies[relPath] = []byte(body)
+}
+
+func (f *localizingRemoteFS) Walk(_ context.Context, _ string, _ corpusfs.Options) ([]corpusfs.DiscoveredFile, error) {
+	out := make([]corpusfs.DiscoveredFile, 0, len(f.bodies))
+	for rel, body := range f.bodies {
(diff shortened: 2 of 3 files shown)


yidakra:
@coderabbitai review

Issue #737, opened 2026-08-03
identity: S3 corpora in the same working directory share the same default server and service name


yidakra:
## Summary

The automatic instance identity is derived only from `RootDir`. For native S3 sources, `RootDir` is not the corpus identity and normally defaults to the process working directory. Two different bucket/prefix corpora launched from the same directory therefore receive the same MCP server name, service label, and default client alias.

## Evidence

`resolveServerName` always computes:

```go
abs, _ := filepath.Abs(cfg.RootDir)
cfg.ServerName = identity.Resolve(abs, cfg.ServerName, buildinfo.IsDev())
```

`resolveClaudeServerName` and `serviceContextFromConfig` reuse that value. The identity package is explicitly documented as deriving a unique instance name from the absolute path of the indexed directory.

For `source.kind=s3`, however, `S3FS.Walk` ignores `RootDir`; the logical corpus root is `bucket + prefix`. The deployment guide's S3 configuration does not require a distinct local root, so separate S3 configs commonly share the same default `.`.

## Reproduction

From one local directory, use two config files:

- bucket `customer-a`, prefix `corpus/`;
- bucket `customer-b`, prefix `corpus/`.

Leave `server_name` and `root_dir` at their defaults. Both configurations resolve the same absolute local `RootDir`, so both derive the same:

- MCP `serverInfo.name`;
- Claude MCP alias;
- launchd/systemd service label.

Installing or registering the second instance can overwrite/collide with the first, and diagnostics cannot distinguish which remote corpus a connection serves.

## Expected behavior

Derive the default identity from a canonical source identity:

- local/NFS: normalized absolute root path, preserving current behavior;
- S3: a stable, non-secret identifier based on bucket + normalized prefix (and endpoint/partition where needed to avoid cross-store collisions).

Never include credentials in the slug or hash input exposed to diagnostics. Explicit `server_name` overrides must remain authoritative.

Add tests proving distinct bucket/prefix pairs get distinct stable names, relocation-relevant normalization is deterministic, and local behavior is unchanged.

Issue #736, opened 2026-08-03
export: SMIL probing bypasses CorpusFS and is silently omitted for S3 media


yidakra:
## Summary

TTML export's optional SMIL companion probes `filepath.Join(cfg.RootDir, rel_path)` directly. For an S3 corpus there is no local source file at that path, so SMIL is silently omitted even when `media.subtitles.smil.enabled` is enabled and the remote media is readable through the configured `CorpusFS`.

## Evidence

`internal/cli/export_ttml.go::emitSMILSidecar` does:

```go
mediaPath := filepath.Join(cfg.RootDir, opts.relPath)
info, err := avutil.ProbeMediaInfo(ctx, mediaPath)
if err != nil {
    // fail open and omit SMIL
    return
}
```

The export command loads source configuration but never builds or injects the configured CorpusFS. By design, S3 discovery emits no local `AbsPath`; callers that require a filesystem path must call `CorpusFS.Localize`.

Because SMIL probe errors intentionally fail open, the command exits successfully after writing TTML. An S3 operator therefore gets no SMIL artifact and only a transient human-readable note (suppressed by `--quiet` and JSON mode), despite explicitly enabling the feature.

This differs from #494, which requests writing generated artifacts back to S3. This issue is about reading/probing the remote source media while writing the requested artifacts locally. It is also separate from #734's ingest-time recognition path.

## Reproduction

1. Configure `source.kind=s3` with an indexed video and stored transcript.
2. Enable TTML and SMIL export.
3. Run `dir2mcp export --format ttml --out game.ttml videos/game.mp4`.
4. `game.ttml` is written successfully.
5. The command probes `<local RootDir>/videos/game.mp4`, which does not exist, and silently produces no `game.smil`, although the S3 object is available through CorpusFS.

## Expected behavior

Build the configured CorpusFS for export, localize the media before probing, defer cleanup, and keep the current fail-open policy for genuine probe failures.

Add an export test with a remote CorpusFS stub proving that:

- the localized media is probed;
- the SMIL file is produced;
- cleanup runs;
- JSON/quiet behavior remains machine-safe on a real failure.

Unsafe traversal rel paths must also be rejected at the source boundary as tracked in #735.

Issue #730, opened 2026-08-03
export: TTML emits empty language metadata when --lang is omitted


yidakra:
## Summary

`export --format ttml` documents `--lang` as optional and correctly selects the first/source transcript when it is omitted, but the TTML renderer is passed the raw empty flag value rather than the selected representation's recorded language.

The result contains `xml:lang=""`, cue language fields are empty, and a companion SMIL omits `systemLanguage`, even when the transcript metadata records a language such as `en`.

## Evidence

- `selectTranscriptRep(reps, "")` selects the first representation.
- `resolveCues` returns only cues, discarding the chosen representation/meta language.
- `runTTMLExport` calls:
  - `MonolingualBilingualCues(primaryCues, opts.lang)`
  - `RenderTTML(bilingual, opts.lang)`
- `emitSMILSidecar` also uses `opts.lang`.

When `--lang` was omitted, all of those receive `""`.

Existing TTML tests always pass `--lang en`, so the documented default path is uncovered.

## Reproduction

With a transcript representation whose `meta_json` contains `{"language":"en"}`:

```bash
dir2mcp export --format ttml media/talk.mp3
```

Observed root:

```xml
<tt ... xml:lang="">
```

Expected: `xml:lang="en"` (and `systemLanguage="en"` in SMIL).

## Expected behavior

Resolve the effective primary/secondary language from the selected transcript representation and carry it alongside the cues through TTML, bilingual alignment, and SMIL packaging. An explicit `--lang` may select/echo that tag; omission must use the recorded tag.

For a genuinely untagged legacy transcript, define whether `xml:lang` is omitted rather than emitted with an empty value.

## Suggested regression coverage

- Omitted `--lang` + tagged transcript emits its recorded language.
- SMIL uses the same resolved language.
- Bilingual behavior retains both resolved tags.
- Untagged legacy transcript has a valid documented fallback.

PR #758, opened 2026-08-05
fix(s3): stop requiring a local corpus root at startup and service install (#738)

### LINKED ISSUE #738: s3: startup and service install still require an unrelated local RootDir
## Summary

Native S3 is documented as a remote corpus whose object prefix is the corpus root, but startup unconditionally requires `cfg.RootDir` to exist as a local directory. Service installation also uses that unrelated path as the supervisor working directory and as the base for relative state/config material.

A remote-only configuration can therefore fail before S3 is touched, or install a service that cannot start, solely because a local corpus root that the S3 backend ignores is absent.

## Evidence

- `S3FS.Walk` explicitly ignores its `root` argument; bucket + prefix define the corpus root.
- Canonical SPEC §7.8 says the corpus root may live on S3 and only the state directory must stay local.
- `validateUpConfig` nevertheless always calls `ensureRootAccessible(cfg.RootDir)`.
- `serviceContextFromConfig` always sets `workingDir = abs(cfg.RootDir)`.
- launchd and systemd render that value as `WorkingDirectory`.
- the S3 deployment example configures bucket/prefix and does not describe any required local corpus directory.

## Reproduction

1. Configure a valid `source.kind=s3`, bucket, prefix, credentials, and a valid local absolute `state_dir`.
2. Set `root_dir` to a nonexistent path, or remove the former local corpus after migrating it to S3.
3. Run `dir2mcp up`.
4. Startup exits with `root inaccessible` before the configured remote corpus can be listed.
5. A service definition based on the same configuration points its `WorkingDirectory` at that nonexistent local corpus path and cannot launch.

The remote object store and local state directory may both be healthy.

## Expected behavior

Make local-root validation and working-directory selection source-aware:

- local/NFS: retain the current accessible-root requirement;
- S3: do not require a local corpus root; require only the local state/config locations that are actually used;
- choose a stable existing service working directory such as the config directory or state-directory parent;
- resolve relative config/state paths before supervisor installation.

Add startup and service-rendering tests for an S3 source with no local corpus directory. Instance naming for remote sources is separately tracked in #737.

yidakra:
Closes #738.

## The brief was right, and here is what I measured

Both claims in the issue check out against the code:

- `internal/corpusfs/s3.go:199` — *"The root argument is ignored: the configured prefix is the corpus root."*
- `dirstral-spec/docs/SPEC.md` §7.8 — *"State stays local. Regardless of `source.kind`, the state directory ... is always local ... Only the corpus content is remote."*
- `internal/cli/up.go` `validateUpConfig` called `ensureRootAccessible(cfg.RootDir)` unconditionally.
- `internal/cli/service.go` `serviceContextFromConfig` always set `workingDir = abs(cfg.RootDir)`, which `renderLaunchdPlist`/`renderSystemdUnit` emit as `WorkingDirectory` (and systemd also as `EnvironmentFile=<workdir>/.env.local`).

Measured before writing any fix, on the pre-fix tree:

```
$ go test ./tests/cli -run 738
    s3 source must not require a local corpus root (#738); got: root inaccessible:
    stat .../does-not-exist: no such file or directory

$ go test ./internal/cli -run TestProbePreFixServiceWorkingDir   # throwaway probe
    workingDir=.../does-not-exist stateDir=.../state
    PRE-FIX DEFECT: rendered WorkingDirectory does not exist
```

No spec change is needed: §7.8 already permits a remote corpus root with a local state directory, so this brings the implementation into conformance rather than introducing new normative behaviour.

## What changed

**1. Local-root validation is now source-aware.** `validateUpConfig` gates `ensureRootAccessible` on the existing `sourceIsRemote(cfg)` helper (`internal/cli/up.go:1357`), which reads the resolved `source.kind` (`config.validateSource` normalizes/validates it and defaults empty to `local`; the same helper already routes retrieval-time reads for #432). `local`, `nfs`, and empty keep the requirement verbatim — an NFS mount is an ordinary local path. The state directory is created and hardened exactly as before, for every kind.

The directory block moved into a new `prepareUpDirectories` helper purely to keep `validateUpConfig` under the gocyclo budget (adding the branch pushed it to 16, over the `-over 15` gate). No behavioural change; the ordering relative to x402 validation is preserved, and the reason for it is now in the helper's doc comment.

**2. Service working directory for a remote corpus: the config file's directory.**

I considered the two candidates the issue names. The state-directory parent is the wrong answer, and I want the reasoning on record. The working directory is not cosmetic — it is the resolution base for three things:

- `config.load` reads dotenv via `loadDotEnvFiles([]string{".env.local", ".env"}, ...)` — **relative**, i.e. against the process cwd.
- `resolveConfigPath` defaults to `".dir2mcp.yaml"` — **relative**, against the process cwd.
- `persistAndWarnCredentials(sc.workingDir, ...)` writes `.env.local` there, and systemd renders `EnvironmentFile=<WorkingDir>/.env.local`.

If the booted daemon's cwd were the state directory, then unless the operator happened to pass an explicit `--config`, the daemon would find no `.dir2mcp.yaml` there and boot on **pure defaults** — including `source.kind: local`. That silently converts an S3 deployment into a local one. The config file's own directory is the only choice that keeps all three consistent: the daemon rediscovers the same config and the same `.env.local` the operator installed from, which is also where `dir2mcp config init` writes credentials.

It also demonstrably exists: it is `filepath.Dir` of the absolute resolved config path, i.e. the installing process's cwd in the default case. For the one case where it may not exist (an explicit `--config` naming a path under a missing directory), the fallback is the **state directory** — always local per §7.8, and already `MkdirAll`'d by `runServiceInstall` before the unit is written. Both are asserted by tests.

**3. Coupled path resolution.** A relative `state_dir` is resolved by the booted daemon against *its* working directory, so install must use the same base or it would create and log into a different directory than the daemon opens. `resolveServiceDirs` now returns both together. For `local`/`nfs` the base is still `abs(RootDir)`, so this is byte-identical to before.

**4. Propagated `--config`/`--state-dir` are absolutized.** A relative `--config` was copied verbatim into the unit and then re-resolved by the daemon against its own working directory, which is not the cwd the operator ran install from. `--state-dir` now propagates the already-resolved `sc.stateDir` (the same directory install creates and writes `service.log` into), which is semantically identical to the old relative value whenever the daemon cwd equals the install base — i.e. always, for local corpora. This is the substance of **#721** (*"service: relative --config and --state-dir paths are reinterpreted after install"*), which I did not set out to fix: it became load-bearing here, because moving the remote working directory off the corpus root changes what a relative `--config` resolves to. **#721 can be closed or re-scoped against this PR.** The only case where a local corpus behaves differently is `--config <relative>` invoked from a cwd other than the corpus root, where the previous unit pointed at a path that does not exist.

## Local corpora do not regress

`TestServiceContext_LocalAndNFSKeepCorpusRootWorkingDir_738` pins `kind` in `{"", "local", "nfs"}` to `workingDir == abs(RootDir)` and `stateDir == RootDir/.dir2mcp` even when the config file lives elsewhere. `TestUpLocalSourceStillRequiresLocalRoot_738` and `TestUpNFSSourceStillRequiresLocalRoot_738` pin that a missing root is still a hard `root inaccessible` startup failure. Both of those two passed on the pre-fix tree as well, which is the point — they are the control.

## Tests

`tests/cli/s3_no_local_root_738_test.go` (new):
- `TestUpS3SourceDoesNotRequireLocalRoot_738` — **fails on the pre-fix tree** with `root inaccessible` (output above). Drives the real `up --foreground` path with a valid s3 config, absent local root, and no embed provider, so the assertion is bounded: startup must reach the embed preflight (the very next check) and fail *there* instead.
- `TestUpLocalSourceStillRequiresLocalRoot_738`, `TestUpNFSSourceStillRequiresLocalRoot_738` — controls; pass before and after.

`internal/cli/service_test.go` (appended to the existing, already-allowlisted file, so `tests/security/cli_boundary_test.go` needs no change — no new file under `internal/cli/`):
- `TestServiceContext_S3AnchorsWorkingDirOnConfigDir_738`
- `TestServiceContext_S3RelativeStateDirFollowsWorkingDir_738`
- `TestServiceContext_S3FallsBackToStateDirWhenConfigDirMissing_738`
- `TestServiceContext_LocalAndNFSKeepCorpusRootWorkingDir_738`
- `TestRenderedUnitsUseAnExistingWorkingDirForS3_738` — renders both the launchd plist and the systemd unit and asserts the rendered `WorkingDirectory` exists and that neither unit references the ignored local corpus root anywhere.
- `TestInstallServiceArgs_PropagatesAbsolutePaths_738`

These service-layer tests are written against the new `serviceContextFromConfig(cfg, nameOverride, configPath)` signature, so on the pre-fix tree they do not compile rather than fail. The defect they encode was verified separately with the throwaway probe quoted at the top, which ran against the unmodified implementation and showed `workingDir` pointing at the nonexistent root.

`make check` passes (vet, golangci-lint 0 issues, gocyclo `-over 15`, misspell, full suite, python).

## Other local-root assumptions I found, and did not fix here

I swept the rest of the startup path. Everything else on it is already remote-safe: `EmbeddingWorker` prefers `Corpus` over `RootDir`, retrieval gets `SetCorpusFS` for remote sources (#432), and `openScanCache` already skips non-local kinds. What remains is out of scope for #738 and, in almost every case, **already tracked**:

- **`ingest.Watch` registers fsnotify on `abs(cfg.RootDir)`** for every kind. For S3 this is not fatal — `addWatchDirs` fails, logs one line, and the loop degrades to the 10-minute safety rescan, which does walk S3 via CorpusFS — but the log line is misleading. Already **#695**.
- **`mcp.Server.resolveDocumentPath`** (on-demand transcribe/annotate) joins and `EvalSymlinks`es `cfg.RootDir`, so those tools cannot read an S3 corpus. `resolveRoot` (list_files) already degrades gracefully. Related open items: **#684**, **#736**, **#734**. The transcribe/annotate path specifically does not appear to be filed — flagging it here rather than filing a near-duplicate.
- **`internal/cli/export_ttml.go:132`** joins `cfg.RootDir` for the media path (same family as **#736** / **#729** / **#730**).
- **`config.load` reads `.env.local` relative to cwd, not to `--config`'s directory** — **#677**. The remote working-dir choice here aligns with, rather than contradicts, whatever #677 lands on.

Deliberately untouched: `resolveServerName` / instance-identity derivation (**#737**). The #738 tests show the derived name still tracking the ignored local root (`dir2mcp-dev-does-not-exist-<hash>`), which is exactly #737's subject; I left it for that PR and shaped the assertions so they do not depend on it.

## Docs

`docs/dual-machine-deployment.md` §2.3 now states plainly that `source.kind: s3` needs no local corpus directory, that only the state directory must be local, and where `service install` anchors the daemon for a remote source.

### DIFF
--- a/docs/dual-machine-deployment.md
+++ b/docs/dual-machine-deployment.md
@@ -184,6 +184,17 @@ persisted** to the config file or the effective-config snapshot. With
 `source.kind: s3`, startup fails loudly (`CONFIG_INVALID`-style) if the bucket is
 missing or the access key + secret are not resolvable.
 
+**No local corpus directory is required for `s3`.** The bucket + prefix *is* the
+corpus root, so `root_dir` (the `--dir` root) is ignored by the S3 backend and
+does not have to exist. Only the **state directory** must be local
+(`dirstral-spec/docs/SPEC.md` §7.8) — that is where SQLite, the index and the
+caches live. `dir2mcp up` on an `s3` source therefore no longer fails with
+`root inaccessible` when no local corpus is present, and `dir2mcp service
+install` anchors the supervised daemon's working directory on the **directory
+holding your config file** (where `.dir2mcp.yaml` and `.env.local` live) instead
+of the ignored corpus root. `local`/`nfs` are unchanged: they still require an
+accessible root directory and still boot the service in it.
+
 `rel_path` (document identity) is stable across schemes: for `local`/`nfs` it is
 the path under the root; for `s3` it is the object key minus the configured
 prefix. A corpus can be relocated `local ⇄ nfs ⇄ s3` without forcing a reindex
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -1402,13 +1402,17 @@ func consumeGlobalFlagValue(flagName string, args []string) (string, int, error)
 	return value, 2, nil
 }
 
+// defaultConfigFileName is the config file `dir2mcp` discovers in the process
+// working directory when no --config is supplied.
+const defaultConfigFileName = ".dir2mcp.yaml"
+
 // resolveConfigPath returns the configured config path, defaulting to
 // ".dir2mcp.yaml" when none was supplied.
 func resolveConfigPath(global globalOptions) string {
 	if trimmed := strings.TrimSpace(global.configPath); trimmed != "" {
 		return trimmed
 	}
-	return ".dir2mcp.yaml"
+	return defaultConfigFileName
 }
 
 // applyGlobalPathOverrides overlays the --dir and --state-dir global flags
--- a/internal/cli/down.go
+++ b/internal/cli/down.go
@@ -86,7 +86,7 @@ func (a *App) runDown(ctx context.Context, global globalOptions, args []string) 
 	// just stopped will be respawned at next login/boot. Inform the operator
 	// (do NOT auto-uninstall) so `down` in a teardown script isn't mistaken
 	// for a permanent stop (#434).
-	writeDownInfo(a.stdout, global.jsonOutput, cfg.StateDir, pid, true, "stopped", a.corpusServiceManaged(cfg))
+	writeDownInfo(a.stdout, global.jsonOutput, cfg.StateDir, pid, true, "stopped", a.corpusServiceManaged(cfg, global))
 	return exitSuccess
 }
 
@@ -95,8 +95,8 @@ func (a *App) runDown(ctx context.Context, global globalOptions, args []string) 
 // unsupported platform) yields false so the informational note is simply
 // omitted and `down` never fails over it. Takes the already-loaded config so it
 // doesn't reload it.
-func (a *App) corpusServiceManaged(cfg config.Config) bool {
-	sc, err := serviceContextFromConfig(cfg, "")
+func (a *App) corpusServiceManaged(cfg config.Config, global globalOptions) bool {
+	sc, err := serviceContextFromConfig(cfg, "", resolveConfigPath(global))
 	if err != nil {
 		return false
 	}
--- a/internal/cli/service.go
+++ b/internal/cli/service.go
@@ -118,29 +118,91 @@ func (a *App) resolveServiceContext(global globalOptions, nameOverride string) (
 	if err != nil {
 		return serviceContext{}, config.Config{}, fmt.Errorf("load config: %w", err)
 	}
-	sc, err := serviceContextFromConfig(cfg, nameOverride)
+	sc, err := serviceContextFromConfig(cfg, nameOverride, resolveConfigPath(global))
 	if err != nil {
 		return serviceContext{}, config.Config{}, err
 	}
 	return sc, cfg, nil
 }
 
+// absOrSelf returns the absolute form of p, falling back to p itself when the
+// working directory cannot be read (the historical behavior of every
+// filepath.Abs call on this path).
+func absOrSelf(p string) string {
+	abs, err := filepath.Abs(p)
+	if err != nil {
+		return p
+	}
+	return abs
+}
+
+// isExistingDir reports whether p names an existing directory.
+func isExistingDir(p string) bool {
+	info, err := os.Stat(p)
+	return err == nil && info.IsDir()
+}
+
+// serviceWorkingDirBase picks the directory the supervised daemon is booted in.
+//
+// For a filesystem-backed corpus (local/nfs) this stays the absolute corpus
+// root, unchanged: that is where the operator's .dir2mcp.yaml and .env.local
+// live by convention, and both the config loader (relative ".dir2mcp.yaml") and
+// the dotenv loader (relative ".env.local") resolve against the process working
+// directory.
+//
+// For an object-store corpus there is no local root to boot in (SPEC §7.8), and
+// rendering the ignored RootDir as launchd's WorkingDirectory / systemd's
+// WorkingDirectory installs a unit that cannot start (#738). The config file's
+// own directory is used instead: it is the one local directory that is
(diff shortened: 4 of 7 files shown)


yidakra:
### Note on the first `race` run (flake, not this PR)

The first CI attempt failed `race` on `TestReindex_RestoresContentHashesOnFailure` (`tests/cli/reindex_rollback_test.go:119`, "want `HASH-BEFORE-REINDEX` got `\"\"`"). The re-run is green. Evidence that it is unrelated to this change:

- `main` is green on every recent run.
- `internal/cli/reindex.go` references none of the symbols this PR touches (`validateUpConfig`, `prepareUpDirectories`, `serviceContextFromConfig`, `sourceIsRemote`, `ensureRootAccessible`, `absOrSelf`, `isExistingDir`) — verified by grep. `reindex` never enters `validateUpConfig`.
- Go registers tests in file order, so `reindex_rollback_test.go` runs *before* the new `s3_no_local_root_738_test.go`; the new tests cannot have influenced it.
- Locally: `go test -race ./internal/cli ./tests/cli` passes in full, and `-race -count=3` on that specific test passes.

Worth filing separately if it recurs. It failed in **0.36s**, so it is not a timeout: `BackupContentHashes` ran, `ClearDocumentContentHashes` ran, and the restore left the column empty. `reindexStaging.restoreContentHashes` is deliberately best-effort — it only `writef`s a warning on error — so a transient `SQLITE_BUSY` on the rollback `UPDATE` would produce exactly this symptom silently, and the test does not assert on that warning. Adjacent to #727.

yidakra:
@coderabbitai review

Issue #735, opened 2026-08-03
security: S3 discovery accepts traversal and absolute rel_path values


yidakra:
## Summary

`S3FS.Walk` accepts object keys whose prefix-relative name is absolute or contains `.` / `..` traversal segments. That violates the canonical remote-corpus boundary and lets untrusted bucket keys become unsafe `rel_path` values consumed by code that assumes a normalized root-relative path.

## Evidence

`relForKey` only checks the raw string prefix and strips it:

```go
if !strings.HasPrefix(key, s.prefix) { return "", false }
return strings.TrimPrefix(key, s.prefix), true
```

`discoveredFromObject` then rejects only empty/directory placeholders, excluded directory names, and oversized objects. It does not reject or normalize absolute paths, `.`, or `..`.

Examples accepted under prefix `corpus/` include:

- `corpus/../outside.mp4` -> `rel_path="../outside.mp4"`
- `corpus/a/../../outside.txt` -> `rel_path="a/../../outside.txt"`
- `corpus//absolute.mp4` -> `rel_path="/absolute.mp4"`

The canonical spec §7.8 says traversal/root-escape protections apply to every scheme and such object keys must be rejected with `PATH_OUTSIDE_ROOT`.

This is not only malformed metadata:

- video recognition joins `cfg.RootDir` and `doc.RelPath`, so a traversal key can make the recognition backend read or upload a local file outside the corpus root;
- TTML/SMIL export similarly probes `filepath.Join(cfg.RootDir, opts.relPath)`;
- local path filters clean these values differently from S3 `Open`, producing inconsistent listing/retrieval behavior;
- leading-slash rel paths do not round-trip through `keyForRel`, which strips the slash and addresses a different object key.

The existing S3 walk test covers prefix mapping but has no adversarial key cases.

## Reproduction

With prefix `corpus/`, return `corpus/../outside.mp4` from `ListObjectsV2`. `S3FS.Walk` returns a discovered file whose `RelPath` is `../outside.mp4` rather than rejecting it.

## Expected behavior

Define and enforce one backend-independent rel-path validator at the S3 discovery boundary:

- reject absolute, empty, `.`, `..`, and traversal-segment paths;
- preserve valid S3 key bytes without silently retargeting a different object;
- surface the canonical `PATH_OUTSIDE_ROOT` error/diagnostic rather than ingesting or silently skipping unsafe keys;
- add tests for traversal, leading slash, repeated separators, platform separator edge cases, and ordinary valid keys.

PR #760, opened 2026-08-05
fix(identity): derive an S3 instance's name from bucket and prefix (#737)

### LINKED ISSUE #737: identity: S3 corpora in the same working directory share the same default server and service name

yidakra:
Closes #737. Completes the S3/CorpusFS cluster (#734, #735, #736, #738).

`resolveServerName` always hashed the absolute local `RootDir`. For `source.kind=s3` that is not the corpus identity — `S3FS.Walk` ignores its root argument, bucket plus prefix **is** the corpus root (SPEC §7.8) — and an S3 deployment commonly leaves `root_dir` at its default.

Driven through the real CLI with the issue's own reproduction (one directory, two buckets, no explicit `server_name`), both configs resolved to the same name:

```
both S3 corpora resolved to "dir2mcp-dev-cli-b05a0c"; the second `claude mcp add` overwrites the first
```

That name feeds the MCP `serverInfo.name`, the launchd/systemd service label and the `claude mcp add` alias, so the second install collides with the first and diagnostics cannot say which remote corpus a connection serves.

## The key, and what is deliberately in and out of it

`identity.AutoServerNameForS3` derives from a canonical `s3://[endpoint/]bucket/prefix`:

**In:**
- bucket and normalized prefix — what actually identifies the corpus
- the endpoint host **only when configured**. Two S3-compatible stores can host the same bucket name, while AWS bucket names are globally unique; including it unconditionally would rename every existing AWS deployment

**Out:**
- **every credential** — this string reaches diagnostics, `claude mcp list` and a systemd unit file
- **the region** — AWS bucket names are global so it adds no uniqueness, and setting it later on a live deployment would silently rename the instance

## Normalization is slash-based, not `filepath.Clean`

An object key is slash-separated on every platform, and `Clean` rewrites `\` on Windows. So `corpus`, `corpus/`, `/corpus/` and `corpus//` are one corpus rather than four instances — an operator who adds a slash must not find their service label orphaned and their client alias pointing at nothing.

## Nothing local changes

Local and NFS derivation is byte-identical: the same `filepath.Clean(rootAbs)` hash through the same helper, so no existing install is renamed. An explicit `server_name` is resolved **before** the source branch, so the override cannot be bypassed by the new path.

## Tests

Twelve. The end-to-end pair drives `print-config claude` against two real S3 configs in one directory and was **verified to fail with the identity package present but the wiring stashed**, reproducing the collision quoted above. The rest pin prefix/bucket/endpoint normalization, that no credential reaches the name, that the name stays terminal-friendly and names its bucket, and that local derivation and the override are unchanged.

`make check` passes.

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -337,6 +337,23 @@ func applyScalarOverrides(cfg *config.Config, opts upOptions) {
 // distinct entries in `claude mcp list` instead of colliding on the
 // same auto-derived name.
 func resolveServerName(cfg *config.Config) {
+	if t := strings.TrimSpace(cfg.ServerName); t != "" {
+		// An explicit name stays authoritative on every backend, and is
+		// resolved before the source branch so the two paths cannot disagree
+		// about what an override means.
+		cfg.ServerName = t
+		return
+	}
+	// A remote corpus is not identified by its local root. `S3FS.Walk` ignores
+	// that root, and an S3 deployment commonly leaves `root_dir` at its
+	// default, so deriving from it gave two different buckets launched from one
+	// directory the same server name, service label and client alias (#737).
+	if sourceIsRemote(*cfg) {
+		cfg.ServerName = identity.AutoServerNameForS3(
+			cfg.Source.S3Bucket, cfg.Source.S3Prefix, cfg.Source.S3Endpoint, buildinfo.IsDev(),
+		)
+		return
+	}
 	abs, err := filepath.Abs(cfg.RootDir)
 	if err != nil {
 		abs = cfg.RootDir
--- a/internal/identity/name.go
+++ b/internal/identity/name.go
@@ -61,13 +61,97 @@ func Resolve(rootAbs, override string, dev bool) string {
 // the caller's normalization.
 func AutoServerName(rootAbs string, dev bool) string {
 	clean := filepath.Clean(rootAbs)
-	slug := slugify(filepath.Base(clean))
-	sum := sha256.Sum256([]byte(clean))
+	return nameFromKey(clean, filepath.Base(clean), dev)
+}
+
+// AutoServerNameForS3 derives the name for a corpus that lives in an object
+// store, where the local root is not the corpus identity.
+//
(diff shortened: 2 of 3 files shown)


yidakra:
@coderabbitai review

Issue #545, opened 2026-07-06
Apply subtitle drop_phrases/scrub_phrases at ingest (before embedding), not export-only


yidakra:
## Problem

`media.subtitles.drop_phrases` (#539) and `scrub_phrases` (#541) remove Whisper keyword-spam hallucination (e.g. `"Донбасс, Крым, Украина, Иван Плющ, НАТО"` — measured at ~35% of cues on some RFE/RL originals) **only at export time**: `DropSet.IsSpam` / `DropSet.Scrub` live in `internal/subtitle/clean.go` and are invoked solely from `internal/cli/export.go` (`renderTranscriptExport` → `CleanCues`).

That cleans the exported SRT/VTT sidecar, but the **same spam cues are still chunked and embedded into the retrieval index**. dir2mcp's primary surface is `search` / `ask` / citations over those embeddings — so the hallucinated keyword-spam continues to pollute retrieval quality, inflate the index, and surface as junk citations, even with drop/scrub configured.

This is inconsistent with `media.filter_words`, which is applied at **ingest, before embedding** — `internal/ingest/represent.go` `applyWordFilterToSegments` (via `chunkTranscriptByTimeWithWordsFiltered`, called from `service.go:2985` and `sidecar.go:336`) — so filter_words cleans both the index and the export. Drop/scrub should reach the same layer.

## Proposal

Apply the `drop_phrases` / `scrub_phrases` cleaning to transcript **chunk segments at ingest, before embedding**, mirroring `applyWordFilterToSegments`:

- Run `DropSet.IsSpam` (drop a wholly-spam segment) and `DropSet.Scrub` (excise a leaked phrase, keep the real speech) over the chunk segments in the ingest path, so spam is never embedded and a scrubbed segment is embedded clean.
- Prefer a **single shared cleaning entry point** used by both ingest and export, so the two paths can't diverge (today export owns the only call site).
- Preserve current export behavior (a segment dropped at ingest is simply absent from both the index and the export; scrubbed text flows to both).
- Keep it **off by default** (empty phrase lists = no-op), unchanged config keys.

## Considerations

- **Consistency:** ingest and export must apply the identical DropSet so the index and the sidecar agree cue-for-cue.
- **Whole-doc vs surgical:** this is complementary to `internal/quality` (which quarantines a whole representation on repetition-loop/gibberish); keyword-spam is real Cyrillic words fragmented across cues, which quality can't catch and shouldn't nuke the 65% genuine speech for.
[...]

PR #762, opened 2026-08-05
fix(export): route TTML through the shared cue pipeline and resolve its language (#729, #730)

### LINKED ISSUE #729: export: TTML bypasses configured subtitle glossary and cleaning pipeline
## Summary

The VTT/SRT export path applies the configured subtitle editorial pipeline, but the TTML path only applies `media.filter_words`. As a result, the same transcript exported as TTML can contain un-replaced terms, dropped/spam phrases, scrub targets, URL-only cues, and repeated cues that are removed from VTT/SRT.

This is surprising for settings under `media.subtitles.*`, and conflicts with the canonical description of `media.subtitles.glossary` as a deterministic “export-time” replacement rather than a VTT/SRT-only feature.

## Evidence

`internal/cli/export.go` builds:

- `subtitle.NewGlossary`
- drop/scrub sets
- `DropURLs`
- `CollapseRepeats`

and passes them through `subtitle.CleanCues` for VTT/SRT.

`internal/cli/export_ttml.go` instead:

```go
cues := subtitle.FilterCues(subtitle.BuildCues(chunks), filter)
```

It never constructs or applies the glossary/clean options before monolingual or bilingual alignment/rendering.

Config comments currently mention VTT/SRT for several fields, while the canonical spec uses format-neutral export wording, so the shipped behavior and contract also need reconciliation.

## Reproduction

1. Enable TTML and configure any of:
   - `media.subtitles.glossary`
   - `drop_phrases`
   - `scrub_phrases`
   - `collapse_repeats`
   - `drop_urls`
2. Seed/export the same transcript as SRT and TTML.
3. SRT reflects the configured cleanup; TTML preserves the original cues (except `media.filter_words`).

## Impact

- Broadcast TTML can reintroduce terms or spam an operator explicitly configured away.
- Bilingual cue alignment happens on a different cue set than the cleaned exports.
- Multiple advertised subtitle formats are not equivalent derivations of the selected transcript/config.

## Expected behavior

Define one shared cue-preparation pipeline and apply it consistently before rendering every subtitle format, including each language before bilingual alignment.

If some settings are intentionally format-specific, rename/scope them explicitly in config and canonical spec rather than silently ignoring them.

- Parameterized VTT/SRT/TTML tests for every glossary/clean option.
- Bilingual TTML cleans both languages before alignment.
- Empty config remains byte/behavior compatible.
- Invalid regex/glossary errors remain config-time failures.

### LINKED ISSUE #730: export: TTML emits empty language metadata when --lang is omitted

yidakra:
Fixes #729. Fixes #730.

Both issues live in the TTML branch of `export`, one line apart in `resolveCues`, so they are fixed together rather than as two conflicting PRs.

Rebased on current `main` (`57e1bc2`). Everything #736 landed is preserved: SMIL probing still goes through `CorpusFS.Localize`, the `warnSMILSkipped` reasons are still fixed labels with no error interpolation, and `MediaSrc` is still the full `rel_path`.

---

## #729 — TTML bypassed the cleaning/glossary pipeline

**Verified before changing anything.** `runTTMLExport` is dispatched at `export.go:78`, which returns *before* `export.go` ever builds `CleanOptions`. So `subtitle.CleanCues` had exactly one production call site in the repo, on the VTT/SRT path. TTML did `subtitle.FilterCues(subtitle.BuildCues(chunks), filter)` and nothing else.

### Where cleaning is applied on the ingest side, and why export did not go through it

This was the key question, and the answer changes where the fix belongs. Measured, not assumed:

| config key | applied at ingest (indexed text) | export vtt/srt | export ttml (before) |
|---|---|---|---|
| `media.filter_words` | yes | yes | yes |
| `media.subtitles.drop_phrases` | yes (#545) | yes | **no** |
| `media.subtitles.scrub_phrases` | yes (#545) | yes | **no** |
| `media.subtitles.glossary` | **no** | yes | **no** |
| `media.subtitles.collapse_repeats` | **no** | yes | **no** |
| `media.subtitles.drop_urls` | **no** | yes | **no** |

Ingest applies `filter_words` plus drop/scrub in all three chunking paths (`ingest/service.go:4462` STT, `:4850` translate, `ingest/sidecar.go:370` sidecar), via `applyWordFilterToSegments` and `applyDropScrubToSegments`. That cleaned text is what lands in `chunks.text` and is what `TranscriptSpanChunks` reads back — so yes, export re-reads already-partly-cleaned cues from the store.

But **glossary, collapse_repeats and drop_urls are applied nowhere on the ingest side** (grep-verified: their only readers outside `internal/config` were `export.go`). The stored text is *not* glossary-rewritten. So the export path genuinely has to apply them, and moving the fix to ingest would not work: a glossary is an editorial rewrite of rendered cue text, and the spec calls it exactly that — "a deterministic, export-time find/replace on **already-rendered cue text**" (SPEC §8.6.2). The fix belongs in export, and specifically in the shared step both formats run.

Re-running drop/scrub at export on top of ingest's pass is idempotent and was already the pre-existing VTT/SRT behavior; this PR does not change that.

### Which surface this fixes, and which it does not

To be explicit, given the #545 history:

- **Fixed here:** the *export* surface. `--format ttml` (and its companion SMIL) now applies the same five `media.subtitles.*` rules that `--format vtt|srt` already applied.
- **Not touched here:** the *retrieval index*. `glossary`, `collapse_repeats` and `drop_urls` are still export-only — a hallucinated URL cue is still embedded, cited and searchable, and a glossary rewrite still does not reach the indexed text. That is the residual half of the gap #545 named ("prefer a single shared cleaning entry point used by both ingest and export"); #545 closed having moved only drop/scrub to ingest. Changing what gets embedded is an ingest-side, re-index-forcing behavior change and is deliberately out of scope for a TTML export fix. The config doc comments in this PR now state, per key, which of the two it is, so the gap is documented rather than silent.

### The fix

A single `cuePipeline` (`export.go`) holding the word filter plus `CleanOptions`, with one `apply` used by both renderers. Formats may differ in how cues are **built** (segmentation, bilingual alignment); they must not differ in how cues are **cleaned**.

For bilingual TTML, each language is cleaned **before** `AlignBilingual`, not after. Aligning pre-clean cues would time-region-merge pairs that cleaning then drops, which makes alignment depend on config in a way §8.6.10's determinism and same-segment-span guarantees do not allow.

`media.subtitles.segmentation` is deliberately **not** in the shared pipeline. It rewrites cue *boundaries* for VTT/SRT reading speed, and TTML cue regions are the alignment unit for bilingual packaging and must stay the stored segment spans. Per the issue's "if some settings are intentionally format-specific, scope them explicitly", its config comment now says so and why.

### Contract reconciliation

The issue notes the config comments and the canonical spec disagreed. Checked the submodule: SPEC.md never mentions `drop_phrases`, `scrub_phrases`, `collapse_repeats`, `drop_urls` or `filter_words` at all, and its only wording for `media.subtitles.glossary` is format-neutral export-time language. **The spec was already right; the code and its comments were the outliers.** No spec change is needed and the submodule pin is untouched — this PR moves the code toward the existing spec text rather than changing the contract.

Four comments were factually wrong and are corrected: `drop_phrases` and `scrub_phrases` claimed "Only affects VTT/SRT export" while having run at ingest since #545, and `media.filter_words` claimed export coverage of "VTT/SRT" when TTML already filtered too.

---

## #730 — empty language metadata when `--lang` is omitted

**Verified before changing anything.** Reproduced exactly as reported: a transcript whose `meta_json` is `{"language":"en"}`, exported with `--lang` omitted, emitted `<tt ... xml:lang="">`, bare `<span>` runs, and a SMIL with no `systemLanguage`.

### What the correct value is, and why

**The selected representation's own recorded language.** Not a configured default, and not a refusal.

`--lang` is a *selector*: `selectTranscriptRep(reps, "")` already picks the first/source transcript correctly. The bug was purely that the chosen representation was thrown away and the raw flag echoed onward. The representation's `meta_json` language is the authoritative record of what language those cues are actually in, so it is the tag to emit. Refusing the export would break the documented optional-`--lang` path; a configured default would be a value the operator set for something else.

This also canonicalizes spelling: `--lang` matching is case-insensitive, so `--lang EN` against a transcript recording `en` now emits `en` rather than the caller's `EN`.

**For a genuinely untagged legacy transcript the tag stays empty, by design.** No fallback is invented. `xml:lang=""` means "no language information is available" per XML 1.0 §2.12, and TTML1's own examples use `<tt xml:lang="">`; text runs omit `xml:lang` entirely and SMIL omits `systemLanguage`. Substituting a guess (or `und`) would put a plausible-but-wrong BCP-47 tag into a broadcast subtitle file, and players act on that tag for track selection and font/shaping. A wrong tag misroutes; an absent tag merely degrades. This case is only reachable with `--lang` omitted anyway — a non-empty `--lang` can only match a representation that recorded a language, so there is never a caller tag to echo.

The resolved tag is threaded through all four consumers the issue lists: `MonolingualBilingualCues`, `AlignBilingual` (both tags), `RenderTTML`, and the SMIL sidecar.

---

## Testing

New tests under `tests/cli/`, **proven to fail first**: source change stashed (`git stash push -- internal/`, tests kept), suite re-run, then restored. 7 of 9 fail without the fix:

```
--- FAIL: TestExportTTMLAppliesCueCleaning729
--- FAIL: TestExportTTMLBilingualCleansBothLanguages729
--- FAIL: TestExportTTMLOmittedLangUsesRecordedLanguage730
--- FAIL: TestExportTTMLOmittedLangMatchesExplicitLang730
--- FAIL: TestExportTTMLBilingualOmittedLangResolvesBothTags730
--- FAIL: TestExportTTMLSecondaryLangTagComesFromRepresentation730
--- FAIL: TestExportTTMLSMILUsesResolvedLanguage730
```

`export_ttml_cleaning_729_test.go` reuses the *exact* fixture the existing SRT cleaning test uses (`seedCleanExportStore`) with the same config, so any future divergence between the two formats fails here.

The two that pass both before and after are deliberate contract guards, not regressions:

- `TestExportTTMLNoCleaningConfigUnchanged729` — empty config leaves TTML byte-compatible, guarding the fix from becoming an unconditional rewrite (the issue's "empty config remains byte/behavior compatible").
- `TestExportTTMLUntaggedTranscriptEmitsNoLanguage730` — pins the no-fallback decision above, asserting no `en`/`und`/`ru` is ever invented.

Invalid glossary/regex still fails at config time on the TTML path too; the pipeline is built *after* the enabled gate, so a disabled surface still reports "TTML export is disabled" rather than an unrelated config error (`TestExportTTMLDisabledByDefault` still passes).

No new file under `internal/cli/`, so `tests/security/cli_boundary_test.go` needs no allowlist entry — checked, and that suite passes.

`make check` passes locally, including `gocyclo -over 15`, `go vet` and `golangci-lint`.

## Not done

- Moving `glossary`/`collapse_repeats`/`drop_urls` to ingest so the retrieval index matches the exports. Real, and named above; it forces a re-index and is a separate change.

### DIFF
--- a/internal/cli/export.go
+++ b/internal/cli/export.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/dirstral/dir2mcp/internal/config"
 	"github.com/dirstral/dir2mcp/internal/model"
 	"github.com/dirstral/dir2mcp/internal/store"
 	"github.com/dirstral/dir2mcp/internal/subtitle"
@@ -78,36 +79,12 @@ func (a *App) runExport(ctx context.Context, global globalOptions, args []string
 		return a.runTTMLExport(ctx, global, cfg, ts, opts)
 	}
 
-	filter := subtitle.NewWordFilter(cfg.MediaFilterWords)
-	// The glossary was already validated at config load, so a parse error here is
-	// unexpected; surface it rather than silently dropping the glossary.
-	glossary, err := subtitle.NewGlossary(cfg.MediaSubtitlesGlossary)
+	pipe, err := newCuePipeline(cfg)
 	if err != nil {
-		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid, fmt.Sprintf("invalid media.subtitles.glossary: %v", err))
+		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid, err.Error())
 		return exitConfigInvalid
 	}
-	// The drop set was already validated at config load, so a parse error here is
-	// unexpected; surface it rather than silently dropping the rules.
-	drop, err := subtitle.NewDropSet(cfg.MediaSubtitlesDropPhrases)
-	if err != nil {
-		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid, fmt.Sprintf("invalid media.subtitles.drop_phrases: %v", err))
-		return exitConfigInvalid
-	}
-	// The scrub set was already validated at config load, so a parse error here is
-	// unexpected; surface it rather than silently dropping the rules.
-	scrub, err := subtitle.NewDropSet(cfg.MediaSubtitlesScrubPhrases)
-	if err != nil {
-		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid, fmt.Sprintf("invalid media.subtitles.scrub_phrases: %v", err))
-		return exitConfigInvalid
-	}
-	clean := subtitle.CleanOptions{
-		DropURLs:        cfg.MediaSubtitlesDropURLs,
-		Drop:            drop,
-		Scrub:           scrub,
-		CollapseRepeats: cfg.MediaSubtitlesCollapseRepeats,
-		Glossary:        glossary,
-	}
-	rendered, code := a.renderTranscriptExport(ctx, global, ts, opts, filter, clean, cfg.MediaSubtitlesSegmentation)
+	rendered, code := a.renderTranscriptExport(ctx, global, ts, opts, pipe, cfg.MediaSubtitlesSegmentation)
 	if code != exitSuccess {
 		return code
 	}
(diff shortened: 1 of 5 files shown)


yidakra:
@coderabbitai review

PR #763, opened 2026-08-05
fix(mcp): resolve on-demand media through the CorpusFS (#759)

### LINKED ISSUE #759: mcp: on-demand media tools bypass CorpusFS and are unusable on S3 corpora
Found while landing the CorpusFS cluster (#734, #736, #738). Same root cause, different surface: MCP tool handlers reconstruct a local path from `RootDir + rel_path` instead of resolving through the configured `CorpusFS`.

## Evidence

`internal/mcp/tools.go::resolveDocumentPath` (line 2655):

```go
rootAbs, err := filepath.Abs(strings.TrimSpace(s.cfg.RootDir))
...
absPath := filepath.Join(rootAbs, filepath.FromSlash(normalized))
targetReal, err := filepath.EvalSymlinks(absPath)
```

`S3FS.Walk` ignores its `root` argument and emits `DiscoveredFile{AbsPath: ""}` (the code says so in a comment), so for `source.kind=s3` there is no file at `RootDir/rel_path`. `EvalSymlinks` on a path that does not exist returns `ENOENT`, so the resolve fails before any tool logic runs.

## Blast radius is wider than the CorpusFS cluster suggested

Three distinct entry points, not just the on-demand transcribe path:

| caller | tool surface |
|---|---|
| `handleOpenMediaClipTool` (line 1621) | `open_media_clip` |
| `initAudioDocumentOnDemand` (line 2378) | on-demand audio init for transcription |
| `readDocumentContent` (line 2647), called from three sites | on-demand content reads |

So an S3 corpus that indexed successfully still cannot serve `open_media_clip` or any on-demand media path.

## Why it is not simply "call Localize"

`resolveDocumentPath` is doing three jobs at once, and only the first is path construction:

1. building the local path,
2. enforcing root containment (`ErrPathOutsideRoot`) via `filepath.EvalSymlinks` on both the root and the target,
3. applying the corpus path-exclusion policy (`ingest.MatchesAnyPathExclude`, #407), which exists precisely because these on-demand branches would otherwise be the only way to reach a file the operator deliberately excluded.

Points 2 and 3 must survive any fix. The containment check in particular is currently expressed in terms of a real local filesystem (`EvalSymlinks`), which has no meaning for an object store; `internal/relpath` (added in #735) is the backend-independent form of that check and is the natural replacement, with `CorpusFS.Localize` supplying the bytes.

Note also that a `Localize` for S3 downloads the object, so a fix must keep the exclusion check BEFORE the fetch, or an excluded file gets downloaded before being refused.

## Expected behaviour

- On-demand media tools resolve content through the configured `CorpusFS`, as ingest's probe paths and (after #734/#736) recognition and SMIL export do.
- Root containment and path exclusion are enforced backend-independently and BEFORE any fetch.
- Temporary local copies are cleaned up on every path, including error paths.
- Local corpora keep their current behaviour exactly, including symlink resolution and the containment guarantee.

- Each of the three entry points against an S3-backed corpus (there is a network-free S3 fake at `tests/corpusfs/s3_test.go::newFakeS3FS`, and #755/#756 added ingest- and CLI-level harnesses worth copying).
- An excluded path is refused on S3 **without** the object being fetched.
- A traversal/absolute `rel_path` is still refused on both backends.
- Local behaviour unchanged, including that no copy is made.

yidakra:
Closes #759.

## What was broken

`internal/mcp/tools.go::resolveDocumentPath` built `filepath.Join(RootDir, rel_path)` and then `filepath.EvalSymlinks`'d it. `S3FS.Walk` ignores its `root` argument and emits `DiscoveredFile{AbsPath: ""}`, so for `source.kind=s3` there is no file at that path and the resolve failed with ENOENT before any tool logic ran. All three callers were affected: `handleOpenMediaClipTool`, `initAudioDocumentOnDemand`, `readDocumentContent` (three call sites).

I reproduced it rather than taking the issue's word for it: with the fix's dispatch forced back to the old branch, the new tests report `FILE_NOT_FOUND` (the ENOENT, mapped) on the clip path, the annotate read path and the on-demand init path. See "Proving the tests fail first" below.

## What changed

- `mcp.WithCorpusFS` injects the corpus filesystem at construction; `up.go` builds it once (moved a few lines earlier) and passes it to both the ingestor/retriever wiring and the server.
- `Server.localizeDocument(ctx, relPath) (path, cleanup, error)` is the single resolution seam for all three entry points. The returned cleanup is **never nil**, so `defer cleanup()` on the line after the call is always correct, and it is called on every path: success, `MEDIA_CLIP_FAILED`, `CLIP_TOO_LARGE`, `FILE_TOO_LARGE`, store errors, and the failed-`Localize` path itself (where a non-nil cleanup is invoked defensively even though the contract says nothing was materialized).
- `readDocumentContent` takes a `ctx` (all three call sites already had one).

### The two guarantees

**Containment.** For a non-local corpus it comes from `relpath.Normalize` (#735), the same rule S3 discovery applies to every key it emits, because `EvalSymlinks` means nothing for an object store. A local corpus keeps `resolveDocumentPath` verbatim, including the symlink-target re-check.

**Exclusion (#407).** `checkRemoteDocumentPolicy` runs `ingest.MatchesAnyPathExclude` **before** `Localize`, so an excluded object is never fetched. This is load-bearing, not decorative: with the exclusion check deleted, `TestOpenMediaClip_S3Corpus_ExcludedPathRefusedWithoutFetch` goes green-path and returns the clip bytes of the excluded file. The test asserts `GetObject` was called **zero** times, not merely that the call was refused.

### Why the local branch is not routed through `LocalFS.Localize`

`LocalFS.Localize` would give the right path (resolved in-root, no copy), but `resolveDocumentPath` does one thing it cannot: after `EvalSymlinks`, it re-applies the exclusion policy to the **symlink-resolved** rel path, so `link.wav -> private/voice.wav` is refused when `private/**` is excluded. That check has no object-store analogue, so rather than reimplement it generically I left the local path untouched and gated the CorpusFS branch on `corpusFS != nil && source.kind == s3`. Gating on the source kind as well as on injection means an accidentally injected local backend cannot silently downgrade that guarantee. `TestOnDemand_LocalCorpus_RefusesSymlinkToExcludedTarget` pins it, and the local tests inject a CorpusFS whose every method fails the test if called.

## Where the issue's framing needed adjusting

Two places, both about what is reachable rather than about the diagnosis:

1. **`..` and absolute rel_paths never reach the on-demand branch on either backend.** The store's own `normalizeRelPath` rejects them during the document lookup that precedes it, so the observable code is `STORE_CORRUPT`, not `PATH_OUTSIDE_ROOT`. I did not change that (it is a correct refusal with a poor code, and it is outside this issue). The traversal tests therefore assert "refused, and nothing fetched, and no document row created" rather than pinning a code that the on-demand branch does not get to produce. The case that *does* reach the new containment check is a doubled separator (`media//voice.wav`, which the store cleans for the lookup but which names a different S3 object), and that one is pinned to `PATH_OUTSIDE_ROOT`.

2. **The remote branch is deliberately stricter than the local one.** The local branch cleans (`filepath.Clean`), so `media//voice.wav` and `./a.mp3` resolve. `relpath.Normalize` refuses them instead of rewriting them, because an S3 key is an opaque byte string: cleaning would address a different object than the caller named, and would also let a path dodge an exclusion glob that the uncleaned form matches. So the same rel_path can be accepted on a local corpus and refused on an S3 one. That asymmetry is intentional and it is what #735 established for discovery; the alternative (clean, then fetch) is the unsafe one.

## Known, unchanged, out of scope

- A `Localize` failure that is not a containment refusal (a missing object, a network error) maps to the existing default, `PERMISSION_DENIED`. `corpusfs` does not classify a NoSuchKey as `fs.ErrNotExist`, and teaching it to do so touches the embedding worker's error classification, so I left it alone rather than widen the blast radius. Worth a follow-up.
- `initAudioDocumentOnDemand` stats the localized copy, so for S3 the on-demand row's `MTimeUnix` is the download time, not the object's `LastModified`. Size and content hash are the object's. Discovery overwrites the row on the next scan. Commented in place.
- Transcribe reads the object twice (on-demand init, then `ensureTranscriptForAudioDoc`), so a first-time transcribe on S3 costs two GETs. Pre-existing shape; no correctness impact.

No spec change is needed: this restores behaviour SPEC §7.8 already requires on every backend.

## Tests

`tests/mcp/media_corpusfs_759_test.go`, driving a **real** `corpusfs.S3FS` over a network-free stub client (the shape of `tests/corpusfs/s3_test.go::newFakeS3FS`, which is in another package), so the S3 key mapping, `guardRel`, the temp download and the cleanup are the production ones. `RootDir` is a real but empty directory, so a failure can only come from the missing `RootDir/rel_path` and not from a missing root.

- all three entry points against an S3 corpus: the clip is cut from a copy holding the object's bytes (and *not* from the `RootDir` path); the on-demand row is created with the object's size and content hash; annotate's read is proven by the secret-content gate firing (`FORBIDDEN` only happens on bytes that were actually read, and it needs no provider)
- excluded path refused on S3 with `GetObject` called zero times, on both the transcribe and the clip path, and no document row created
- traversal/absolute/doubled-separator refused on S3 with nothing fetched; traversal refused locally with the out-of-root file never reached
- local unchanged: extraction gets the symlink-resolved in-root path holding the source bytes, no copy, and an injected CorpusFS is never consulted; a symlink to an excluded target is still `FORBIDDEN`
- cleanup: the cache dir is asserted empty after every request, including the failed-extraction one, and the localized path is asserted gone

### Proving the tests fail first

I did not stash the whole source change (the tests would not compile without `WithCorpusFS`). Instead I forced `corpusFSForOnDemand` to return nil, which is exactly the pre-fix dispatch, and ran the suite:

```
--- FAIL: TestOpenMediaClip_S3Corpus_CutsFromLocalizedCopy
    expected success, got error: {"code":"FILE_NOT_FOUND", ...}
--- FAIL: TestOpenMediaClip_S3Corpus_ReleasesCopyOnExtractFailure
    got="FILE_NOT_FOUND" want="MEDIA_CLIP_FAILED"
--- FAIL: TestTranscribeOnDemand_S3Corpus_InitializesDocumentFromObject
    on-demand init did not create the document row: file does not exist
--- FAIL: TestAnnotate_S3Corpus_ReadsDocumentContent
    got="FILE_NOT_FOUND" want="FORBIDDEN"
--- FAIL: TestOnDemand_S3Corpus_RefusesPathsOutsideRoot/doubled_separator
    error code = "FILE_NOT_FOUND", want "PATH_OUTSIDE_ROOT"
```

The exclusion and local-behaviour tests pass in both states by design: they are guard tests for what must *not* change, and their power is shown by the separate mutation above (deleting the exclusion check makes both exclusion tests fail, one of them by returning the excluded file's clip).

`make check` is green.

## Review

`coderabbit review --agent --committed` raised two findings; both are addressed in the follow-up commit:

- *"persist the configured corpus source type instead of hardcoding `SourceType: \"filesystem\"`"* — not valid as stated: `store.normalizeSourceType` collapses every value to `archive_member` or `filesystem`, and ingest does not set it either, so an S3-discovered document is already stored as `filesystem`. Writing `"s3"` would be normalized straight back. I did fix the mirror image of it in the test, which seeded `source_type="s3"` and implied a value that does not exist.
- *"preserve leading/trailing whitespace in `lookupDocumentForTool`"* — out of scope and moot: `store.normalizeRelPath` trims too, so a key with edge whitespace cannot round-trip through the corpus today regardless. I corrected my own comment, which overstated what the no-trim in `checkRemoteDocumentPolicy` buys.

### DIFF
--- a/internal/cli/up.go
+++ b/internal/cli/up.go
@@ -121,7 +121,19 @@ func (a *App) runUp(ctx context.Context, opts upOptions) int {
 		return !indexingState.Snapshot().Running
 	})
 
+	corpusFS, err := buildCorpusFS(ctx, cfg)
+	if err != nil {
+		writeCLIError(a.stderr, opts.jsonOutput, exitConfigInvalid, fmt.Sprintf("initialize corpus source: %v", err))
+		return exitConfigInvalid
+	}
+
 	serverOptions := buildMCPServerOptions(&cfg, st, indexingState, emitter)
+	// Route the on-demand tool paths (open_media_clip, on-demand audio init,
+	// on-demand content reads) through the corpus filesystem so they work on an
+	// object-store corpus, which has no file at RootDir/rel_path (#759). The
+	// server ignores it for a local/NFS corpus, which keeps its filesystem-native
+	// resolution.
+	serverOptions = append(serverOptions, mcp.WithCorpusFS(corpusFS))
 	mcpServer := mcp.NewServer(cfg, ret, serverOptions...)
 	ing, err := a.newIngestor(cfg, st)
 	if err != nil {
@@ -136,11 +148,6 @@ func (a *App) runUp(ctx context.Context, opts upOptions) int {
 	})
 	wireDerivationCacheIdentities(ret, ing)
 
-	corpusFS, err := buildCorpusFS(ctx, cfg)
-	if err != nil {
-		writeCLIError(a.stderr, opts.jsonOutput, exitConfigInvalid, fmt.Sprintf("initialize corpus source: %v", err))
-		return exitConfigInvalid
-	}
 	setCorpusFSIfSupported(ing, corpusFS)
 	// Route retrieval-time reads (open_file raw text / OCR / transcript) through
 	// the corpus filesystem for object-store backends so open_file works on an S3
--- a/internal/mcp/server.go
+++ b/internal/mcp/server.go
@@ -29,6 +29,7 @@ import (
 	"github.com/dirstral/dir2mcp/internal/avutil"
 	"github.com/dirstral/dir2mcp/internal/buildinfo"
 	"github.com/dirstral/dir2mcp/internal/config"
+	"github.com/dirstral/dir2mcp/internal/corpusfs"
 	"github.com/dirstral/dir2mcp/internal/identity"
 	"github.com/dirstral/dir2mcp/internal/model"
 	"github.com/dirstral/dir2mcp/internal/protocol"
@@ -146,6 +147,13 @@ type Server struct {
 	// to avutil.ExtractSegment and is an injectable seam so tests can stub
 	// extraction without ffmpeg on PATH.
 	extractSegment func(ctx context.Context, path string, startMS, endMS int) ([]byte, error)
+
+	// corpusFS supplies the bytes for the on-demand tool paths (open_media_clip,
+	// on-demand audio init, on-demand content reads) when the corpus lives on a
+	// non-local backend (#759). It is set once at construction and read-only
+	// thereafter. nil (the default) and a local corpus source both keep the
+	// historical local-filesystem resolution; see corpusFSForOnDemand.
+	corpusFS corpusfs.CorpusFS
 }
 
 type rpcRequest struct {
@@ -228,6 +236,21 @@ func WithTTS(tts TTSSynthesizer) ServerOption {
 func WithExtractSegment(fn func(ctx context.Context, path string, startMS, endMS int) ([]byte, error)) ServerOption {
 	return func(s *Server) {
 		s.extractSegment = fn
+	}
+}
+
+// WithCorpusFS injects the corpus filesystem backend the on-demand tool paths
+// read through (issue #759). It matters only for a non-local corpus source: an
+// object store has no file at RootDir/rel_path, so open_media_clip, on-demand
+// audio init and on-demand content reads have to obtain the bytes through the
+// backend instead of joining a path that names nothing.
+//
+// A local/NFS corpus ignores it and keeps the filesystem-native resolution
+// (EvalSymlinks containment plus the symlink-target exclusion re-check), which
+// has no object-store analogue and must not be weakened by an injection.
+func WithCorpusFS(fsys corpusfs.CorpusFS) ServerOption {
+	return func(s *Server) {
+		s.corpusFS = fsys
 	}
 }
 
--- a/internal/mcp/tools.go
+++ b/internal/mcp/tools.go
@@ -21,12 +21,14 @@ import (
 
 	"github.com/dirstral/dir2mcp/internal/avutil"
 	"github.com/dirstral/dir2mcp/internal/config"
+	"github.com/dirstral/dir2mcp/internal/corpusfs"
 	"github.com/dirstral/dir2mcp/internal/ingest"
 	"github.com/dirstral/dir2mcp/internal/mistral"
 	"github.com/dirstral/dir2mcp/internal/model"
 	"github.com/dirstral/dir2mcp/internal/protocol"
 	"github.com/dirstral/dir2mcp/internal/provider"
 	"github.com/dirstral/dir2mcp/internal/providerfactory"
+	"github.com/dirstral/dir2mcp/internal/relpath"
 )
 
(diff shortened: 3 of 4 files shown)


yidakra:
@coderabbitai review

PR #766, opened 2026-08-05
fix(mcp): report a secret-excluded document as skipped, not ok (#712)

### LINKED ISSUE #712: mcp: list_files reports secret-excluded documents as status ok
## Summary

`dir2mcp_list_files` reports every persisted `secret_excluded` document as `status: "ok"`, even though the file was deliberately withheld and has no searchable chunks. This undoes the operator-facing audit value restored by #425 and contradicts `CorpusStats`, which correctly counts the same row as skipped.

## Evidence

- Ingest persists `documents.status = secret_excluded` when content matches secret detection.
- `store.normalizeStatus` now preserves that state after #425.
- `CorpusStats` counts `secret_excluded` together with `skipped`.
- `handleListFilesTool` calls `normalizeFileStatus(doc.Status)`.
- `normalizeFileStatus` recognizes only `skipped` and `error`; every other value, including `secret_excluded`, falls through to `ok`.

The advertised/published list-files schema permits only `ok | skipped | error`, so the conforming honest projection is at minimum `secret_excluded -> skipped` unless the spec is intentionally extended with a detailed status/reason field.

## Reproduction

1. Ingest a file that matches a configured secret pattern.
2. Confirm the stored document has `status=secret_excluded`, zero searchable chunks, and increments the skipped count.
3. Call `dir2mcp_list_files`.
4. The same path is returned with `status: "ok"`.

## Impact

Agents and operators using the MCP surface are told a sensitive withheld file was successfully indexed. Coverage/accounting becomes internally contradictory: `stats` reports a skip while `list_files` reports the document as healthy.

## Expected behavior

Project `secret_excluded` to `skipped` under the current schema, or introduce a spec-first detailed status/skip reason that preserves the distinction. Never report a withheld/unindexed document as `ok`.

Add a tool-level regression test using a `secret_excluded` store row and assert the structured output conforms to the advertised schema while reporting a non-ok status.

yidakra:
Closes #712.

`normalizeFileStatus` recognised only `skipped` and `error`, so every other stored status fell through to `ok`. A document deliberately **withheld for containing secrets** was reported to callers as successfully indexed, despite having zero searchable chunks.

The surfaces contradicted each other about the same row:

- `CorpusStats` counts `status IN ('skipped','secret_excluded')` as skipped
- `skip_summary.go` pairs them the same way, twice
- `list_files` called it healthy

This function was the only place that disagreed. It also undid the operator-facing audit value #425 restored by persisting the state at all.

The published schema pins status to `ok|skipped|error` (`spec/tools/schemas/list_files.json`), so `skipped` is the conforming honest projection — **no spec change needed**.

## The `pending` question, answered rather than ignored

The store also normalizes a `pending` document status, which falls through the same default arm. I left it alone deliberately, and recorded why at the call site:

- nothing in production writes it — verified, every `pending` in the store is a **chunk's** `embedding_status`, not a document status
- the published enum has no state that honestly describes "not indexed yet", so projecting it to `skipped` would be a different lie

If a document ever does become pending, that wants a spec-first status extension, not a guess.

## Tests

Two, driven through the real MCP tool over HTTP rather than the internal helper — the defect was in the projection the *tool* applies, and a store-level test would have shown the correct `secret_excluded` all along.

The first was **verified to fail pre-fix** with the reported symptom:

```
a withheld document is reported as "ok"; an agent is told a secret-bearing file was indexed (#712)
```

The second pins the projection itself: whatever status the store grows next, this tool may only emit the three states it advertises.

**Note for the next author of a `list_files` test:** the listing resolves every row against the corpus root and drops what does not exist there, so a store-only fixture returns zero files and makes such a test vacuous. My first draft did exactly that and passed trivially. The helper here writes a real file per document.

`make check` passes.

### DIFF
--- a/internal/mcp/tools.go
+++ b/internal/mcp/tools.go
@@ -3144,9 +3144,33 @@ func parseOptionalStringSlice(args map[string]interface{}, key string) ([]string
 	}
 }
 
+// normalizeFileStatus projects a stored document status onto the three values
+// the published list_files schema allows (`ok|skipped|error`, SPEC §5.1 and
+// spec/tools/schemas/list_files.json).
+//
+// The projection has to be conservative, because the default arm is what a
+// caller is told about any status this function does not know. Reporting a
+// withheld file as `ok` tells an agent a sensitive document was indexed
+// successfully, which is the opposite of true and the opposite of what every
+// other surface says about the same row.
+//
+// `secret_excluded` is a skip, not a success (#712). A document withheld for
+// containing secrets has zero searchable chunks, and the rest of the codebase
+// already agrees: CorpusStats counts `status IN ('skipped','secret_excluded')`
+// as skipped, and skip_summary.go pairs them the same way twice. This function
+// was the only place that disagreed, so `stats` reported a skip while
+// `list_files` reported the same document healthy.
+//
+// The store also normalizes a `pending` document status, which would fall
+// through to `ok` here. It is left alone deliberately: nothing in production
+// writes that value (every `pending` in the store is a CHUNK's
+// embedding_status), and the published enum has no state that honestly
+// describes "not indexed yet" — `skipped` would be a different lie. If a
+// document ever does become pending, this needs a spec-first status extension
+// rather than a guess.
 func normalizeFileStatus(status string) string {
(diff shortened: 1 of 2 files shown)


yidakra:
@coderabbitai review

Issue #726, opened 2026-08-03
security: state directory and derived-content caches are created world-readable


yidakra:
## Summary

dir2mcp stores corpus-derived plaintext and retrieval payloads under `StateDir`, but several normal startup/ingest paths create the state/cache hierarchy as `0755` and files as `0644`. With the common `022` umask, other local users can read OCR output, transcripts, translations, summaries, contextual text, and vector-index payload metadata/snippets.

This contradicts the privacy posture used for `meta.sqlite`, tokens, and support bundles, which are explicitly owner-only.

## End-to-end evidence

A default support-bundle/config load against a fresh state path created:

```
state directory: 0755
.dir2mcp.yaml.snapshot: 0600
```

The directory mode is observable with:

```bash
umask 022
dir2mcp --state-dir /tmp/new-state support-bundle --output /tmp/bundle.tar.gz
stat -f '%Lp' /tmp/new-state   # 755 on macOS
```

The main startup and reindex paths likewise call `os.MkdirAll(cfg.StateDir, 0o755)`.

## Sensitive files written 0644

Examples in current code:

- HNSW snapshots use `os.Create` and a `0755` parent.
- Disk-index segments and identity sidecars use `0644`.
- `IndexPayload` persists `RelPath`, `Title`, `Snippet`, speaker data, spans, and media references.
- Docling/Pandoc/OCR caches persist extracted document text as `0644`.
- transcript, word-timing, translation, summary, and contextual-retrieval caches use `0644`.
- cache directories are repeatedly created as `0755`.

These are not harmless model blobs; many are direct or generated plaintext representations of corpus content.

## Impact

On a multi-user host, another local account can traverse the state tree and read corpus-derived content even if the original corpus directory itself is private.

A user-supplied existing permissive `StateDir` is also not tightened by `MkdirAll`.

## Expected behavior

Treat the whole local state hierarchy as confidential by default:

- create and enforce `StateDir` and content-bearing subdirectories as `0700`;
- create all corpus-derived state files as `0600`;
- tighten existing files/directories where ownership permits;
- use secure atomic-write helpers that preserve the final restrictive mode;
- fail clearly when a confidentiality guarantee cannot be established rather than silently continuing insecurely.

Non-secret externally consumed artifacts can be explicitly exempted if needed.

## Suggested regression coverage

Under a permissive umask:

- fresh startup/reindex/support commands leave state dirs `0700`;
- HNSW/disk indexes and every derived cache type are `0600`;
[...]

Issue #727, opened 2026-08-03
reindex: a new run destroys the last-known-good backups left by a crashed rebuild


yidakra:
## Summary

The reindex staging design preserves old index files and content hashes for rollback within the current process, but a crash/power loss leaves those backups on disk. The next `reindex` unconditionally deletes them before deciding whether the current live files/hashes are complete.

As a result, the first retry after a crash can destroy the only last-known-good index and replace its backup with the partial rebuild left by the crash.

## Evidence

Vector staging (`internal/cli/reindex.go`, lines 214–227):

```go
dst := src + reindexBackupSuffix
os.Remove(dst)       // deletes prior run's last-known-good file
os.Rename(src, dst)  // moves current possibly-partial file into its place
```

Content-hash staging (`internal/store/sqlite_store.go`, lines 1231–1248):

```sql
DROP TABLE IF EXISTS _reindex_hash_backup;
CREATE TABLE _reindex_hash_backup AS
  SELECT doc_id, content_hash FROM documents;
```

After a crash mid-reindex, current `documents.content_hash` values and current vector files may be cleared/partial, while the `.reindex-old` files and `_reindex_hash_backup` table are precisely the recovery copies. The retry destroys both recovery generations.

The comments claim interrupted reindex recovery, but only handled errors/Ctrl-C that run in-process cleanup are protected.

## Reproduction

1. Start with a fully indexed corpus.
2. Run `reindex`.
3. After old vector files have been renamed to `*.reindex-old` and hashes cleared, force-kill the process (SIGKILL/power-loss simulation) after partial new output exists.
4. Confirm the old backups remain.
5. Run `reindex` again.
6. The retry deletes the good `*.reindex-old` files and backup table, then replaces them with partial/cleared current state.
7. If the second run fails or is cancelled, rollback restores the partial state rather than the original good state.

## Expected behavior

Detect an unfinished prior generation before staging a new one. Preserve or restore the last-known-good generation; never discard it merely because a new retry began.

Possible designs include a transaction/journal with generation IDs and committed markers, or recovery-on-start that deterministically restores/validates the prior backup before taking a new snapshot.

## Suggested regression coverage

- Crash simulation at every staging/rebuild/commit boundary.
- A second run after crash preserves the original good index and hashes.
- Repeated crashes do not rotate partial state into the recovery slot.
- Recovery covers HNSW, disk segment + identity sidecar, and all stale/legacy index files.
- Commit cleanup only removes a generation proven durable.

PR #767, opened 2026-08-05
fix(auth): enforce owner-only permissions on a reused secret.token (#715)

### LINKED ISSUE #715: security: auto auth reuses secret.token without enforcing 0600 permissions
## Summary

Auto auth trusts an existing `<state-dir>/secret.token` without checking or repairing its permissions, despite the canonical requirement that this file be restricted to `0600` on Unix-like systems.

## Evidence

- `dirstral-spec/docs/SPEC.md` §4.2 says `secret.token` permissions **MUST** be restrictive (`0600`).
- `prepareAutoAuthMaterial` calls `readToken(tokenPath, true)`; any non-empty existing token is returned unchanged.
- `readToken` only reads bytes and never stats or chmods the file.
- `writeSecretToken` passes `0600` to `os.OpenFile`, but creation modes do not change an existing file’s permissions.
- Foreground startup creates the state directory with `0755` in `internal/cli/up.go`; `MkdirAll` also does not tighten a pre-existing permissive directory. A `0644` token can therefore be readable by other local users.

## Reproduction

1. Create a state directory with mode `0755`.
2. Write a non-empty `secret.token` and chmod it to `0644`.
3. Start `dir2mcp up` with default/auto auth and otherwise valid provider configuration.
4. Startup reuses that token and publishes its path in `connection.json`; the file remains `0644`.

The same mode-preservation problem applies if `writeSecretToken` truncates an existing permissive file.

## Impact

A copied, restored, or manually-created auto-auth token can silently remain group/world-readable on a shared host. The runtime claims the secured auto-auth path while violating the normative credential-storage requirement.

## Expected behavior

Before accepting an auto-managed `secret.token`, verify it is a regular file owned/usable by the current user and enforce `0600` safely (or fail closed with actionable remediation). Writing an existing token must also result in `0600`.

## Acceptance criteria

- Enforce `0600` for both newly-created and reused auto-managed token files on Unix-like systems.
- Avoid following/replacing unsafe non-regular token paths.
- Add regression tests starting from `0644` and verifying the final mode or a fail-closed error.
- Keep user-supplied `--auth file:<path>` behavior explicit and document whether it is validated or operator-managed.

yidakra:
Closes #715.

## What was actually broken

`prepareAutoAuthMaterial` called `readToken(tokenPath, true)` and used any
non-empty result as-is. `writeSecretToken` passed `0600` to `os.OpenFile`, but a
create mode says nothing about a file that already exists, and `readToken` never
stat'd anything. SPEC §4.2: `secret.token` permissions **MUST** be restrictive
(0600 on Unix-like systems).

I measured the reported scenario through a real `up` before writing anything,
and the picture is more specific than the issue text:

| seeded state | unfixed behaviour |
| --- | --- |
| `.dir2mcp/` 0755 + `secret.token` 0644 | reused, and the file **did** end at 0600 |
| `secret.token` -> symlink to a 0644 file outside the state dir | reused at **0644**, server started as if private, exit 0 |
| same symlink, target trims to empty | generated token **written through the link**, truncating that outside file in place, exit 0 |

The first row is not the fix landing early: #726's `statefs.HardenTree` walks the
state directory in `prepareUpDirectories`, which runs before auth prep, so it
repairs a plain 0644 token as a side effect, and does so silently. Two things
that leaves:

1. `HardenTree` deliberately **skips symlinks** (right for a cache: the mode that
   matters is the target's, and the target may live outside the tree on purpose).
   A symlinked credential is the opposite case, and it is a live hole today: read
   from outside the state directory at an arbitrary mode, and rewritten through
   the link. Rows 2 and 3 above are real, not theoretical.
2. The auto-auth path enforced nothing of its own. Its compliance with §4.2 was
   inherited from a distant caller that happens to run first.

## The fix

`internal/statefs`

- `HardenSecret(path) (prior fs.FileMode, exists bool, err error)`: tightens an
  existing regular file to `FileMode`, reusing the existing "only ever remove
  bits" chmod, and reports the mode it found so a caller can say what it
  repaired. Refuses non-regular paths via `Lstat`, without following them
  (`ErrNotRegular`). A missing path reports `exists=false`, not an error.
- `WiderThanOwnerOnly(mode)`: the "was this exposed" predicate, so call sites do
  not re-derive the bitmask.

`internal/cli/app.go`

- `prepareAutoAuthMaterial` hardens the token before reading or rewriting it.
- `writeSecretToken` uses `statefs.Create`, so rewriting an existing permissive
  token also lands at 0600 (the issue's "same problem applies if
  writeSecretToken truncates an existing permissive file" half). It hardens the open
  descriptor, which also covers a path swapped between the check and the write.
- `--auth file:<path>` documented as operator-managed and left alone (below).

`internal/cli/up.go` is touched on **one line** (`prepareAuthMaterial(cfg,
a.stderr)`), so the warning has somewhere to go. Nothing near
`support_bundle.go` or `service*.go`.

## The three decisions you asked for

**Repair, or refuse? Repair, and say so out loud.** This file is dir2mcp's own in
auto mode, and its mode is lost for mundane reasons: a corpus restored from a
tarball, `cp -r`'d, or created by an older build under a 022 umask. Refusing
turns each of those into a dead daemon for a condition the daemon can fix in one
syscall. But repairing silently would be the wrong kind of quiet, because a
chmod does not un-read a file: if another local account could read the token, it
may already have. So the repair is announced on stderr, naming the mode found,
and it is announced regardless of `--quiet`/`--json` (it goes to stderr; JSON
output is on stdout) because a possibly-exposed credential is not status chatter.

Honest caveat about reach: in `up` today, `HardenTree` usually repairs a plain
0644 token before auth prep sees it, so this warning is mostly a backstop for
the create/rewrite path and for any caller that does not harden the tree first.
Making #726's tree walk itself report what it tightened would be the way to
surface the plain-0644 case loudly, and that belongs in a `HardenTree` change
across `up.go`/`reindex.go`, not here. Say the word and I will file it.

**Rotate, or just chmod?** No automatic rotation, and I want to be explicit
rather than silent about it. A token found world-readable is genuinely suspect,
and chmod does not restore its secrecy. But rotating it on startup invalidates
every client that already holds it (Claude/ElevenLabs registrations,
`connection.json` consumers, anything with the bearer baked into a config), so a
permission repair would silently break a working deployment: the exact failure
mode the "repair, do not refuse" choice was meant to avoid, arriving by a
different door. Whether the exposure matters depends on who else has an account
on this host, which the daemon cannot know. So the warning states plainly that
tightening does not undo the exposure and names the rotation, which is one
command: stop the server, delete `secret.token`, start again (auto mode
regenerates), re-register clients.

**Windows.** No POSIX mode bits. Go synthesizes 0666/0444 from the read-only
attribute and `os.Chmod` can only toggle that attribute, so the tightening
degrades to a no-op there and the file's confidentiality rests on the NTFS ACLs
it inherits from the state directory. No ACL repair is attempted, and this is
stated in the `HardenSecret` doc comment rather than left to be discovered. The
**regular-file refusal is enforced on every platform**. The tests are
`//go:build unix` because `syscall.Umask` does not exist on Windows.

**`--auth file:<path>`** (the issue's last acceptance criterion): explicitly
operator-managed. The path is read and required non-empty; its mode is never
inspected and never changed. A deliberately group-readable token shared with
other processes on the host is a legitimate setup there, and SPEC §17 already
describes that flag as "a user-provided token file with restrictive
permissions". Documented at the call site so the asymmetry with auto mode is
intentional and visible.

## Tests

`tests/security/auth_token_permissions_715_test.go`, `//go:build unix`, forcing
`syscall.Umask(0o022)` via the existing `withPermissiveUmask` helper from the
#726 suite (this class of defect is invisible under a strict umask). They
measure the resulting file rather than asserting on log lines: mode, content,
and whether an outside file survived.

- `TestAutoAuthRefusesASymlinkedToken` and
  `TestAutoAuthDoesNotWriteThroughASymlink` **fail without the source change**.
  Proven by reverting only `internal/cli/app.go` + `internal/cli/up.go` (keeping
  the additive `statefs` helper so the package still compiles) and re-running:

  ```
  --- FAIL: TestAutoAuthRefusesASymlinkedToken (0.42s)
      up served a symlinked secret.token as if it were private
  --- FAIL: TestAutoAuthDoesNotWriteThroughASymlink (0.40s)
      up accepted a symlinked secret.token
  ```

  Both pass with the change restored.
- `TestAutoAuthTightensAWorldReadableTokenAndKeepsIt` (0644 -> 0600, token value
  unchanged), `TestAutoAuthLeavesAPrivateTokenAlone` (0600 untouched, no
  warning), `TestAutoAuthDoesNotWidenAMoreRestrictiveToken` (0400 stays 0400),
  `TestAutoAuthCreatesAFreshTokenOwnerOnly` (fresh token 0600, 64 hex chars).
  These pass on the unfixed tree too, for the `HardenTree` reason above: they
  are policy guards (repair, never widen, never rotate), not the proof.
- Unit coverage on the new primitive: prior-mode reporting for 0644/0640/0600/
  0400, missing path, and refusal of FIFO / directory / symlink with the link
  target measured unchanged.

No new file under `internal/cli/`, so `tests/security/cli_boundary_test.go`
needs no allowlist entry (checked; the boundary suite passes).

`make check` passes locally (fmt, vet, golangci-lint 0 issues, gocyclo -over 15,
ineffassign, misspell, full test suite, build).

### DIFF
--- a/internal/cli/app.go
+++ b/internal/cli/app.go
@@ -1620,7 +1620,20 @@ func ensureRootAccessible(root string) error {
 // prepareAuthMaterial resolves the bearer-token auth material for the
 // configured auth mode: "none" disables auth, "auto" delegates to
 // prepareAutoAuthMaterial, and "file:<path>" loads a token from disk.
-func prepareAuthMaterial(cfg config.Config) (authMaterial, error) {
+//
+// The two token modes have deliberately different contracts. "auto" owns
+// <state-dir>/secret.token: dir2mcp creates it, may rewrite it, and enforces
+// the SPEC §4.2 owner-only requirement on it. "file:<path>" is
+// operator-managed: the path is read and required to be non-empty, and
+// nothing else. Its mode is not checked and never changed, because a
+// deliberately group-readable token shared with other processes on the host
+// is a legitimate setup there, and because that flag is the documented escape
+// hatch (SPEC §17) for a token that must live outside the state directory.
+//
+// warn receives operator-facing security notices; they are written whatever
+// the output mode, because a credential that may have been exposed is not
+// status chatter that --quiet asked to be spared.
+func prepareAuthMaterial(cfg config.Config, warn io.Writer) (authMaterial, error) {
 	mode := strings.TrimSpace(cfg.AuthMode)
 	if mode == "" {
 		mode = "auto"
@@ -1634,7 +1647,7 @@ func prepareAuthMaterial(cfg config.Config) (authMaterial, error) {
 	}
 
 	if strings.EqualFold(mode, "auto") {
-		return prepareAutoAuthMaterial(cfg)
+		return prepareAutoAuthMaterial(cfg, warn)
 	}
 
 	if len(mode) >= len("file:") && strings.EqualFold(mode[:len("file:")], "file:") {
@@ -1670,7 +1683,11 @@ func prepareAuthMaterial(cfg config.Config) (authMaterial, error) {
 // prepareAutoAuthMaterial resolves auth material in "auto" mode: it prefers
 // the env-var token, otherwise reads the persisted secret.token, generating
 // and persisting a new random token when none exists.
-func prepareAutoAuthMaterial(cfg config.Config) (authMaterial, error) {
+//
+// The persisted token is hardened before it is read or rewritten, so an
+// existing file is held to the same SPEC §4.2 owner-only requirement as one
+// this run creates (#715).
+func prepareAutoAuthMaterial(cfg config.Config, warn io.Writer) (authMaterial, error) {
 	if token := strings.TrimSpace(os.Getenv(authTokenEnvVar)); token != "" {
 		return authMaterial{
 			mode:              "auto",
@@ -1680,6 +1697,9 @@ func prepareAutoAuthMaterial(cfg config.Config) (authMaterial, error) {
 		}, nil
 	}
 	tokenPath := filepath.Join(cfg.StateDir, secretTokenName)
+	if err := hardenSecretToken(tokenPath, warn); err != nil {
+		return authMaterial{}, err
+	}
 	token, err := readToken(tokenPath, true)
 	if err != nil {
 		return authMaterial{}, err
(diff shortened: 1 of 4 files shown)


yidakra:
CI note: the `race` job failed once on `TestReindex_AfterCrash_RepeatedCrashesDoNotRotatePartialState` (`tests/cli/reindex_backup_preservation_727_test.go:220`, "content_hash must still be the pre-crash snapshot ... got \"\""), which is #727/#761 territory and touches nothing in this diff. It passed on rerun, and passes locally under `-race` with `-count 3`. Looks like a timing flake in that suite rather than anything this PR introduced; worth a separate look.

yidakra:
@coderabbitai review

PR #768, opened 2026-08-05
fix(watch): enforce resolved-root containment on followed symlinks (#717)

### LINKED ISSUE #717: security: watcher follows newly created file symlinks outside the configured root
## Summary

With `follow_symlinks: true`, the initial discovery walk enforces resolved-root containment, but the live watcher does not. A newly created in-root symlink to a regular file outside the corpus root is accepted and indexed.

This violates SPEC §1 / §7.1 root isolation: no content outside the configured root may be accessible, and followed symlink targets must remain under that root.

## Evidence

In `internal/ingest/watch.go`, `fsWatchLoop.indexableFile` handles a symlink by calling `os.Stat(absPath)` and accepting any regular target (lines 334–345). It never calls `filepath.EvalSymlinks` or `withinResolvedRoot`, even though the initial walker uses those checks.

It then returns the symlink path as `DiscoveredFile.AbsPath` (lines 357–363), and `processDocument` reads through it.

## Reproduction

1. Start watching a local corpus with `follow_symlinks: true`.
2. After the watcher is running, create `<root>/leak.txt` as a symlink to a regular file outside `<root>`.
3. The fsnotify create event reaches `indexableFile`; `os.Stat` sees the external regular target and the file is indexed.
4. Search/open operations can then expose its content.

The same symlink present during initial discovery is correctly rejected, so behavior differs depending on creation time.

## Expected behavior

Watcher reconciliation must apply the same resolved-target containment policy as initial discovery. Any symlink target outside the resolved root should be ignored/rejected and must never be read.

- Start watcher with `follow_symlinks=true`, create an out-of-root file symlink, and assert it is never stored/indexed.
- Cover symlink retargeting from an in-root target to an out-of-root target.
- Keep parity coverage proving an in-root followed symlink still works.

yidakra:
Closes #717.

## What was actually happening (issue framing corrected)

The report is right about the hole and overstates the blast radius. I measured it before touching anything, by pointing an in-root symlink at an out-of-root file after the watcher was already running:

```
doc: {RelPath:leak.txt DocType:text SourceType:filesystem SizeBytes:27
      ContentHash: Status:error
      ErrorMessage:read leak.txt: corpusfs: path escapes the corpus root: "leak.txt"}
```

So:

- **Confirmed:** `fsWatchLoop.indexableFile` accepted the link (`os.Stat` on the link, no `EvalSymlinks`, no containment check), and the out-of-root file went all the way into `processDocument`. The same link present at startup is refused by the discovery walk, so admission depended purely on *when* the link appeared. That is the SPEC §1 / §7.1 violation the issue describes ("follow only if target resolves under root").
- **Not confirmed:** "search/open operations can then expose its content". They cannot today. Every content read goes through `corpusfs` by rel_path (`LocalFS.Open`/`Localize` → `resolveWithinRoot`), which re-resolves symlinks and refuses the escape. `DiscoveredFile.AbsPath` has no non-test consumers left. So the bytes were never read.
- **What did leak:** the out-of-root file's *existence, size and mtime* were written into the index as a document row (`size_bytes: 27` above is the outside file's size), the row is visible to `list_files`, and it burned an `errors` count with a confusing message. Root isolation was being upheld only by a second layer at read time; the discovery-policy layer was not enforcing it at all.

Severity is therefore metadata disclosure + index pollution, not content disclosure. The fix is the same either way: the watcher must apply the discovery policy itself rather than depend on a downstream check.

## The fix

One rule, one implementation. `internal/corpusfs` already owned the containment check (`discoverWalker.handleSymlink` → `isWithinRoot`); it now exposes it and calls it itself:

- `corpusfs.ResolveRoot(absRoot)` — the resolved containment anchor (was inlined in `Walk`, and duplicated in `ingest.resolvedRoot`).
- `corpusfs.ResolveSymlinkWithinRoot(rootResolved, linkPath)` — `EvalSymlinks` + `isWithinRoot`, returning the resolved target. `handleSymlink` now calls it, so the discovery walk and the watcher execute *the same function*, not two copies of the same idea.

`internal/ingest/watch.go` deletes its parallel `resolvedRoot`/`withinResolvedRoot` copies and routes every containment decision through those helpers. Net: one containment rule in the tree instead of two, which is the actual defence against this class recurring.

Change is confined to `internal/ingest` + `internal/corpusfs`. **`internal/cli/app.go` is untouched.**

### Points the issue asked me to get right

**Resolved target, resolved at handling time (TOCTOU).** `followedSymlinkTarget` runs on every debounced job, and `ResolveSymlinkWithinRoot` re-runs `EvalSymlinks` every call; nothing is cached from creation time. The resolved path is then what goes downstream as `DiscoveredFile.AbsPath` (matching what the discovery walker records), so a consumer reading the absolute path reads the target that was checked, not a link that may have been repointed since.

What this **does** guarantee: a link is judged on the target it has when its event is processed, and the value handed to the pipeline is the checked real path (no symlink components left to swap).

What it **does not** guarantee: it is a check-then-use, so a repoint in the window between `EvalSymlinks` and the read is not excluded, and a rename of the *resolved* file itself is not either. Two things bound that: the read boundary in `corpusfs` re-resolves and refuses out-of-root paths independently, and the safety rescan reconciles through the corpusfs walk. This PR closes the policy gap; it is not a TOCTOU-proof gate, and I would not claim one without an openat2/`O_NOFOLLOW`-style resolve-and-hold, which is out of scope here.

**`follow_symlinks: false` unaffected.** The not-following branch returns before any resolution, exactly as before, and is pinned by a test that also asserts an *in-root* link is still refused when following is off (i.e. the policy did not accidentally loosen).

**In-root symlinks still work.** Not fixed by refusing symlinks. Covered by a positive control in the escape test, plus the pre-existing `TestWatch_FollowsSymlinkedDir` (passes).

**Directory vs file symlinks — both checked.** They are genuinely different paths in the watcher:
- File links: `handleEvent` → `arm` → `process` → `indexableFile`. This is where the hole was.
- Directory links: `handleEvent` (`os.Stat` says dir) → `registerNewTree` → `walkWatchTree`, and initial registration via `addWatchDirs` → `walkWatchTree`. `walkWatchTree` *was* already enforcing containment on entry, so out-of-root directory links were never watched; it now enforces it via the shared helper, and `registerNewTree` logs the refusal instead of returning silently.
- Side effect of routing `walkWatchTree` through the shared helper: an unresolvable directory is now refused rather than falling back to its lexical path. For the walk entry that is the corpus root, so `addWatchDirs` now returns an explicit error ("corpus root did not resolve; no directories watched") rather than registering zero watches in silence. `Watch` already logs that non-fatally and the safety rescan still covers.

**Observability.** A refusal is logged, following the existing precedent for the analogous S3 case (`OnUnsafeKey`, #735):

```
watch: refusing leak.txt: symlink target does not resolve inside the corpus root; not indexed
watch: refusing directory /corpus/linked: it does not resolve inside the corpus root; not watched or indexed
```

Silence here would be indistinguishable from "the watcher missed my file", which is exactly how the next report of this would get misfiled. The test asserts on the log line, not just on absence from the store.

## Tests

`tests/ingest/watch_symlink_escape_717_test.go`, deliberately **not** gated behind `RUN_INTEGRATION_TESTS` (unlike the older watcher tests) — this is a root-isolation regression and must run on every `make check`.

1. `TestWatch_RefusesSymlinkEscapingRoot_717` — link created *after* the watcher starts, pointing at a file outside the root: never stored, and the refusal is logged. Waiting for the refusal line first also proves the watcher *saw* the link, so the "not indexed" assertion cannot pass vacuously. Positive controls in the same test: an in-root symlink still indexes, and a plain regular file still indexes.
2. `TestWatch_RefusesRetargetedSymlink_717` — a link indexed while in-root, then repointed outside via an atomic `rename` over it: the document must never take on the out-of-root content (content hash unchanged, or the document gone).
3. `TestWatch_SymlinkPolicyUnchangedWhenNotFollowing_717` — with `follow_symlinks=false`, neither the out-of-root nor the in-root link is indexed, and regular files still are.

**Proven to fail first.** With only the source change stashed (`git stash push -- internal/ingest/watch.go internal/corpusfs/local.go`), test 1 fails on both assertions:

```
--- FAIL: TestWatch_RefusesSymlinkEscapingRoot_717 (5.22s)
    watcher indexed leak.txt, a symlink to a file outside the corpus root (issue #717)
    no log line refused leak.txt; a silent drop is indistinguishable from a missed event. Log:
        watch: index leak.txt: read leak.txt: corpusfs: path escapes the corpus root: "leak.txt"
```

Honest caveat on test 2: on darwin I measured that repointing an existing symlink (both `remove`+`recreate` and `rename`-over) produces **no delivered event** — fsnotify's kqueue backend diffs directory listings, and a same-name replace nets zero change. So test 2 is vacuous on macOS and is expected to do real work on inotify (CI). Test 1 does not depend on that: it creates a new name. I kept test 2 because it is deterministic-pass on both (never flaky) and does exercise the repoint path where the platform delivers it.

Stability: `go test ./tests/ingest -run 717 -count=15` green (and `-count=10` again after the final restructure); `RUN_INTEGRATION_TESTS=1 go test ./tests/ingest -run TestWatch` green, including the pre-existing symlinked-directory watch test.

## Checks

- `make check` passes (fmt, vet, golangci-lint 0 issues, gocyclo `-over 15`, ineffassign, misspell, full `go test ./...`, build).
- No spec change: this makes the code match SPEC §7.1 as already written ("if enabled: follow only if target resolves under root"), so no `dirstral-spec` PR is needed.

### DIFF
--- a/internal/corpusfs/local.go
+++ b/internal/corpusfs/local.go
@@ -54,11 +54,7 @@ func (l *LocalFS) Walk(ctx context.Context, _ string, opts Options) ([]Discovere
 		return nil, fmt.Errorf("root is not a directory: %s", absRoot)
 	}
 
-	rootResolved := absRoot
-	if resolved, err := filepath.EvalSymlinks(absRoot); err == nil {
-		rootResolved = resolved
-	}
-	rootResolved = filepath.Clean(rootResolved)
+	rootResolved := ResolveRoot(absRoot)
 
 	files := make([]DiscoveredFile, 0, 256)
 	walker := discoverWalker{
@@ -144,6 +140,42 @@ func resolveWithinRoot(root, relPath string) (string, error) {
 		return "", fmt.Errorf("%w: %q", ErrPathEscapesRoot, relPath)
 	}
 	return resolved, nil
+}
+
+// ResolveRoot returns the canonical (symlink-resolved, cleaned) form of an
+// absolute corpus root. It is the anchor every root-containment decision is made
+// against, so callers outside this package (the ingest watcher) compute it the
+// same way the discovery walk does. A resolve failure falls back to the lexical
+// form: the root is expected to exist, and a usable anchor is better than none.
+func ResolveRoot(absRoot string) string {
+	if resolved, err := filepath.EvalSymlinks(absRoot); err == nil {
+		return filepath.Clean(resolved)
+	}
+	return filepath.Clean(absRoot)
+}
+
+// ResolveSymlinkWithinRoot resolves linkPath (typically an in-root symlink) and
+// enforces that its target stays inside rootResolved, which must already be a
+// resolved root (see ResolveRoot). It returns the resolved target and ok=true
+// only when the link resolves AND lands in the corpus; ok=false means the caller
+// must treat the path as if it did not exist.
+//
+// This is the single implementation of the followed-symlink containment rule
+// required by SPEC §1/§7.1 root isolation: the discovery walk (handleSymlink
+// below) and the live watcher (internal/ingest/watch.go) both call it, so the
+// two cannot drift into judging the same link differently depending on when it
+// appeared (issue #717). Resolution happens on every call, never cached, so the
+// decision is always made against the link's current target.
+func ResolveSymlinkWithinRoot(rootResolved, linkPath string) (string, bool) {
+	resolved, err := filepath.EvalSymlinks(linkPath)
+	if err != nil {
+		return "", false
+	}
+	resolved = filepath.Clean(resolved)
+	if !isWithinRoot(rootResolved, resolved) {
+		return "", false
+	}
+	return resolved, true
 }
 
 func shouldSkipDirectory(name string) bool {
@@ -500,12 +532,8 @@ func (w *discoverWalker) handleSymlink(ctx context.Context, symlinkPath, relPath
 		return err
 	}
 
-	resolvedPath, err := filepath.EvalSymlinks(symlinkPath)
-	if err != nil {
-		return nil
-	}
-	resolvedPath = filepath.Clean(resolvedPath)
-	if !isWithinRoot(w.rootResolved, resolvedPath) {
+	resolvedPath, ok := ResolveSymlinkWithinRoot(w.rootResolved, symlinkPath)
+	if !ok {
 		return nil
 	}
 
--- a/internal/ingest/watch.go
+++ b/internal/ingest/watch.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/dirstral/dir2mcp/internal/corpusfs"
 	"github.com/fsnotify/fsnotify"
 )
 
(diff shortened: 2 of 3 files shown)


yidakra:
@coderabbitai review

(The bot's first pass was rate limited. The CodeRabbit CLI review of the same commit returned 0 findings across all three changed files; requesting the bot pass for the record.)

yidakra:
@coderabbitai review

Issue #497, opened 2026-07-02
ingest: default 20MB file cap silently drops media files at discovery (skipped=0, no signal)


yidakra:
## Summary
The default ingest file-size cap is small (`corpusfs.defaultMaxFileSizeBytes = 10 MiB`; config default `ingest_max_file_mb: 20`), so **media files are silently dropped at discovery** — not scanned, not indexed, and **not reported as skipped** in a discoverable way (`status` shows `skipped=0`).

## Impact
For an audio/video corpus this is a silent no-op: `dir2mcp up` reports `scanned=N` counting only the small files, and the operator has no signal that 300 MB mp4s were excluded. I lost time before finding `ingest.max_file_mb`.

## Repro
- Corpus with a 23 MB media file + a tiny config file, `dir2mcp up`.
- `status` → only the tiny file appears (`scanned=1`), `skipped=0`. The media file is absent with no explanation.
- Set `ingest: { max_file_mb: 4000 }` → the media file is discovered.

## Suggested fix
- Surface size-based exclusions explicitly (a `skipped_oversize` counter and/or a per-file reason in `status`/`doctor`), so silent drops become visible.
- Consider a higher/uncapped default for recognized media types (audio/video), whose whole point is to be large — or at least call out the cap prominently in the media docs.

## Refs
`internal/ingest/discover.go` (over-limit skip), `internal/corpusfs/*` (`defaultMaxFileSizeBytes`), config `ingest_max_file_mb` (default 20).

PR #772, opened 2026-08-05
fix(ingest): complete the default ignore list and stop dropping valid archive members (#716, #718)

### LINKED ISSUE #716: ingest: default discovery does not exclude dist, build, or .venv
## Summary

The canonical default discovery contract excludes `dist/`, `build/`, and `.venv/`, but both local and S3 discovery currently enumerate and ingest those directories.

## Evidence

- `dirstral-spec/docs/SPEC.md` §7.1 and `docs/specs/behavior/bs-002-ingestion-pipeline.md` list `.git/`, `node_modules/`, `dist/`, `build/`, `.venv/`, and `.dir2mcp/` as the default ignore set.
- `internal/corpusfs/corpusfs.go` `defaultExcludedDirs` contains `.git`, `.dir2mcp`, `node_modules`, `vendor`, and `__pycache__`, but omits `dist`, `build`, and `.venv`.
- Local discovery calls `shouldSkipDirectory`; S3 discovery calls `keyHasExcludedDir`; both use that incomplete map.
- The later default `security.path_excludes` list also omits all three, so the ingest service does not filter them after discovery.
- Existing LocalFS/S3 tests cover `.git` and `node_modules` but not the three normative omissions.

## Reproduction

1. Create a corpus containing:
   - `dist/bundle.js`
   - `build/generated.txt`
   - `.venv/lib/site-packages/pkg.py`
2. Run default discovery/indexing with no custom ignores.
3. All three files are returned by CorpusFS and proceed to ingestion, contrary to §7.1.

The same behavior occurs for S3 keys under those path segments.

## Impact

Default scans ingest generated bundles, build artifacts, and entire Python environments. This inflates scan/embed cost, creates duplicate/noisy retrieval results, and can pull generated dependency/config content into the corpus that operators were promised would be ignored.

## Expected behavior

The default excluded-directory set should implement the complete normative list consistently for LocalFS and S3, with downstream path exclusions aligned where needed.

## Acceptance criteria

- Add `dist`, `build`, and `.venv` to default discovery exclusions.
- Add parity tests for local and S3 backends.
- Verify a regular file whose basename is one of these tokens is still allowed, matching the existing directory-vs-file rule.
- Reconcile the runtime defaults, config documentation, and canonical spec if any omission is intentional.

### LINKED ISSUE #718: ingest: archive traversal check silently drops valid member names containing two dots
## Summary

Archive ingestion rejects every member path containing the substring `..`, not only actual parent-directory traversal segments. Valid files such as `v1..v2.txt` and `draft..final/report.md` are silently omitted.

The single-compressed-member validator has the same false positive, so a valid name such as `report..final.txt.gz` is replaced with the synthetic `member` name.

## Evidence

`internal/ingest/archive.go`:

- `isSafeArchivePath` returns false for `strings.Contains(p, "..")` (lines 87–95).
- `isSafeArchiveMemberName` likewise rejects any `..` substring (lines 98–109).
- ZIP/TAR extraction skips unsafe members rather than surfacing an error, making the data loss easy to miss.

A `..` substring inside a path segment is not traversal. Only a segment equal to `..` (after normalizing both slash styles and rejecting absolute/escaping paths) is unsafe.

## Reproduction

Create a ZIP or TAR containing `v1..v2.txt` and `draft..final/report.md`, then ingest it. Neither member is indexed, although both are safe relative paths.

For a bare compressed file named `report..final.txt.gz`, the derived member path loses the original safe stem.

## Expected behavior

Reject absolute paths and paths whose normalized segments escape the archive root, but accept ordinary segments containing consecutive dots.

- Accept `v1..v2.txt`, `draft..final/report.md`, and similar benign names.
- Continue rejecting `../escape`, `a/../../escape`, backslash traversal, absolute paths, and drive/UNC-style escapes.
- Verify ZIP, TAR, and single-compressed-member naming paths.

yidakra:
Closes #716. Closes #718.

Both issues are ingest-side discovery/admission, and both were verified against
the code before anything was changed. Neither is stale.

## #718: the archive traversal check refused valid member names

`isSafeArchivePath` rejected any member path *containing* the substring `..`.
A `..` inside a segment is an ordinary filename, so `v1..v2.txt` and
`draft..final/report.md` were dropped, and the drop was invisible: the archive
indexed fine, the member simply was not there. `isSafeArchiveMemberName` had the
same test, so `report..final.txt.gz` lost its stem to the synthetic `member`.

### The rule, and why it is not `relpath.Normalize` verbatim

`internal/relpath` (#735) is the backend-independent containment rule and it has
the final word here: whatever `archiveMemberRelPath` returns must be a path
`relpath.Normalize` accepts, so a member can never be admitted under a name S3
discovery or the local walker would refuse. Containment inside `<archive>/...`
then holds by construction, since the caller prefixes the archive's own
(already valid) rel_path.

What is deliberately *different* is the reject-vs-clean call. relpath refuses a
path that would change under normalization because an S3 key is the address used
to re-fetch the object: rewriting `a//b.txt` to `a/b.txt` would make discovery
record a rel_path that `Open` later resolves to a **different object**. An
archive member has no such property. Its bytes are read from the archive stream
by position and the name is never used to re-fetch anything, so cleaning cannot
retarget.

And the difference is not academic. Measured:

```
$ tar -czf out.tgz .      # the ordinary way to make a tarball
$ tar -tzf out.tgz
./
./sub/
./a.txt
./sub/b.txt
```

`relpath.Normalize("./a.txt")` returns `ErrNotRelative`. Applying the S3 rule
verbatim would refuse **every member of an ordinary tarball**, trading one
silent-drop bug for a far bigger one. So normalization noise (`./`, `//`,
trailing `/`) is cleaned; everything that could mean two different things is
refused:

| shape | verdict |
| --- | --- |
| `v1..v2.txt`, `draft..final/report.md`, `notes...md` | admitted (this is the bug being fixed) |
| `./a.txt`, `sub//b.txt` | admitted, stored cleaned |
| `../escape.txt`, `a/../../escape.txt`, `..` | refused (`..` segment) |
| `a/../b.txt` | refused, even though it does not escape: cleaning it to `b.txt` would file the member under a name the archive gave to a **different** entry |
| `/absolute.txt`, `//absolute.txt` | refused |
| `..\escape.txt`, `\\server\share\x`, `C:\Windows\x` | refused: a backslash is a legal byte in a POSIX member name and a separator on Windows, where `..\x` escapes exactly as `../x` does. Same call relpath makes. |

The one risk cleaning introduces is two distinct members collapsing onto one
rel_path. That is bounded to inside the archive's namespace, and it is reported
(see below) rather than silent. Worth noting: the store *already* cleaned
rel_paths on upsert, so this collapse was happening one layer down anyway; the
change makes it explicit at admission time, where it can be observed.

### A dropped member is now observable

Following the `Options.OnUnsafeKey` (#735) and `OnOversize` (#497) precedent,
extraction returns an admission record (`archiveExtraction.Skips`) and the
service logs it with each member's name and reason:

```
archive test.zip: 1 member(s) not indexed: "../../etc/passwd" (member name has a `..` segment) (#718)
```

This covers refused names, per-member size-cap drops, open/read failures, the
single-compressed synthetic-name substitution, and rel_path collisions. The
count is exact; the per-member sample is capped at 20 so a hostile archive with
100k bad entries cannot turn an observability record into a memory problem.
Log-only, like `OnUnsafeKey`: there is no usable rel_path to key a skipped
document row on, because the unusable name **is** the finding.

### Same defect in the store, found while fixing this

`store.normalizeRelPath` used `strings.Contains(normalized, "/..")`, so an
ordinary corpus file in a subdirectory whose name starts with two dots
(`sub/...notes.md`, `sub/..draft.txt`) was refused on upsert. No archive
involved. Measured on a plain local corpus of five files before the fix:

```
doc "...notes.txt" ok      doc "..hidden.txt" ok      doc "ok.txt" ok
doc "v1..v2.txt"   ok      (sub/...x.md: absent)
snapshot: Scanned:5 Indexed:4 Errors:1
```

One anonymous error, no document row, no path in the log. The substring test
could never have caught a real traversal the segment test misses, because
`filepath.Clean` hoists every surviving `..` to the front of the path, so a `..`
segment after the first is unreachable by construction and the leading case is
already covered by the existing `../` prefix check. Removing it weakens nothing.

The segment test now lives in one place, `relpath.HasDotDotSegment`, since
re-implementing it as a substring test is exactly what caused this bug twice.

## #716: `dist/`, `build/`, `.venv/` missing from default discovery

SPEC 7.1 makes the default ignore list `.git/`, `node_modules/`, `dist/`,
`build/`, `.venv/`, `.dir2mcp/`. `corpusfs.defaultExcludedDirs` had five names
and none of the three, on both LocalFS and S3. `internal/ingest/gitignore.go`
kept a **second copy** of the map for the file watcher, which is how the two
were able to drift; the watcher now delegates to `corpusfs.IsExcludedDir` so
there is one list.

### Compatibility: yes, a reindex will drop documents. That is the point.

Anything currently indexed under `dist/`, `build/`, or `.venv/` becomes
undiscoverable and is tombstoned on the next scan. I think that is acceptable:
the spec promised these were ignored, so every such document is content the
operator was told would not be there, and it is generated output (bundles, build
artifacts, an entire vendored Python environment) rather than authored source.
It also removes real cost, since those trees dominate embedding spend and
duplicate retrieval results on any repo-shaped corpus.

### The escape hatch question, answered honestly: there is none

I checked. The excluded-directory set is **hardcoded and not configurable**.
`security.path_excludes` is operator-configurable but only *adds* exclusions,
`.gitignore` support likewise, and there is no `.dir2mcpignore`. So an operator
who genuinely wants `dist/` indexed has no way to say so, and today has the same
problem with `node_modules/`, `vendor/`, and `__pycache__/`.

This PR does not invent a config key for it, because the config surface is
spec-governed and adding one is a spec-first change (a `dirstral-spec` PR first,
then a gated submodule re-pin). This PR is pure conformance to a spec section
that already exists, so it needs no spec change at all. Filed #773 for the
override key as a spec-first follow-up. The tradeoff is stated in `defaultExcludedDirs`'s
doc comment so the next person adding a name knows the list is a hard floor.

## Tests

Each new test was proven to fail first by stashing **only** the source change
(`git stash push -- internal`) and re-running:

| test | fails without fix |
| --- | --- |
| `TestLocalFSWalk_ExcludesSpecDefaultDirs` | yes |
| `TestS3FSWalk_ExcludesSpecDefaultDirs` | yes |
| `TestDiscoverFiles_ExcludesSpecDefaultDirs` | yes |
| `TestArchiveMember_BenignDoubleDotNamesIndexed` (zip + tar.gz) | yes |
| `TestArchiveMember_SingleCompressedNamePreserved` | yes |
| `TestArchiveMember_TraversalShapesRefused_Zip` / `_Tar` (14 shapes each) | yes (4 shapes each) |
| `TestCorpusFile_LeadingDotsInSubdirIndexed` | yes |
| `TestArchiveMember_RefusalIsObservable` | yes |
| `TestLocalFSWalk_SpecDefaultDirNamesAsFilesStillDiscovered` | no, by design (guard: a *file* named `dist` must still be indexed) |
| `TestArchiveMember_DotSlashPrefixedTarMembersIndexed` | no, by design (guard: the rule must not be tightened into relpath's reject-don't-clean form) |

Both directions of #718 are covered: the valid members that were being dropped
are now indexed, and every traversal shape is still refused with no document
produced under any name, while the safe member in the same archive still lands.

`make check` passes (exit 0), gocyclo `-over 15` included.

## Notes for review

- Not a live escape on the platforms dir2mcp ships. Before this change,
  `/absolute.txt` and `C:\Windows\x` members landed *inside* `test.zip/` anyway
  because the store cleaned the rel_path. The value of the tightened rule is
  Windows correctness and path-shape hygiene, not patching an exploitable hole.
  The false-positive data loss was the live problem.
- A member name containing a literal backslash was accepted before and is
  refused now. Rare on POSIX, refused for the reason relpath refuses it, and no
  longer silent.

### DIFF
--- a/internal/corpusfs/corpusfs.go
+++ b/internal/corpusfs/corpusfs.go
@@ -13,6 +13,7 @@ import (
 	"context"
 	"io"
 	"io/fs"
+	"strings"
 )
 
 // defaultMaxFileSizeBytes mirrors the ingest discovery cap so a zero/negative
@@ -26,12 +27,41 @@ func DefaultMaxFileSizeBytes() int64 {
 
 // defaultExcludedDirs are directory names skipped during discovery regardless of
 // backend. For S3 these are matched against path segments of the object key.
+//
+// SPEC §7.1 makes `.git/`, `node_modules/`, `dist/`, `build/`, `.venv/`, and
+// `.dir2mcp/` the normative default ignore list; `vendor/` and `__pycache__/`
+// are dir2mcp additions of the same kind (checked-in dependency trees and
+// generated caches). `dist`, `build`, and `.venv` were missing until #716, so a
+// default scan indexed generated bundles, build output, and entire Python
+// virtualenvs that operators had been told were ignored.
+//
+// This set is a hard floor: there is no config key that re-includes an excluded
+// directory (`security.path_excludes` only adds exclusions). Adding a name here
+// therefore removes documents from every existing corpus on the next reindex, so
+// it stays anchored to the spec rather than growing by taste.
 var defaultExcludedDirs = map[string]struct{}{
 	".git":         {},
 	".dir2mcp":     {},
 	"node_modules": {},
 	"vendor":       {},
 	"__pycache__":  {},
+	"dist":         {},
+	"build":        {},
+	".venv":        {},
+}
+
+// IsExcludedDir reports whether a DIRECTORY with this name is skipped by default
+// discovery. It is exported so the ingest file watcher applies the same list as
+// the initial scan: the watcher used to keep its own copy of the map, and the
+// copies drifted (#716).
+//
+// The rule is directory-only. A regular FILE named `dist` or `vendor` is an
+// ordinary corpus document and must still be discovered, so callers must only
+// consult this for entries they already know are directories (the S3 backend
+// tests it against ancestor key segments only).
+func IsExcludedDir(name string) bool {
+	_, ok := defaultExcludedDirs[strings.TrimSpace(name)]
+	return ok
 }
 
 // DiscoveredFile holds metadata collected during corpus discovery.
--- a/internal/corpusfs/corpusfs.go
+++ b/internal/corpusfs/corpusfs.go
@@ -13,7 +13,6 @@ import (
 	"context"
 	"io"
 	"io/fs"
-	"strings"
 )
 
 // defaultMaxFileSizeBytes mirrors the ingest discovery cap so a zero/negative
@@ -59,8 +58,14 @@ var defaultExcludedDirs = map[string]struct{}{
 // ordinary corpus document and must still be discovered, so callers must only
 // consult this for entries they already know are directories (the S3 backend
 // tests it against ancestor key segments only).
+//
+// The match is exact. The name is a real directory entry or object key segment,
+// and ` dist ` is a different directory than `dist`: the old TrimSpace here made
+// discovery skip a legitimately-named tree whose documents nothing else would
+// have excluded, which is the same silent over-exclusion this change set exists
+// to remove.
 func IsExcludedDir(name string) bool {
-	_, ok := defaultExcludedDirs[strings.TrimSpace(name)]
+	_, ok := defaultExcludedDirs[name]
 	return ok
 }
 
--- a/internal/corpusfs/local.go
+++ b/internal/corpusfs/local.go
@@ -179,8 +179,7 @@ func ResolveSymlinkWithinRoot(rootResolved, linkPath string) (string, bool) {
 }
 
 func shouldSkipDirectory(name string) bool {
-	_, ok := defaultExcludedDirs[strings.TrimSpace(name)]
-	return ok
+	return IsExcludedDir(name)
 }
 
 type discoverWalker struct {
--- a/internal/corpusfs/s3.go
+++ b/internal/corpusfs/s3.go
@@ -388,7 +388,7 @@ func (s *S3FS) discoveredFromObject(obj s3types.Object, opts Options) (Discovere
 func keyHasExcludedDir(rel string) bool {
 	segs := strings.Split(rel, "/")
 	for _, seg := range segs[:len(segs)-1] {
-		if _, ok := defaultExcludedDirs[seg]; ok {
+		if IsExcludedDir(seg) {
 			return true
 		}
 	}
--- a/internal/ingest/archive.go
+++ b/internal/ingest/archive.go
@@ -9,8 +9,11 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
+
+	"github.com/dirstral/dir2mcp/internal/relpath"
 )
 
 const archiveMemberMaxBytes = 10 * 1024 * 1024 // 10 MiB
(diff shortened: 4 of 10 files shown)


yidakra:
@coderabbitai review

yidakra:
@coderabbitai review

Issue #732, opened 2026-08-03
perf(store): ListEmbeddedChunkMetadata is O(N^2) across the startup warm-load and never uses the #631 read pool


yidakra:
Found while measuring #429 F6 (PR #731). Fixing F6 in `internal/cli` took cold start on a 100k-chunk corpus from 57.5s to 8.4s, but ~7s of what is left is a single store query shape, and it is quadratic, so it gets much worse as a corpus grows.

## What is wrong

`SQLiteStore.ListEmbeddedChunkMetadata` builds its page like this:

```sql
WITH filtered_chunks AS (
  SELECT ... FROM chunks c
  WHERE c.embedding_status = 'ok' AND c.deleted = 0 AND c.chunk_id > ?   -- keyset cursor
),
ranked_spans AS (
  SELECT ..., ROW_NUMBER() OVER (PARTITION BY s.chunk_id ORDER BY s.span_id) AS rn
  FROM spans s JOIN filtered_chunks fc ON fc.chunk_id = s.chunk_id
)
SELECT ... FROM filtered_chunks fc
LEFT JOIN ranked_spans sp ... LEFT JOIN documents d ...
WHERE fc.index_kind = ?      -- <- the kind filter lands AFTER the CTE
ORDER BY fc.chunk_id LIMIT ?
```

The `index_kind` predicate and the `LIMIT` apply *after* `filtered_chunks` (and the `ranked_spans` window over its spans) have been materialised. So every page scans every embedded chunk above the cursor, and the startup warm-load costs **O(N^2 / pageSize)**, not O(N).

Measured on a synthetic 100k-chunk corpus (50k text + 50k code):

| page size | full preload |
| --- | --- |
| 500 | 67.2s |
| 2000 | 19.0s |
| 5000 | 9.4s |
| 20000 | 4.1s |

The per-kind split shows the shape directly: the `text` walk (chunk_ids 1..50k, so most rows are still above the cursor on every page) took 44-49s while `code` (50k..100k) took 15-17s for the same row count.

PR #731 raised the caller's page size 500 -> 5000, which is a constant-factor mitigation only. At 1M chunks the same shape predicts roughly 100x the 100k figure.

## Second finding: the listing never reaches the read pool

[...]

Issue #176, opened 2026-05-11
list_files emits malformed rel_path for filenames containing `--` (causes open_file FILE_NOT_FOUND)


yidakra:
## Summary

For source files whose name contains a double-dash (`--`), `dir2mcp_list_files` returns a malformed `rel_path` with a phantom `-/` prefix and the wrong extension. Subsequent `dir2mcp_open_file` calls against that rel_path fail with `FILE_NOT_FOUND`.

## Reproduction

Corpus: BVI legal corpus (stas-legal). One file has a `--` in its name:

```
$ ls /Users/adeliev/Development/stas-legal | grep AML | grep 2025
SINo3of2025--AML(Amendment)Regulations2025.pdf
```

Call `list_files`:

```jsonc
// dir2mcp_list_files {"limit": 1}
{
  "rel_path": "-/SINo3of2025--AML(Amendment)Regulations2025.md",
  "doc_type": "md",
  "size_bytes": 3140,
  ...
}
```

Two things wrong with that `rel_path`:
1. Phantom `-/` directory prefix — there is no `-/` directory on disk.
2. `.md` extension — the source file is `.pdf`. The OCR cache lives at `.dir2mcp/cache/ocr/<sha>.md`, but that's an internal artifact and shouldn't surface as the file's external rel_path.

Round-trip:

```jsonc
// dir2mcp_open_file {"rel_path": "-/SINo3of2025--AML(Amendment)Regulations2025.md"}
{
  "structuredContent": {
    "error": {"code": "FILE_NOT_FOUND", "message": "file not found", "retryable": false}
  },
  "isError": true
}
```

In Claude Desktop this surfaces as `Failed to call tool "dir2mcp_open_file"` — which is how this was found.

## Scope

Out of 91 files in the test corpus, 1 has the bug. All other rel_paths round-trip correctly. The common factor on the affected file is the `--` (with parens, but other parens-containing names round-trip fine — so it's the double-dash that's load-bearing).

## Likely cause

Some path normalization step appears to split the filename on `--` and treat the leading segment as a directory:
- `SINo3of2025--AML...pdf` → split → `["SINo3of2025-", "AML...pdf"]`
- The first segment is then trimmed/transformed and the file is registered as `<segment>/<rest>.md` against the OCR cache.

Worth a unit test that walks a synthetic corpus with adversarial filenames (`--`, leading `-`, parens, spaces, unicode) and asserts `list_files` rel_paths round-trip through `open_file`.

## Impact

[...]

PR #775, opened 2026-08-05
fix(mcp): stop resurrecting deleted sessions and rescanning the corpus per list_files page (#694, #696)

### LINKED ISSUE #694: mcp: every list_files page rescans and stats the entire matching corpus
## Summary

The store now has efficient/cached `ListFiles` pagination, but the MCP handler wraps it in `listFilesFiltered`, which walks **every raw store row from offset zero on every tool call** to recompute hidden/resolvability filtering and the visible total. A request for `limit=1, offset=0` on a million-file local corpus still issues roughly 2,000 store page reads and up to a million path-resolution/stat operations.

This effectively defeats the `ListFiles` paging work from #429 F10 at the public tool boundary.

## Evidence

- `internal/mcp/tools.go:629-677` initializes `storeOffset := 0` for every request and loops `store.ListFiles(..., pageSize=500, storeOffset)` until `storeOffset >= storeTotal`.
- It continues after `collected` reaches the requested `limit` because it computes an exact visible `total`.
- `internal/mcp/tools.go:663-665` calls `isResolvableSourceWithRoot` for every filesystem row.
- `internal/mcp/tools.go:737-769` runs `filepath.EvalSymlinks` per candidate, which performs filesystem I/O.
- The next client page (`offset=200`) repeats the entire scan/stat walk from zero.
- Store-level #429 tests validate efficient `SQLiteStore.ListFiles` paging, but no MCP-level test bounds store calls or path-resolution work.

## Additional correctness problem

The single MCP response is assembled from many independent store calls rather than one SQLite read snapshot. If ingest inserts/deletes rows before the current offset between batches, offset pagination can duplicate or skip rows and the final `visibleSeen`/`total` can describe a mixture of database states. The store cache's `data_version` invalidation keeps each call fresh but cannot make the multi-call handler atomic.

## Impact

- `list_files` latency and filesystem load are O(total matching documents), independent of requested page size.
- Walking all client pages is O(number_of_pages × total), i.e. quadratic in corpus size.
- Concurrent ingest can make a page internally inconsistent even though it is one tool call.
- Slow network filesystems multiply the per-row `EvalSymlinks` cost.

This is distinct from #684 (S3 rows incorrectly gated through the local filesystem) and is a follow-up to #429 F10 (store-level pagination/count cost).

## Expected behavior

Filtering, counting, and pagination should occur over one authoritative/snapshot-consistent visible set, and work per request should be proportional to the page plus a bounded/count operation—not a full per-row filesystem walk.

## Acceptance criteria

- Push hidden filtering into the store query/path predicate where possible.
- Replace per-request path existence validation with ingest-time tombstoning/reconciliation or a bounded/cached validation strategy; do not stat the entire corpus for one page.
- Return an exact visible `total` without materializing every row on every request (cached count keyed by filter + visibility policy is acceptable with commit invalidation).
- Ensure one response observes a consistent SQLite snapshot or uses keyset/cursor semantics robust to concurrent writes.
- Add MCP-level instrumentation tests asserting a small first page does not read/stat the whole corpus and later pages do not restart from zero.
- Add concurrent insert/delete tests proving no duplicates/skips within one response.
- Preserve #176's guarantee that emitted local paths round-trip through `open_file` without turning every list request into a corpus-wide filesystem audit.

### LINKED ISSUE #696: mcp: a concurrent session touch can resurrect a deleted session after restart
## Bug

Session mutation is ordered only in memory. The persistence calls happen after releasing sessionMu:

- hasActiveSession updates lastSeen, unlocks, then calls persistSession (internal/mcp/server.go:608-640).
- forgetSession deletes the in-memory entry, unlocks, then calls deletePersistedSession (server.go:652-660).
- cleanupExpiredSessions follows the same delete-after-unlock pattern (server.go:717-739).

This admits a stale-write race:

1. Request A validates an active session, updates lastSeen, releases sessionMu, and pauses before/during UpsertMCPSession.
2. Request B handles DELETE for that session, removes it from memory, and completes DeleteMCPSession.
3. Request A resumes and its older upsert lands after the delete.
4. The current process still rejects the ID because its memory entry is gone, but a restart calls restoreSessions and loads the resurrected row as active.

The same ordering can undo periodic expiration cleanup.

## Impact

Explicit session termination and expiry are not durable under concurrent traffic. A terminated MCP session can become valid again after daemon restart until its refreshed inactivity timeout expires.

## Expected

Persistence ordering must preserve the newest logical session transition. For example, serialize touch/delete for an ID through the same critical section, use a generation/CAS or tombstone, or make the store operation itself conditional on a current generation.

## Acceptance criteria

- A touch that began before DELETE cannot recreate the persisted session afterward.
- Expiry cleanup cannot be undone by an older touch write.
- Deterministic tests use a controllable persistence store to force delete-before-delayed-upsert ordering and verify restart restoration does not revive the ID.
- Normal touch persistence remains race-safe under concurrent requests.

yidakra:
Closes #696. Closes #694.

Both defects live in `internal/mcp` and both were verified against the code before anything was written. They ship together because they overlap: `listFilesFiltered` and the session lifetime path sit in the same package, and #694 reshapes the very store call that #696's test fixture wraps. As separate PRs they would have conflicted.

---

# #696 — a concurrent touch could resurrect a deleted session

This is the serious one. A session that was explicitly terminated coming back to life defeats whatever the termination was for.

## The interleaving

Session mutation was ordered **only in memory**. Every mutation released `sessionMu` and *then* wrote to the store, so nothing tied the order of the in-memory transitions to the order of the store writes.

Thread A — any authenticated request:

1. `hasActiveSession(id)` takes `sessionMu`, finds the session live, sets `si.lastSeen = now`, writes it back into `s.sessions`, and **releases `sessionMu`**.
2. A is descheduled before/inside `persistSession` → `store.UpsertMCPSession(id, …)`.

Thread B — `DELETE` for the same id:

3. `forgetSession(id)` takes `sessionMu`, does `delete(s.sessions, id)`, and **releases `sessionMu`**.
4. `deletePersistedSession(id)` → `store.DeleteMCPSession(id)` **completes**. The row is gone.

Thread A resumes:

5. A's `UpsertMCPSession` executes and **re-inserts the row**, carrying the `lastSeen` it captured back at step 1.

The in-memory order (touch, then delete) was correct. Only the **store** order got inverted (delete, then touch). And the damage is invisible while the process lives: memory has no entry, so the id is still rejected. The next restart is where it surfaces — `restoreSessions` reads the resurrected row, finds it inside the inactivity/max-life window, and loads it as **active**. The terminated session is valid again, for a freshly refreshed timeout.

The same inversion undoes expiry: `hasActiveSession`'s two expiry branches and `cleanupExpiredSessions` all delete-after-unlock too.

## The fix

A generation token per transition, allocated **inside the same critical section that performs the in-memory transition**, so token order equals transition order:

- `reserveSessionPersistLocked(id)` is called with `sessionMu` held, at all five mutation sites (touch, create, forget, both expiry branches, and the sweep).
- `commitSessionPersist(id, seq, apply)` drops the write when a newer token already reached the store, and **holds `sessionPersistMu` across the store call**.

Both halves are load-bearing, and the second is the non-obvious one. A check that does *not* hold the lock across the I/O does **not** fix this: the stale touch would pass its check, the DELETE would pass its own check and complete, and the touch would still land last. Serializing across the store call is what makes the store's final state agree with the newest logical transition.

`sessionPersistMu` is deliberately **not** `sessionMu`: `sessionMu` is taken on every request and must not be held across store I/O. The per-id seq record is dropped once its in-flight count hits zero — it cannot go earlier (it is exactly what a late write is rejected against) and must not go never (the map would otherwise grow for the life of the daemon).

## Tests — `tests/mcp/session_resurrection_696_test.go`

The interleaving is **forced through the persistence store, not slept on**. Request A is parked *inside* `UpsertMCPSession`, a point that exists both before and after the fix, so the tests compile and run against unfixed source.

Proven to fail with only `internal/mcp/server.go` stashed:

```
--- FAIL: TestTerminatedSessionIsNotResurrectedByARacingTouch_696 (0.13s)
    a terminated session is still persisted ([V22BQNTVRCK2Y7DUBIYX7L7ZQ5]); an older
    touch overtook the DELETE and a restart will revive it (#696)

--- FAIL: TestExpiredSessionIsNotResurrectedByARacingTouch_696 (0.43s)
    an expired session is still persisted ([OC6ZMRAVZENM6NGTRS57JIWYLF]); an older
    touch write undid the expiry (#696)
```

The DELETE test then goes the whole way: it stands up a **second server over the same store** (so `restoreSessions` runs) and asserts the terminated id now gets a 404.

Two deliberate calls worth flagging for review:

- **The wait for B's delete is asymmetric, and it is not a "sleep and hope".** Against unfixed code nothing serializes B behind A, so B's delete lands promptly and the signal always fires: the resurrection is *forced*, not raced — note the 0.13s/0.43s pre-fix runtimes. Against fixed code B is parked behind the very serialization the fix adds, so the signal cannot arrive until A is released, and the timeout is simply the test observing the fix work.
- **The expiry test asserts on the persisted row rather than on a restart.** `restoreSessions` re-applies the expiry rules it was configured with, so a resurrected *expired* row gets filtered out again on load and a restart-based assertion would have passed vacuously against broken code. The durable defect is the row surviving the expiry at all.

The third test is a **guard, not a reproduction**: it passes before and after by design, and exists so the new ordering machinery cannot silently break ordinary concurrent touches. `tests/mcp` is added to `make test-race` so that guard keeps running under `-race`, which is the only way it stays meaningful.

---

# #694 — every `list_files` page rescanned and stat'ed the whole corpus

`listFilesFiltered` walked the entire matching corpus from store offset zero **on every call**, in 500-row pages, running `filepath.EvalSymlinks` on every row, purely to compute a post-filter `total`. #429 F10's store-level pagination was fully undone at the tool boundary.

## The fix

Same shape as #732/#743: **push the selective predicate down to where the LIMIT already is.**

`isListFilesNoisePath` is exactly "the normalized rel_path starts with a dot" — it takes the first `/`-delimited segment and tests `HasPrefix(first, ".")`, and testing the first segment's prefix *is* testing the whole string's prefix. So it becomes `rel_path NOT LIKE '.%'` in the `where` slice, which is already threaded into the page query, `countListFiles`, **and** the glob scan. All three start agreeing on the visible set for free.

- `SQLiteStore.ListVisibleFiles` is the new entry point; `ListFiles` is a thin delegate, so `model.Store` and its ~30 callers are untouched.
- The MCP handler asks the store for **one page** through an optional `visibleFilesLister` capability (the `recentFailuresLister` idiom), then runs the #176 round-trip gate over **the rows of that page**.
- The pre-#694 walk is kept as the fallback for a store that cannot push the predicate down — and the perf harness reuses it as its honest "before" arm rather than reimplementing it.

The **memo key now carries the visibility policy**. This was the real trap: #429 F10 caches the exact total per listing, and two visibility policies over the same prefix/glob are two different listings. Without it a hidden-inclusive page gets served a hidden-excluded total, and a client paging on `offset < total` silently stops early.

## Measured — 20,000 documents (18,000 visible, 10% hidden), real files on disk

| | store reads | rows stat'ed | elapsed |
|---|---|---|---|
| first page (`limit=50, offset=0`) **before** | 40 | 20,000 | 3.304s |
| first page **after** | **1** | **50** | **33ms** |
| deep page (`offset=17,900`) **before** | 40 | 20,000 | 1.750s |
| deep page **after** | **1** | **50** | **59ms** |
| walking 10 pages **before** | 400 | 200,000 | 35.373s |
| walking 10 pages **after** | **10** | **500** | **339ms** |

Roughly 100x on a first page, and the quadratic term is gone: ten pages cost ten reads instead of 400. Rows-returned *is* the `EvalSymlinks` syscall count, which is the term that dominates on a network filesystem.

The harness is opt-in, following #742/#748, so the default suite stays fast:

```
DIR2MCP_PERF=1 go test ./tests/mcp -run ListFilesPagePerf -v
```

## What changed about `total` — stated plainly

`total` now means *"matching, non-deleted, non-hidden rows in the store"* rather than *"rows that also survived a filesystem stat"*.

Those differ **only** when the store holds a row whose backing file is gone — i.e. when ingest's `deleted = 0` tombstoning has drifted, which is the degraded case the stat filter exists as a safety net for. In a healthy corpus they are identical, and `documents.deleted` is already the authoritative liveness signal `ListFiles` filters on in SQL. Paying O(corpus) stats on every request to keep the total exact under drift **is the defect being fixed**, so the trade is deliberate rather than incidental.

Visible consequence: a page can come back shorter than `limit` when a dead row is dropped from it. That was already possible at the end of a listing, and a client looping `while offset < total` still terminates and still sees every live file. What does **not** change is #176's actual guarantee: every path the tool *emits* round-trips through `open_file`, because the page-scoped gate still covers every emitted row.

## Acceptance criteria not fully met

Two of the issue's criteria are only partially addressed, and I would rather say so than imply otherwise:

- **"Exact visible total"** is exact for a healthy corpus but is an upper bound under tombstoning drift, per the trade above. Making it exact *and* cheap needs ingest-time reconciliation so liveness becomes a SQL predicate — the issue's own first-choice option, and a considerably larger change than this.
- **Single-snapshot consistency** is much improved but not literally satisfied. One response is now one store call, so the page rows all come from a single query and the multi-call state-mixture the issue describes is gone. The total can still come from a separate `COUNT(*)` (or the #429 memo), so page and total can describe adjacent commits — already true before this PR, and documented on `listFilesSQLPaged`. There is no `BEGIN`-wrapped snapshot and no keyset/cursor API.
- **"Concurrent insert/delete tests proving no duplicates/skips within one response"** is not written, because with one page query there is no intra-response window left to test. A meaningful test would have to target the page/COUNT gap, which is pre-existing #429 territory and belongs in `tests/store`. Flagging it rather than silently dropping it.

## Tests

All of them drive the **real tool over HTTP** and **write real files on disk** — per PR #766's warning, `list_files` drops rows whose file does not exist, so a store-only fixture returns zero files and would make every assertion pass vacuously.

**Reproductions**, proven to fail first. With the handler pointed back at `listFilesFilteredByWalk`:

```
--- FAIL: TestListFilesServesASmallFirstPageInOneStoreRead_694
    a 5-row first page cost 3 store reads (ListFiles=3 ListVisibleFiles=0); the
    handler is still walking the whole corpus to compute a total (#694)

--- FAIL: TestListFilesDeepPageDoesNotRestartFromZero_694
    offset=1000 cost 3 store reads (ListFiles=3 ListVisibleFiles=0); a later page
    must be served by the store's own LIMIT/OFFSET, not by re-walking every
    preceding row (#694)
```

And with only the visibility component removed from the memo keys (SQL predicate left in place), which is how the cache-key trap was confirmed to be real on **both** paths:

```
--- FAIL: .../sql_path/visible_then_all
    second call returned 5 rows but total=3 (want 5): the memoized total of the
    previous visibility policy was reused, so the page and its total describe
    different listings (#694)
--- FAIL: .../glob_path/visible_then_all
```

**Guards, not reproductions** — flagged rather than dressed up as failures. `TestListFilesHiddenFilteringSurvivesTheSQLPushdown_694` and `TestListFilesStillDropsRowsWhoseFileIsGone_694` pass before and after by construction: they assert behavior that must stay *unchanged* across the pushdown. The first pins that `visible/.config/b.md` is **not** hidden (only the first segment counts, which is exactly what a naive `LIKE '%/.%'` translation breaks); the second is the #176 gate. `TestListFilesStaysHiddenInclusive_694` likewise pins that the `ListFiles` delegate stays hidden-inclusive for its existing callers.

---

`make check` passes. `make test-race` passes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,8 +79,11 @@ test:
 # race detector so concurrency regressions (issue #419) are caught. It needs the
 # C toolchain (CGO_ENABLED=1); it is a separate target rather than folded into
 # `check` because `check`/`ci` run with CGO_ENABLED=0.
+# tests/mcp joins the list for #696: session lifetime transitions are ordered
+# across goroutines now, and the regression test for that ordering is only a
+# durable guard if it keeps running under the race detector.
 test-race:
-	CGO_ENABLED=1 go test -race ./internal/cli/... ./internal/index/... ./tests/cli ./tests/index ./tests/store
+	CGO_ENABLED=1 go test -race ./internal/cli/... ./internal/index/... ./tests/cli ./tests/index ./tests/mcp ./tests/store
 
 test-release-tools:
 	cd scripts && python3 -m unittest discover -p 'test_*.py'
--- a/internal/mcp/server.go
+++ b/internal/mcp/server.go
@@ -97,6 +97,25 @@ type Server struct {
 	// optional absolute lifetimes.
 	sessions map[string]sessionInfo
 
+	// sessionSeq orders session state transitions. It is allocated under
+	// sessionMu (see reserveSessionPersistLocked) so a token's order matches the
+	// order of the in-memory transition it describes, which is what lets the
+	// persistence layer tell an older write from a newer one (#696).
+	sessionSeq uint64
+
+	// sessionPersistMu serializes the store writes that mirror those in-memory
+	// transitions, and guards the two maps below. It is deliberately NOT
+	// sessionMu: sessionMu is taken on every request and must not be held across
+	// store I/O.
+	sessionPersistMu sync.Mutex
+	// sessionPersistSeq records, per session ID, the newest transition that has
+	// reached the store. A write carrying an older token is dropped.
+	sessionPersistSeq map[string]uint64
+	// sessionPersistInflight counts the reserved-but-not-yet-committed
+	// transitions per session ID, so the seq record can be dropped once nothing
+	// can still arrive late for that ID.
+	sessionPersistInflight map[string]int
+
 	rateLimiter *ipRateLimiter
 
 	x402Client      x402.FacilitatorClient
@@ -288,6 +307,10 @@ func NewServer(cfg config.Config, retriever model.Retriever, opts ...ServerOptio
 		retriever:       retriever,
 		sessions:        make(map[string]sessionInfo),
 		paymentOutcomes: make(map[string]paymentExecutionOutcome),
+
+		sessionPersistSeq:      make(map[string]uint64),
+		sessionPersistInflight: make(map[string]int),
+
 		paymentTTL:      paymentOutcomeTTL,
 		paymentMaxItems: paymentOutcomeMaxEntries,
 		nonceLedger:     make(map[string]nonceLedgerEntry),
@@ -642,15 +665,17 @@ func (s *Server) hasActiveSession(id string, now time.Time) (bool, string) {
 	}
 	if now.Sub(si.lastSeen) > inactivity {
 		delete(s.sessions, id)
+		seq := s.reserveSessionPersistLocked(id)
 		s.sessionMu.Unlock()
-		s.deletePersistedSession(id)
+		s.deletePersistedSession(id, seq)
 		log.Printf("session %s expired due to inactivity", maskSessionID(id))
 		return false, "inactivity"
 	}
 	if maxLife > 0 && now.Sub(si.created) > maxLife {
 		delete(s.sessions, id)
+		seq := s.reserveSessionPersistLocked(id)
 		s.sessionMu.Unlock()
-		s.deletePersistedSession(id)
+		s.deletePersistedSession(id, seq)
 		log.Printf("session %s expired due to max lifetime", maskSessionID(id))
 		return false, "max-lifetime"
 	}
@@ -658,8 +683,9 @@ func (s *Server) hasActiveSession(id string, now time.Time) (bool, string) {
 	// update lastSeen
 	si.lastSeen = now
 	s.sessions[id] = si
+	seq := s.reserveSessionPersistLocked(id)
 	s.sessionMu.Unlock()
-	s.persistSession(id, si)
+	s.persistSession(id, si, seq)
 	return true, ""
 }
 
@@ -668,8 +694,9 @@ func (s *Server) storeSession(id string, authScope string) {
 	si := sessionInfo{created: now, lastSeen: now, authScope: authScope}
 	s.sessionMu.Lock()
 	s.sessions[id] = si
+	seq := s.reserveSessionPersistLocked(id)
 	s.sessionMu.Unlock()
-	s.persistSession(id, si)
+	s.persistSession(id, si, seq)
 }
 
 // forgetSession removes a session from the in-memory map and the persistence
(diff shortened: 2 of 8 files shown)


yidakra:
@coderabbitai review

PR #776, opened 2026-08-05
fix(embedqueue): give distributed embedding jobs a corpus and a payload identity (#708, #709, #710)

### LINKED ISSUE #708: distributed embed: shared brokers can collide and execute jobs across corpora
## Summary

Distributed embedding jobs carry `corpus_id` and `source`, but the queue and worker treat those fields as inert metadata. Two corpora using the same broker and overlapping chunk IDs can suppress, steal, or execute each other’s jobs.

This violates SPEC §8.7.2, where the corpus reference is required so a worker can identify the exact corpus/source for the job, and §6.3/§8.7.4, where shared Tier-C infrastructure must remain corpus-isolated.

## Evidence

- `embedqueue.Job` declares `CorpusID` and `Source`, explicitly saying they scope the corpus and let a worker reject a job for a corpus it does not serve.
- `embedqueue.Config` has no expected corpus/source fields, and `prepare` never checks `job.CorpusID` or `job.Source`.
- Both `MemBroker.Enqueue` and `SQLiteBroker.Enqueue` deduplicate only by `(chunk_id, index_kind)`, omitting `corpus_id`.
- `startDistributedEmbedding` fills `Job.CorpusID` with `rootDir`, not the stable persisted `settings.corpus_id` required by SPEC §5.5.
- A worker always fetches `ChunkTaskByID(job.ChunkID)` from its locally configured metadata store and writes to its locally configured Tier-C namespace, regardless of which corpus produced the leased job.

## Reproduction

1. Configure corpora A and B with the same SQLite broker path and the same embed identity.
2. Give each corpus a pending text chunk with `chunk_id=1`.
3. Enqueue A’s job, then enqueue B’s job. B’s enqueue is suppressed because a live `(1,text)` job already exists.
4. A worker configured for corpus B can lease A’s job, fetch B’s local chunk 1, embed/write it into B, and Ack/delete A’s job.

Depending on timing, A loses work, B executes work attributed to A, and queue depth appears healthy.

## Expected behavior

- Resolve jobs against the persisted stable corpus ID, not a root path string.
- Namespace broker deduplication and leasing by corpus ID, or require workers to reject/Nack jobs whose corpus/source binding differs from their configured binding.
- Ensure a shared broker can safely carry multiple corpora with overlapping SQLite chunk IDs.
- Add two-corpus tests covering enqueue dedup, worker rejection/routing, and identical embed identities.

### LINKED ISSUE #709: distributed embed: dead-lettered preparation failures are re-enqueued forever
## Summary

The distributed embed retry limit is not terminal for failures that occur before `EmbedAndIndex`. Identity mismatches, missing axis embedders, and persistent metadata fetch failures eventually move a broker row to `dead`, but the chunk remains `embedding_status=pending`. The coordinator then creates a brand-new job with a reset attempt counter, repeating forever.

With the SQLite broker this also grows the durable `embed_jobs` table without bound because every exhausted cycle leaves another `state=dead` row.

## Evidence

- `embedqueue.Config.prepare` Nacks identity mismatches, unknown `index_kind`, and non-not-found fetch errors without updating chunk status.
- `MemBroker` and `SQLiteBroker` dead-letter the job at `maxAttempts`, but expose that only through `Stats.DeadLettered`.
- No production CLI/coordinator path calls `Broker.Stats` or reconciles a dead letter back to the chunk/document status.
- `Coordinator.EnqueuePending` continues selecting the still-pending chunk.
- Both broker `Enqueue` implementations deduplicate only against `pending`/`inflight`; dead jobs deliberately do not block a fresh enqueue.

The existing `TestWorker_EmbedIdentityMismatchRejected` stops as soon as the first dead letter appears. It does not run a coordinator concurrently, so it misses the immediate fresh enqueue/reset loop.

## Reproduction

1. Run a coordinator with a pending chunk and a worker whose embed identity does not match the job.
2. The worker Nacks until `maxAttempts`, producing one dead letter.
3. The chunk remains pending.
4. On the next coordinator tick, another job is inserted for the same chunk and the retry budget starts again.
5. Repeat indefinitely; SQLite accumulates dead rows while indexing never converges.

## Impact

A persistent configuration mismatch produces endless broker churn, logs, and retries rather than the specified bounded failure. Queue dead-letter counts grow, the chunk remains permanently pending, and status does not surface the terminal cause.

## Expected behavior

Dead-lettering must transition the corresponding chunk to a durable failed state with an actionable error/category, or the coordinator must suppress fresh jobs until an explicit retry/reset action. Production status/doctor output should surface dead letters. Add an integration test running coordinator and worker together past `maxAttempts` and prove the job is not recreated.

### LINKED ISSUE #710: distributed embed: jobs omit payload identity and can route a changed chunk to the stale axis
## Summary

Distributed jobs never carry the required text payload identity, and workers discard the authoritative hash returned by the store. A queued job can therefore outlive an in-place chunk update and route the updated task using stale job metadata—most critically, to the wrong text/code index axis.

This contradicts SPEC §8.7.2, which requires `text_hash` so workers can identify the exact payload a job names.

## Evidence

- `Coordinator.jobFromTask` deliberately leaves `Job.TextHash` empty, saying it is not load-bearing.
- `TaskFetcher.ChunkTaskByID` returns the current `textHash` separately, but `worker.prepare` discards it (`task, _, err := ...`).
- The worker chooses the embedder and `indexKind` from the leased job before fetching the authoritative task; it never compares that routing value with `task.IndexKind`.
- SQLite chunk upsert preserves `chunk_id` for the same `(rep_id, ordinal)` while updating `text_hash`, `index_kind`, text, modality, and embedding status in place.

## Reproduction race

1. Enqueue chunk 42 while it is a `code` chunk.
2. Before a worker leases it, re-ingest the same representation/ordinal as normal text. SQLite keeps chunk ID 42 but updates it to `index_kind=text`, a new hash, and `pending`.
3. Lease the old job. The worker fetches the new authoritative task but routes it through the old job’s `code` embedder and code index.
4. `EmbedAndIndex` marks chunk 42 embedded. If this happens before the coordinator enqueues a new text job, the correct text-axis job is never created; a stale code-axis vector remains searchable.

Even when the axis does not change, the broker payload falsely claims to identify exact work while carrying no hash and cannot distinguish superseded deliveries.

## Expected behavior

- Populate every text job with the persisted chunk `text_hash` required by the spec.
- On lease, compare the job hash and routing metadata with the authoritative task/hash.
- Ack a superseded job as stale without embedding it; let the coordinator enqueue the current pending form.
- Never mark a chunk embedded through an axis that differs from its current `task.IndexKind`.
- Add a deterministic test that updates a stable chunk ID between enqueue and lease.

yidakra:
Fixes #708. Fixes #709. Fixes #710.

All three were verified against the code before anything was changed; all three
are real and accurately described. They share one root cause, which is why they
are one PR: **a job did not carry enough about WHICH corpus and WHICH payload it
named**, so a broker several corpora share could cross wires, and a job that
could never succeed was re-created forever. Split apart, the three fixes touch
the same `prepare` path, the same `Job`, and both brokers.

### #708 — shared brokers can collide and execute jobs across corpora

| Change | Effect |
|---|---|
| `Coordinator.CorpusID` is the stable persisted corpus id, not `rootDir` | jobs name a corpus that survives a move and is unique for an S3 corpus |
| `MemBroker`/`SQLiteBroker` dedup on `corpus_id+chunk_id+index_kind` | corpus B's job is no longer swallowed by corpus A's live job for the same id |
| new optional `CorpusScopedBroker` (`LeaseForCorpus`), implemented by both built-ins | a worker claims only its own corpus's jobs |
| `Config.CorpusID` + `servesCorpus` check, first, before the store is touched | a foreign job is returned unexecuted and never marks a chunk in the wrong store |

### #709 — dead-lettered preparation failures are re-enqueued

| Change | Effect |
|---|---|
| `Lease.Attempts` / `Lease.MaxAttempts` / `Lease.Final()` | the worker can tell a Nack that redelivers from one that dead-letters |
| new optional `Config.Status` (`MarkFailedWithCategory`), wired from the chunk source in both CLI paths | the final delivery records `embedding_status=error` + category |
| all reject paths route through `rejectJob` | applies to identity mismatch, unserved axis, persistent fetch failure, and embed failure alike |

### #710 — jobs omit payload identity and can route a changed payload to a stale consumer

| Change | Effect |
|---|---|
| `model.ChunkTask.TextHash`, populated by `NextPending` and `ChunkTaskByID` | the coordinator has a hash to stamp |
| `Job.TextHash` is populated | the job names its payload (SPEC §8.7.2) |
| worker keeps the store's hash and compares it, and routes by `task.IndexKind` | a superseded job is acked without embedding; a chunk is never marked embedded through an axis it is not on |

## Design decisions

**What identifies a corpus.** `identity.CorpusKey` / `CorpusKeyForS3` are the
key derivations `AutoServerName` / `AutoServerNameForS3` (#737) already used,
factored out so there is exactly one notion of "which corpus is this" and both
consumers share it — `AutoServerNameForS3` now calls `CorpusKeyForS3`, and its
existing tests pass unchanged. `identity.CorpusID` hashes that key to
`corpus-<32 hex>` and `ResolveCorpusID` persists it in `settings.corpus_id`
(SPEC §5.5, previously specified but never written).

Three deliberate properties:

- **Opaque, not human-readable.** Unlike the instance name the same key
  produces, the corpus id carries no path, bucket, endpoint or credential. It
  lands in a queue that by design several corpora may share, which on a
  multi-tenant host is the one place corpus A can read rows corpus B wrote. A
  digest identifies a corpus without describing it; an operator maps it back
  from that corpus's own store. Nothing credential-bearing reaches a broker, a
  queue name, or a log — `servesCorpus` logs the digest only.
- **Persisted value wins.** Deriving on every start would rename a corpus that
  was moved or re-mounted, orphaning its queued jobs and its written vectors.
- **No dev/release prefix.** `AutoServerName` carries one; a persisted corpus
  identity that changed when a dev build opened the same corpus would split one
  corpus in two.

**What "dead" means (#709).** Dead-lettering only ends the *job*. The chunk
stayed `pending`, so the coordinator's next tick minted a new job with a new
budget. "Dead" now means the **chunk** is terminal: `embedding_status=error`
with a category. That is what breaks the loop, because `NextPending` selects
only pending chunks — and it is also how an operator sees it, in the existing
`dir2mcp status` / `doctor` error-category breakdown and samples, in the same
shape an in-process failure appears. No new surface was invented for it.

**A job that legitimately deserves a retry** is untouched: the terminal write
fires only on the delivery that exhausts the broker's budget, so a transient
fault that clears inside the budget is redelivered and succeeds with nothing
recorded (pinned by a test). An operator who fixes the cause re-pends the chunk
with `dir2mcp reindex`. Dead rows are now bounded at ~one per permanently-failing
chunk rather than growing without limit.

**Payload-shape identity (#710): `text_hash`, not `EmbedIdentity`, and no new
schema version.** `EmbedIdentity` keys the *embedding space*; it is already
carried and already enforced, and it does not change when a chunk's bytes
change, so it cannot detect a superseded payload. SPEC §8.7.2 names `text_hash`
as the payload identity and that is what is now carried. A separate
payload-*schema* version was considered and rejected: the one problem it would
solve is a durable SQLite queue holding jobs written by an older binary, and
that is handled precisely by treating an empty hash as **unknown** rather than
as a mismatch — the comparison can only prove staleness, never freshness. A
version field would add a migration and a compatibility matrix to buy nothing
the one-sided comparison does not already give. There is a test for the
legacy-job case.

## Also fixed (found by the new tests)

`NewSQLiteBroker` set `busy_timeout` with a single `ExecContext`, which lands on
one pooled connection. With two workers on one queue file, `database/sql` opens
further connections that never got it, so their reclaim-write-first `Lease`
failed *immediately* with SQLITE_BUSY and a failed `Ack` stranded a leased job
until its 30s lease expired. It now travels in the DSN with
`SetMaxOpenConns(1)`, the fix and the reasoning `internal/store` already
documents for the same defect (#429 F11). This is the shape #708 is about — a
shared broker with concurrent workers — and the existing white-box pragma test
still passes.

## Testing — proven to fail first

Each test was run against the tree with **only its source change reverted**
(behaviour removed, API kept, so it still compiles):

**#708** — the cross-corpus test is a real collision, not a mocked assertion:
two real SQLite metadata stores each independently allocating chunk id 1, two
real corpus ids resolved through the real persist path, one real shared
`SQLiteBroker` file, and the real worker loop.

- reverting the corpus term in the dedup predicate:
  `shared queue holds 1 pending job(s), want 2 ... a corpus whose chunk id
  collides with another's has silently lost its job`
- reverting corpus-scoped leasing + the worker's binding check:
  `corpus beta's worker embedded [beta chunk]; it must not execute another
  corpus's job` — the full original bug, with corpus A's job consumed and acked
  by a worker that embedded corpus B's bytes.

**#709** — coordinator and worker run **together**, past the retry limit, over
two consecutive identical windows. The measurement is whether the dead-letter
count keeps climbing with wall-clock time, which is what distinguishes
convergence from the loop (a single snapshot cannot: one redundant job is
legitimately possible from a coordinator read that races the terminal write).
Reverting the terminal write:

- `dead letters grew from 94 to 187 over a second identical window`
- unserved-axis variant: `grew from 86 to 177`

With the fix both settle and stop. A third test pins that a transient failure
still gets its full retry budget and is recorded nowhere.

**#710** — deterministic, no concurrency: enqueue while the chunk is `code`,
re-ingest the same `(rep_id, ordinal)` as `text` (the store preserves the chunk
id and rewrites hash, axis and status), then lease.

- with `Job.TextHash` left empty: `queued job carries text_hash "", want the
  chunk's persisted hash`
- with routing taken from the job: `the stale code-axis job embedded [just prose
  now]; the chunk is a TEXT chunk now and a code vector for it would stay
  searchable on the wrong axis`
- same-axis hash change: `a job enqueued for text_hash hash-v1 embedded [second
  revision]`

Plus guards against over-rejection: an unchanged chunk still embeds, and a job
with no payload identity still executes.

## Checklist

- `make check` green (`gocyclo -over 15` clean; `golangci-lint` 0 issues)
- `go test -race ./tests/embedqueue ./tests/identity -count=4` green
- No new files under `internal/cli/`, so `tests/security/cli_boundary_test.go`
  needs no change (verified — it passes)
- Tests live under `tests/embedqueue/` and `tests/identity/`
- `docs/dual-machine-deployment.md` §4.4/§4.5 document the corpus-isolation and
  terminal-failure behaviour

### DIFF
--- a/docs/dual-machine-deployment.md
+++ b/docs/dual-machine-deployment.md
@@ -380,6 +380,47 @@ Run as many `embed-worker` processes as your GPU(s) can feed. Because all worker
 write to the same Tier-C collection under the same embed identity, the result is a
 single coherent index.
 
+### 4.4 Sharing one broker between corpora
+
+Two corpora may point at the same `sqlite_path`. Each corpus is identified by a
+stable `corpus_id` (SPEC §5.5), derived on first use from the corpus root (or,
+for `source.kind: s3`, from bucket + prefix + endpoint) and persisted in that
+corpus's own metadata store. It is an opaque digest: it carries no path, bucket,
+endpoint or credential, so a queue file readable by both corpora discloses
+neither corpus's layout to the other.
+
+Every job carries its `corpus_id`, and it governs routing end to end:
+
+- the queue deduplicates jobs per corpus, so two corpora whose chunk ids collide
+  (they are per-corpus SQLite rowids, so this is normal, not exotic) each keep
+  their own job;
+- a worker leases only jobs for the corpus its metadata store belongs to, and
+  refuses to execute one for any other corpus even if a broker hands it over.
+
+Moving or re-mounting a corpus does not change its `corpus_id` — the persisted
+value wins over the derived one, so in-flight jobs and already-written vectors
+stay bound to the corpus they came from.
+
+### 4.5 When a job cannot succeed
+
+A job is redelivered up to `max_attempts` times and then dead-lettered. On that
+final attempt the chunk itself is recorded as `embedding_status=error` with a
+category, so:
+
+- it leaves the pending set and the coordinator stops minting new jobs for it —
+  a permanent misconfiguration (mismatched embed identity, an index axis no
+  worker serves) fails **bounded** instead of retrying forever;
+- it appears in `dir2mcp status` and `dir2mcp doctor` under the error-category
+  breakdown, alongside in-process embedding failures.
+
+A failure that clears within the retry budget is simply redelivered and succeeds;
+nothing is recorded. To retry a chunk that was recorded as failed — after fixing
+the cause — re-pend it with `dir2mcp reindex`.
+
+A job whose chunk changed after it was enqueued (re-ingested with different text,
+or moved to the other index axis) is acknowledged as **superseded** without being
+embedded, and the coordinator enqueues the chunk's current form on its next pass.
+
 ---
 
 ## 5. Caveats and current limitations
--- a/internal/cli/embed_worker.go
+++ b/internal/cli/embed_worker.go
@@ -13,6 +13,7 @@ import (
 	"github.com/dirstral/dir2mcp/internal/config"
 	"github.com/dirstral/dir2mcp/internal/embedqueue"
 	"github.com/dirstral/dir2mcp/internal/index"
+	"github.com/dirstral/dir2mcp/internal/model"
 )
 
 // embedWorkerOptions holds the parsed flags for the standalone embed-worker
@@ -133,6 +134,27 @@ func (a *App) requireDistributedPrereqs(cfg config.Config, global globalOptions)
 	return exitSuccess
 }
 
+// requireSharedStoreCapabilities asserts the two optional store capabilities a
+// distributed worker cannot run without: read-by-id (ChunkTaskByID) so it can
+// load the authoritative chunk for a leased job (§8.7.4), and the chunk source
+// so EmbedAndIndex — and the terminal-failure path (#709) — can write status
+// back. Both are satisfied by the SQLite-backed metadata store.
+func (a *App) requireSharedStoreCapabilities(st model.Store, global globalOptions) (embedqueue.TaskFetcher, index.ChunkSource, int) {
+	fetcher, ok := st.(embedqueue.TaskFetcher)
+	if !ok {
+		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid,
+			"CONFIG_INVALID: metadata store does not support ChunkTaskByID (required for distributed embedding)")
+		return nil, nil, exitConfigInvalid
+	}
+	chunkSource, ok := st.(index.ChunkSource)
+	if !ok {
+		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid,
+			"CONFIG_INVALID: metadata store does not support the chunk source interface (required for distributed embedding)")
+		return nil, nil, exitConfigInvalid
+	}
+	return fetcher, chunkSource, exitSuccess
+}
+
 // runEmbedWorkerLoop opens the shared store, the Tier-C indices, the corpus
 // filesystem and the broker, builds per-axis embedders, and runs embedqueue.Run
 // until the context is cancelled. Setup failures are remediable config errors;
@@ -158,21 +180,9 @@ func (a *App) runEmbedWorkerLoop(ctx context.Context, cfg config.Config, global 
 		}
 	}()
 
-	// The shared store must expose read-by-id (ChunkTaskByID) so the worker can
-	// load the authoritative chunk for a leased job (§8.7.4), and the chunk
-	// source so EmbedAndIndex can write status back. Both are satisfied by the
-	// SQLite-backed metadata store.
-	fetcher, ok := st.(embedqueue.TaskFetcher)
-	if !ok {
-		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid,
-			"CONFIG_INVALID: metadata store does not support ChunkTaskByID (required for distributed embedding)")
-		return exitConfigInvalid
-	}
-	chunkSource, ok := st.(index.ChunkSource)
-	if !ok {
-		writeCLIError(a.stderr, global.jsonOutput, exitConfigInvalid,
-			"CONFIG_INVALID: metadata store does not support the chunk source interface (required for distributed embedding)")
-		return exitConfigInvalid
+	fetcher, chunkSource, code := a.requireSharedStoreCapabilities(st, global)
+	if code != exitSuccess {
+		return code
 	}
 
 	embedder, _, textModel, codeModel := a.resolveModelClients(cfg)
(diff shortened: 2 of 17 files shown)


yidakra:
@coderabbitai review

PR #779, opened 2026-08-06
fix(scorebug): make the count-derived pitches opt-in (they are currently on by default)

yidakra:
**This changes current behaviour on `main`, and that is the point.**

The count reader already shipped, ungated. It reached `main` inside #780, and today:

```
$ ScorebugRecognizer(roster, ocr=ocr, crop=WHOLE).recognize(media)   # no flag
   -> 2 count-derived pitch cues
```

So every corpus indexed with `--scorebug` right now is getting count-derived pitches whether the operator wanted them or not. This PR puts them behind a flag and turns them **off** by default.

## Why off

The speed graphic is shown for some pitches and not others: on the pilot game it yielded 299 cues against 344 thrown. The pitch count beside the pitcher's name is on the bug continuously and rises by one per pitch, so each `+1` is a pitch with a timestamp, already attributed. Measured against the pilot's ground truth:

| | recall | precision |
|---|---|---|
| graphic only | 86.9% | **100.0%** |
| with count transitions (**today's default**) | **90.1%** | 94.2% |

Not a strict improvement. `+1` is both what a real pitch looks like and what the most likely OCR misread looks like. Requiring a new count to hold for `COUNT_STABLE_READS` frames removes the 87→88→87→88 flapping but cannot remove a misread that holds.

For a product whose output is **citations**, a cited moment that did not happen costs more than a moment never found. That is a judgement about a deployment, not a fact about the code, so it belongs to the operator:

```
dirstral-annotate serve --roster roster.json --scorebug --scorebug-pitch-counts
```

## What changed

The count reader rode on `pitch_cues`, which also gates the pre-existing graphic-derived pitches, so there was no way to keep one and drop the other. The two channels are now independent (`count_pitch_cues`, surfaced as `--scorebug-pitch-counts`).

Enabling counts **adds** cues and leaves the graphic path unchanged, pinned by a test that compares the two runs directly rather than asserting it in a comment.

## Tests

- the default is off (this is the behaviour change; it fails against `main`)
- enabling counts leaves the graphic-derived cues identical
- inert without `--scorebug`, and inert when `pitch_cues=False` disables the pitch channel
- the flag reaches the recognizer from `argv` (a flag parsed but never threaded is the usual way an option silently does nothing)
- the original count-reader behaviour, now opted in: a step is a pitch, a backwards count resets rather than emits, a real step survives a neighbouring misread

276 annotator tests pass. Note that **CI does not run them** (#787), which is part of how an ungated precision regression shipped unnoticed.

## Note on how this happened

#779 was opened as the count reader and I recommended against merging it on these numbers. The feature nevertheless reached `main` as part of #780, whose branch carried the commit. Rebasing this PR onto `main` dropped its first commit as "patch contents already upstream", which is how it surfaced.

### DIFF
--- a/annotator/dirstral_annotator/cli.py
+++ b/annotator/dirstral_annotator/cli.py
@@ -6,6 +6,10 @@
         [--scorebug] [--jersey] [--faces bank/] [--news] [--ocr-lang rus] \
         [--fps 0.5] [--min-confidence 0.3] [--host 127.0.0.1] [--port 8765]
 
+    # Scorebug pitch counts: more pitches found, some of them invented.
+    # Opt-in, because it spends precision (see ScorebugRecognizer).
+    dirstral-annotate serve --roster roster.json --scorebug --scorebug-pitch-counts
+
     # A news archive resolves no roster, so it needs none:
     dirstral-annotate serve --news --ocr-lang rus
 
@@ -45,6 +49,9 @@ def build_parser() -> argparse.ArgumentParser:
 
     def vision_flags(c):
         c.add_argument("--scorebug", action="store_true", help="enable scorebug OCR")
+        c.add_argument("--scorebug-pitch-counts", action="store_true",
+                       help="also read a pitch from each +1 of the bug's pitch count "
+                            "(more recall, less precision; needs --scorebug)")
         c.add_argument("--jersey", action="store_true", help="enable jersey-number OCR")
         c.add_argument("--faces", type=Path, help="roster image bank dir; enables face recognition")
         c.add_argument("--news", action="store_true",
@@ -100,6 +107,7 @@ def _pipeline(args, roster: Roster, games) -> Pipeline:
         roster=roster,
         games=games,
         scorebug=args.scorebug,
+        scorebug_pitch_counts=args.scorebug_pitch_counts,
         jersey=args.jersey,
         news=args.news,
         ocr_lang=args.ocr_lang,
--- a/annotator/dirstral_annotator/pipeline.py
+++ b/annotator/dirstral_annotator/pipeline.py
@@ -55,6 +55,10 @@ class Pipeline:
     roster: Roster
     games: dict[str, GameConfig] = field(default_factory=dict)
     scorebug: bool = False
+    #: Opt-in: also read a pitch from each +1 transition of the scorebug's
+    #: pitch count. Trades precision for recall (see ScorebugRecognizer), so
+    #: it stays off unless an operator asks for it. Inert without `scorebug`.
+    scorebug_pitch_counts: bool = False
     jersey: bool = False
     news: bool = False
     ocr_lang: str | None = None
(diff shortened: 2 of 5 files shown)


yidakra:
## Measured, and the first version failed

Full proxy, vision-only, correct roster:

| | baseline | count transitions | target |
|---|---|---|---|
| recall | 86.9% | **91.9%** | ≥ 90% ✓ |
| precision | 100.0% | **92.4%** | ≥ 95% ✗ |

It traded one gate miss for another. The number that matters is not the headline but the conversion: **+52 cues, +17 credited — 35 landed wrong**, a 33% hit rate. That is not a near miss, it is a signal that was mostly wrong.

## The flaw was my guard

I restricted to +1 steps reasoning that larger jumps are noise. But **+1 is exactly the most likely OCR error**: one misread digit gives `87 → 88` (phantom pitch), then `88 → 87` (which my rule treated as a benign "reset"), then `87 → 88` again. A single unstable digit manufactures pitches, and my filter selected for the dominant noise mode instead of rejecting it.

## The fix

A new count must survive two consecutive readings before it is committed. A single-frame flip is discarded; a genuine transition costs nothing, since the count sits on the bug for a whole at-bat — far longer than two frames at 0.5 fps.

The fixtures were weak in the same direction: they showed each count for a single frame, which is not how a scorebug behaves. They are realistic now, plus two cases for the mechanism that actually failed — a single-frame misread must manufacture nothing, and a real step must survive a neighbouring misread.

**Remeasuring now. Still claiming no number.** If precision does not recover, that gets reported here rather than the threshold being tuned until the figure looks right.

yidakra:
## Remeasured with the stability guard

| | baseline | naive counts | **stable counts** | target |
|---|---|---|---|---|
| recall | 86.9% | 91.9% | **90.1%** | ≥ 90% ✓ |
| precision | 100.0% | 92.4% | **94.2%** | ≥ 95% ✗ |

The guard did what it was meant to — precision recovered 92.4% → 94.2% — but it did not close the gate, and it cost 1.8 points of recall doing it.

The conversion rate is the honest number: 38 new cues, 11 credited. **~29%**, barely better than the naive version's 33%. Requiring the count to hold removed some phantom pitches, but the signal as a whole is still mostly wrong when it fires on something the graphic missed.

## Where I would look next, and what I am not claiming

Off-target scorebug cues have a median distance of **7.3s against a 5s tolerance**. That shape suggests a systematic lag — the count updating a beat after the pitch — rather than random noise, and a constant correction would convert false positives into true positives, lifting both numbers at once.

I am not claiming that, because the near-miss table mixes `appearance`, `at_bat` and `pitch` cues, so I cannot attribute the 7.3s to count-derived cues specifically. Measuring it needs per-event-type distances, which the report does not currently break out.

## Recommendation: do not merge this yet

It trades a clean 86.9%/100.0% for 90.1%/94.2%. Neither passes both gates, and the second gives up perfect precision — which for a citation product is the more valuable of the two, since a wrong moment is worse than a missing one.

Worth noting what changed underneath this: retrieval now demonstrably works end to end on the real game (see #741), and for the Giants' own games the feed already gives 99.4%/100% with richer text. The feed-free recall gate matters for archives without feeds, but it is not what the pilot turns on, and I would rather leave this open than merge a precision regression to move a proxy metric.

Issue #86, opened 2026-03-01
ElevenLabs STT (Scribe) not wired as server-side ingestion provider


Ali-Bilge:
Closed by PR #95 (Issue #83/#93: close SPEC-gap implementation bundle).

This issue has been resolved:
- Added speech-to-text transcription capabilities with ElevenLabs Scribe support
- STT provider is now configurable via `stt.provider` config key (mistral/elevenlabs)
- ElevenLabs STT properly wired into the server-side ingestion pipeline

PR #798, opened 2026-08-09
feat(ingest): let an operator replace the discovery ignore list (#773)

### LINKED ISSUE #773: ingest: no operator override for the default excluded-directory set
## Summary

The default excluded-directory set is a hard floor with no operator override. An
operator who wants `dist/`, `node_modules/`, `vendor/`, `__pycache__/`, `build/`,
or `.venv/` indexed has no way to say so.

## Evidence

- `internal/corpusfs/corpusfs.go` `defaultExcludedDirs` is a package-level map,
  consulted by `LocalFS` (via `shouldSkipDirectory`), by `S3FS` (via
  `keyHasExcludedDir`), and by the ingest watcher (via `corpusfs.IsExcludedDir`).
  Nothing reads config.
- `security.path_excludes` is operator-configurable but purely additive: it can
  only exclude more, never re-include.
- `.gitignore` support (`ingest.gitignore`) is likewise exclusion-only; a `!`
  negation cannot undo a hardcoded discovery skip.
- There is no `.dir2mcpignore` or equivalent include mechanism.

This became sharper with #716, which added `dist`, `build`, and `.venv` to the
set to conform to SPEC 7.1. That is the right default, but it widens the set of
operators who might legitimately disagree: a corpus of published static-site
output really does live in `dist/`, and a documentation corpus can legitimately
live under `build/`.

## Proposed behavior

A config key (working name `ingest.exclude_dirs`) that REPLACES the default
directory ignore set, so an operator can drop a name from it:

```yaml
ingest:
  exclude_dirs: [".git", ".dir2mcp", "node_modules", "vendor", "__pycache__", "build", ".venv"]
```

Open questions for the spec discussion:

- Replace-the-list versus an additive/subtractive pair. Replacing matches how
  `security.path_excludes` behaves today; a subtractive `include_dirs` is more
  surgical but adds a second ordering rule.
- Whether `.dir2mcp` and `.git` should be removable at all. `.dir2mcp` is
  dir2mcp's own state directory and indexing it is self-referential; a floor of
  one or two non-removable names may be worth keeping.
- LocalFS and S3 must stay in parity, and the ingest watcher must read the same
  resolved list as the initial scan.

## Process note

This is a config-surface change, so it is spec-first: a `dirstral-spec` PR
defining the key (SPEC 7.1 / the config reference), merged, then a gated
submodule re-pin in the dir2mcp code PR.

## Acceptance criteria

- New config key defined in `dirstral-spec` first, then implemented.
- Default value equals today's `defaultExcludedDirs` so behavior is unchanged
  when the key is absent.
- Honored identically by LocalFS discovery, S3 discovery, and the file watcher.
- `config init` writes it and the round-trip test (#626) covers it.
- Tests: a corpus with `dist/` indexed when the operator removes `dist` from the
  list; unchanged default behavior when the key is absent.

yidakra:
Closes #773.

Spec-first: `dirstral-spec` PR #86 (0.47.0) is merged. This PR re-pins the submodule to `dirstral-spec` `main`, which carries 0.46.0 and 0.47.0. PR #797 re-pins to the same commit. If it merges first, this re-pin becomes a no-op.

## The problem

The default excluded-directory set was a hard floor. An operator whose corpus is published static-site output in `dist/`, or documentation under `build/`, had no way to index it.

- `security.path_excludes` only adds exclusions. It cannot re-include.
- A `.gitignore` negation cannot help either. The walk tests the directory NAME before it reads any rule in that subtree, so it never descends and the rule is never consulted.

## Design

`ingest.exclude_dirs` replaces the directory ignore list.

**Why replace and not add/subtract.** A present key replaces the default list in full. It does not add to it. `security.path_excludes` already behaves this way, and `mergeFileConfig` already implements it: a present key wins wholesale. A reader therefore learns one list rule, not two. The alternative, an additive key plus a subtractive `include_dirs`, is more surgical but adds a second ordering rule to a surface that has none today. An absent key keeps the default list, so an existing corpus does not change.

The default list is the EIGHT names in SPEC 7.1: `.git`, `.dir2mcp`, `node_modules`, `vendor`, `__pycache__`, `dist`, `build`, `.venv`. Earlier spec editions named six. The reference implementation has always excluded eight.

**How `.dir2mcp` is forced back.** The state directory lives under `root_dir`. To index it is self-referential: the walk reads the index the same run writes, and watch mode feeds its own writes back into ingest. Two layers add the name back.

1. `internal/config` `resolveExcludeDirs` appends `.dir2mcp` when the operator's list omits it, and returns a warning. The loader appends that warning to `cfg.Warnings`, which is the spec's SHOULD ("an implementation MUST add `.dir2mcp/` back, and SHOULD say that it did").
2. `corpusfs.ResolveExcludedDirs` adds it back again, because it is the enforcement point and a direct `corpusfs` caller must not be able to skip the rule.

**One resolved value, three consumers.** The set is resolved in `internal/config`, carried on `ingest.DiscoverOptions.ExcludeDirs` to `corpusfs.Options.ExcludeDirs`, and read by:

- `LocalFS` (`discoverWalker.shouldSkipDirectory`, `internal/corpusfs/local.go`)
- the S3 lister (`keyHasExcludedDir`, `internal/corpusfs/s3.go`)
- the file watcher (`shouldSkipDirectory` in `internal/ingest/gitignore.go`, used from `internal/ingest/watch.go`)

The watcher takes it from the same `DiscoverOptionsFromConfig` the initial scan uses, so parity holds by construction. `corpusfs.IsExcludedDir` is gone: it read the default list only, so keeping it would leave a way to bypass the operator's list. That is the drift #716 fixed and a second copy would undo.

`ExcludedDirSet`'s zero value resolves to the default list, so a consumer that forgets to resolve still excludes `.git` and `.dir2mcp` instead of indexing them.

## The trap, and where it is documented

**The gates are independent and they compose by AND.** A file reaches the index only if the directory ignore list, `security.path_excludes` and the optional `.gitignore` rules ALL admit it.

Five names are in BOTH the default ignore list and the default `security.path_excludes`: `.git`, `.dir2mcp`, `node_modules`, `vendor`, `__pycache__`. An operator who clears only `ingest.exclude_dirs` sees no change for those five, because `path_excludes` drops the same file a moment later. `node_modules` is blocked in retrieval as well (`internal/retrieval/service.go` `defaultPathExcludes`), so it needs a third edit.

The spec requires this to be documented where the key is documented. It is in two places:

1. The config field comment on `Config.IngestExcludeDirs` (`internal/config/config.go`).
2. The `config init` output. `marshalConfigYAML` now has a `writeComment` helper, and the generated `.dir2mcp.yaml` carries the rule above the key:

```yaml
# ingest_exclude_dirs: directory names discovery does not descend into (SPEC 7.1).
# A present key REPLACES this list in full. It does not add to it.
# An entry is a plain directory name. It is not a path and it is not a glob.
# `.dir2mcp` is always kept: the state directory is under root_dir, so to
# index it is self-referential.
# The gates compose by AND. A file is indexed only if this list,
# path_excludes and the optional .gitignore rules all admit it. These five
# names are in BOTH lists: .git, .dir2mcp, node_modules, vendor,
# __pycache__. To re-index one of them, clear the name from BOTH keys.
# Clearing one key alone changes nothing you can see.
ingest_exclude_dirs:
  - .dir2mcp
  ...
```

The loader already skips comment lines, so the generated file still round-trips.

## Also

An entry that is not a plain directory name is rejected at validation: a path (`src/dist`), a glob (`dist/**`), `.` or `..`. SPEC 7.1 says an entry is a name that matches at any depth, so such a value matches nothing. Rejecting it tells the operator at once instead of leaving them to wonder why the directory is still indexed.

## Verification

`make check` passes, `gocyclo -over 15` included. Two functions were extracted to stay inside the budget rather than work around it: `applyIngestExcludeDirsFileParsed` out of `applyIngestModesFileParsed`, and `assertExcludeDirsIsDocumented` out of the new round-trip test. `validateIngestExcludeDirs` is a `Config` method in the validator list, not another branch in `validateNumericBounds`.

Every new test was confirmed to FAIL against `main`, using the copy-aside method (no `git stash`, since `refs/stash` is shared across worktrees):

- `internal/corpusfs/local.go` and `s3.go` restored from `origin/main` (with a temporary `IsExcludedDir` shim so the package still compiled): the four replace-semantics tests failed, and `dist/bundle.js` stayed unindexed. `AlwaysKeepsTheStateDir` and `AbsentKeepsTheDefaults` passed, which is correct, because they encode unchanged behaviour.
- The watcher: `excluded` set to `corpusfs.ExcludedDirSet{}` at both construction points, which reproduces main's hardcoded default set exactly. `TestAddWatchDirs_HonorsConfiguredExcludeDirs` failed: `dist` was not watched and `notes` was.
- The config: the merge branch and the `writeList` call removed, which reproduces main's "no such key". Five config tests failed.

Each file was then restored from its copy and the suites re-run green.

New coverage:

- `tests/corpusfs/exclude_dirs_773_test.go`: drop a name to index `dist/`; replace does not add; `.dir2mcp` forced back; empty list; absent key unchanged; S3 and local parity; the resolver itself.
- `internal/ingest/exclude_dirs_773_test.go`: the watcher registers watches from the operator's list, and skips `.dir2mcp` when the list omits it. It is internal because it drives `addWatchDirs` directly, which is deterministic and does not wait on fsnotify event timing.
- `tests/config/exclude_dirs_773_test.go`: the key spelling (nested and inline), replace semantics, forced `.dir2mcp` plus its warning, empty list, entry-shape rejection.
- `tests/config/generated_config_roundtrip_test.go`: the list key is written by `config init`, documents the AND trap, loads back, and an operator edit that removes `dist` reaches the loader. The existing guard walks SCALAR keys only, so a list key it dropped would have passed unnoticed.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -16,6 +16,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/dirstral/dir2mcp/internal/corpusfs"
 	"github.com/dirstral/dir2mcp/internal/netutil"
 	"github.com/dirstral/dir2mcp/internal/provider"
 	"github.com/dirstral/dir2mcp/internal/secrets"
@@ -499,11 +500,40 @@ type Config struct {
 	IngestGitignore      bool
 	IngestFollowSymlinks bool
 	IngestMaxFileMB      int
-	IngestPDFMode        string
-	IngestImagesMode     string
-	IngestAudioMode      string
-	IngestArchivesMode   string
-	IngestExtractor      string
+	// IngestExcludeDirs is the directory ignore list applied by discovery
+	// (config `ingest.exclude_dirs`, SPEC §7.1, issue #773). Discovery does not
+	// descend into a directory whose NAME matches an entry. An entry is a plain
+	// directory name: it is not a path and it is not a glob, and it matches at
+	// any depth below the root.
+	//
+	// A present key REPLACES the default list in full. It does not add to it.
+	// That is the list behaviour of `security.path_excludes`, so there is one
+	// rule to learn and not two. An absent key keeps the default list
+	// (corpusfs.DefaultExcludedDirs), so an existing corpus does not change.
+	//
+	// `.dir2mcp` is always added back, even when the operator's list omits it.
+	// The state directory lives under root_dir, so to index it is
+	// self-referential: the walk would read the index the same run writes, and
+	// watch mode would feed its own writes back into ingest. The loader records
+	// a warning when it adds the name back.
+	//
+	// THE GATES ARE INDEPENDENT AND THEY COMPOSE BY AND. A file reaches the
+	// index only if this list, `security.path_excludes` and the optional
+	// `.gitignore` rules ALL admit it. Five names are in BOTH this default list
+	// and the default `security.path_excludes`: `.git`, `.dir2mcp`,
+	// `node_modules`, `vendor` and `__pycache__`. To re-index such a directory
+	// an operator MUST clear the name from EVERY gate that lists it. To clear
+	// this one alone changes nothing the operator can observe, because
+	// `security.path_excludes` drops the same file a moment later.
+	// `node_modules` is blocked in retrieval as well
+	// (internal/retrieval/service.go defaultPathExcludes), so it needs a third
+	// edit.
+	IngestExcludeDirs  []string
+	IngestPDFMode      string
+	IngestImagesMode   string
+	IngestAudioMode    string
+	IngestArchivesMode string
+	IngestExtractor    string
 	// IngestOnUnsupported is the §7.4.B.2 degradation mode for a format no active
 	// extraction engine can read: "lenient" (default) skips with a warning and
 	// names the gap in the honest coverage report, "strict" records a non-fatal
@@ -1141,6 +1171,7 @@ type fileConfig struct {
 	IngestGitignore                    *bool
 	IngestFollowSymlinks               *bool
 	IngestMaxFileMB                    *int
+	IngestExcludeDirs                  []string
 	IngestPDFMode                      *string
 	IngestImagesMode                   *string
 	IngestAudioMode                    *string
@@ -1303,6 +1334,7 @@ type persistedConfig struct {
 	IngestGitignore                    bool          `yaml:"ingest_gitignore"`
 	IngestFollowSymlinks               bool          `yaml:"ingest_follow_symlinks"`
 	IngestMaxFileMB                    int           `yaml:"ingest_max_file_mb"`
+	IngestExcludeDirs                  []string      `yaml:"ingest_exclude_dirs"`
 	IngestPDFMode                      string        `yaml:"ingest_pdf_mode"`
 	IngestImagesMode                   string        `yaml:"ingest_images_mode"`
 	IngestAudioMode                    string        `yaml:"ingest_audio_mode"`
@@ -1555,6 +1587,7 @@ func Default() Config {
 		IngestGitignore:           true,
 		IngestFollowSymlinks:      false,
 		IngestMaxFileMB:           20,
+		IngestExcludeDirs:         corpusfs.DefaultExcludedDirs(),
 		IngestPDFMode:             "ocr",
 		IngestImagesMode:          "ocr_auto",
 		IngestAudioMode:           "auto",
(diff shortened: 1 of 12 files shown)


yidakra:
@coderabbitai review

PR #815, opened 2026-08-10
fix(retrieval): answer the adaptive skip verdict instead of the empty-context fallback (#685)

### LINKED ISSUE #685: retrieval: adaptive skip verdict never calls the generator and returns the empty-context fallback
## Summary

The adaptive retrieval gate's `Retrieve=false` path is intended to answer trivial/conversational prompts without corpus retrieval. Instead, `Ask` leaves `hits` empty and only invokes the generator when `len(hits) > 0`, so every skip verdict returns `No relevant context found in the indexed corpus.`

This is a correctness follow-up to #338: the gate makes a skip decision, but the answer path cannot execute the advertised no-retrieval behavior.

## Evidence

- `adaptiveGate` classifies trivial asks such as greetings as `{Retrieve:false, K:0, Class:"skip"}` (`internal/retrieval/adaptive.go`).
- `Ask` sets `skipRetrieval=true` and leaves `hits` empty (`internal/retrieval/service.go:1556-1590`).
- It initializes the answer through `buildFallbackAnswer`, whose zero-hit output is the fixed no-context message (`service.go:1604`, `service.go:3230-3234`).
- Grounded generation is gated on both `s.gen != nil` and `len(hits) > 0` (`service.go:1605`), so the configured generator is never called.
- Existing adaptive tests cover gate classification/k bounds, not `Ask` behavior for a skip verdict.

## Impact

Enabling adaptive retrieval makes harmless prompts like `hello` produce a misleading corpus-failure answer instead of a low-cost conversational answer. The feature saves retrieval cost by also disabling useful generation.

## Acceptance criteria

- Define and implement the no-retrieval answer path: invoke the generator with a bounded non-RAG prompt, or explicitly define skip as abstention and change the feature description accordingly.
- Do not fabricate citations/hits for no-retrieval answers.
- Preserve grounded-generation safeguards for retrieved answers.
- Add an end-to-end `Ask` test proving a skip verdict produces the intended answer and does not call embedding/search.

yidakra:
Closes #685

## What the user sees

Set `retrieval.adaptive.enabled: true`, then send `hello` to `ask`.

**Today (main):**

```
No relevant context found in the indexed corpus.
```

The server never searched the corpus for that message. The gate decided to skip
retrieval. The message therefore reports a corpus result that the server never
obtained. The user reads it as a corpus failure or an empty index.

**After this PR:**

```
Hello. I can search the indexed documents for you. Ask a question about them.
```

The text comes from the configured generator. `citations` stays empty. `hits`
stays empty. Without a generator, or after a generator error, the user gets a
deterministic reply that states the true server state:

```
No corpus lookup was needed for this message. Ask a question about the indexed
documents to search them.
```

## Mechanism

The adaptive gate returns `{Retrieve:false, K:0, Class:"skip"}` for a trivial
message. `Ask` set `skipRetrieval=true` and left `hits` empty. `Ask` called the
generator only when `s.gen != nil && len(hits) > 0`. So the skip verdict always
fell through to `buildFallbackAnswer`, whose zero-hit branch is the fixed
no-context string. The gate saved the retrieval cost and also removed the
answer.

`Ask` now selects the answer path on the verdict:

- skip verdict: `answerWithoutRetrieval` calls the generator with a bounded
  non-RAG prompt (`noRetrievalPrompt`, 128 max tokens through the existing
  `boundedGenerate` helper). The prompt carries no document and no untrusted
  document fence.
- retrieved ask: unchanged `generateGroundedAnswer` path.

New code lives in `internal/retrieval/adaptive.go`. The `service.go` edit is one
call site plus the gate block, which moves into `applyAdaptiveGate`.

## Guard: which text the gate classifies

An ask carries two texts. `query.Query` is the retrieval query. `question` is
the text the answer must satisfy. `Ask` copies the question into an empty query,
so the two agree unless a caller overrides the query.

The gate classified `query.Query` alone. That was harmless while a skip verdict
returned the no-context string. It is not harmless once a skip verdict answers:
a caller could pair a substantive question with a trivial query override, and
the generator would answer a real question with no evidence.

`applyAdaptiveGate` therefore re-classifies `question` before it skips. It skips
only when the question also carries no information need. It retrieves in every
other case. The k decision still follows the retrieval query, so no k class
moves and every pre-existing gate test keeps its expected k.

## Grounding safety

- The model receives no document, so `citations` and `hits` stay empty. The PR
  fabricates nothing (SPEC §9.4.1).
- `stripHallucinatedCitations(answer, nil)` removes any `[rel_path]` tag the
  model writes on the no-retrieval path. The allowed set is empty by
  construction.
- `ensureAnswerAttributions` already removes a model-authored `Sources:` footer
  and adds none when the citation list is empty.
- The prompt tells the model to state no fact about the indexed documents.

## SPEC check (§9.4.3)

The skip verdict is **not** the insufficient-evidence abstention. §9.4.3 judges
an **eligible** hit set that retrieval produced, and it fires when that set is
empty or does not clear the absolute evidence threshold. The skip path runs
**before** retrieval, so no eligible set exists and §9.4.3 has nothing to judge.
`abstainOnWeakEvidence` returns early on `len(hits) == 0` and is untouched, so
the real abstention keeps its exact answer text and its empty citations array.

The SPEC does not mention adaptive retrieval at all. The gate is a config-only
dir2mcp feature (`retrieval.adaptive.*`), so no spec change is needed and no
spec-pinned behavior moves. The pinned `dirstral-spec` commit is unchanged.

## Tests

`tests/retrieval/adaptive_skip_answer_685_test.go`, seven cases:

1. `CallsGeneratorWithoutRetrieval`: the skip verdict returns the generator
   output, never queries the index, never embeds the query, carries no citation
   and no hit, and sends a prompt with no document fence. **Fails on main.**
2. `NoGeneratorFallback`: without a generator the answer is not the empty-corpus
   message. **Fails on main.**
3. `GeneratorErrorFallsBack`: a failed generation degrades to the deterministic
   reply, not to the empty-corpus message. **Fails on main.**
4. `StripsFabricatedSources`: an invented file tag and an invented `Sources:`
   footer do not reach the caller.
5. `SubstantiveQuestionNeverSkips`: a real question with a trivial query
   override still retrieves and still embeds. **Fails on main.**
6. `TrivialQuestionWithRichQueryRetrieves`: a hard query override keeps its
   retrieval and its widened k.
7. `RetrievedAskUnchanged`: a real question still retrieves, still embeds, and a
   zero-hit retrieved ask still returns the empty-corpus fallback without a
   generator call.

Proof against the sources on `main`:

```
--- FAIL: TestAdaptiveSkip685_CallsGeneratorWithoutRetrieval
--- FAIL: TestAdaptiveSkip685_NoGeneratorFallback
--- FAIL: TestAdaptiveSkip685_GeneratorErrorFallsBack
--- PASS: TestAdaptiveSkip685_StripsFabricatedSources
--- FAIL: TestAdaptiveSkip685_SubstantiveQuestionNeverSkips
--- PASS: TestAdaptiveSkip685_TrivialQuestionWithRichQueryRetrieves
--- PASS: TestAdaptiveSkip685_RetrievedAskUnchanged
```

All seven pass on this branch. The six pre-existing `TestAdaptive_*` gate tests
also still pass. `make check` passes, gocyclo `-over 15` included.

## Out of scope

- The `Hits` field stays `nil` on the skip path, exactly as before. The MCP layer
  serializes it to an empty array.
- The gate's own classification rules are untouched. This PR changes the answer
  for a skip verdict, not which messages a trivial-token scan calls trivial.
- No new config key. The behavior follows `retrieval.adaptive.enabled`, which is
  off by default, so a default install sees no change.

### DIFF
--- a/internal/retrieval/adaptive.go
+++ b/internal/retrieval/adaptive.go
@@ -5,6 +5,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/dirstral/dir2mcp/internal/model"
 	"github.com/dirstral/dir2mcp/internal/usage"
 )
 
@@ -178,6 +179,45 @@ func normalizeAdaptiveBounds(kMin, kMax int) (int, int) {
 		kMax = kMin
 	}
 	return kMin, kMax
+}
+
+// applyAdaptiveGate resolves the opt-in adaptive retrieval verdict for one ask
+// (config retrieval.adaptive.*). It reports whether retrieval must be skipped
+// and it writes the effective k into query when retrieval runs. With the gate
+// disabled it is a no-op, so the fixed-k path is unchanged.
+//
+// An ask carries two texts. `query.Query` is the retrieval query. `question` is
+// the text the answer must satisfy. Ask copies the question into an empty query,
+// so the two agree unless a caller overrides the query. The gate keeps its k
+// decision on the retrieval query, because k bounds that search. A SKIP verdict
+// needs more care: it removes the evidence from an answer, so the answered
+// question must carry no information need either. A caller that pairs a
+// substantive question with a trivial query override would otherwise get an
+// ungrounded answer to a real question (#685). The gate therefore re-classifies
+// the question before it skips, and it retrieves when the question asks for
+// something.
+func (s *Service) applyAdaptiveGate(question string, query *model.SearchQuery) bool {
+	s.metaMu.RLock()
+	enabled := s.adaptiveEnabled
+	kMin := s.adaptiveKMin
+	kMax := s.adaptiveKMax
+	s.metaMu.RUnlock()
+	if !enabled {
+		return false
+	}
+	decision := adaptiveGate(query.Query, query.K, kMin, kMax)
+	s.logf("adaptive gate: class=%s retrieve=%t k=%d", decision.Class, decision.Retrieve, decision.K)
+	if decision.Retrieve {
+		query.K = decision.K
+		return false
+	}
+	asked := adaptiveGate(question, query.K, kMin, kMax)
+	if !asked.Retrieve {
+		return true
+	}
+	s.logf("adaptive gate: skip on the query, class=%s on the question; retrieving with k=%d", asked.Class, asked.K)
+	query.K = asked.K
+	return false
 }
 
 // The no-retrieval answer path (issue #685).
--- a/internal/retrieval/adaptive.go
+++ b/internal/retrieval/adaptive.go
@@ -1,8 +1,11 @@
 package retrieval
 
 import (
+	"context"
 	"regexp"
 	"strings"
+
+	"github.com/dirstral/dir2mcp/internal/usage"
 )
 
 // adaptiveDecision is the verdict of the training-free retrieval gate for a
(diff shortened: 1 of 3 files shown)


yidakra:
@coderabbitai review

The second commit had no review pass, because the CLI hit its rate limit during the run. Please review the full diff again, with attention to `applyAdaptiveGate`: it classifies the retrieval query first and then re-classifies the question before it skips retrieval.

PR #816, opened 2026-08-10
fix(retrieval): refill the result slots that dropped summary hits took (#686)

### LINKED ISSUE #686: retrieval: summary hits consume top-k slots before being dropped, causing hierarchical result loss
## Summary

Search retrieves/truncates to the requested candidate pool before hierarchical processing. If a `summary` vector occupies a slot and is then dropped—because hierarchical mode is disabled, expansion is unavailable/fails, or all children are filtered—the pipeline does not refill from the next fine-chunk candidates. Search returns fewer than `k` even when enough eligible fine chunks exist.

This is a correctness follow-up to #329.

## Evidence

- With recency decay off, `poolK := k` and `searchExpanded(..., poolK)` completes before `expandHierarchical` (`internal/retrieval/service.go:1125-1158`).
- `expandHierarchical` drops every summary when disabled or unsupported (`internal/retrieval/hierarchical.go:62-73`).
- Expansion errors and empty/filtered expansions also drop the summary (`hierarchical.go:75-105`).
- No path asks the index for replacement candidates after a dropped summary.
- The canonical contract says disabled/default retrieval behaves exactly as flat retrieval and summaries only add coarse vectors; stale summary vectors can remain after toggling the feature off.

## Reproduction shape

Index ordered candidates `[summary S, fine A, fine B]`, request `k=2`, and disable hierarchical retrieval (or make S expansion fail). Search first obtains `[S, A]`, then drops S and returns only `[A]`; fine B is never considered.

## Impact

Results silently under-return and can lose the second-best real source chunk. Turning hierarchical retrieval off does not restore flat behavior if summary vectors already exist; fail-open expansion failures also degrade recall.

## Acceptance criteria

- Over-fetch/refill enough candidates so dropping summary routing nodes does not consume the caller's `k` fine-hit budget.
- Disabled mode behaves identically to a corpus/index without summary vectors.
- Expansion errors remain citation-safe while returning the best available fine hits.
- Add deterministic enabled-disabled, expansion-error, and filtered-child tests with `k` smaller than the indexed candidate set.

yidakra:
Closes #686.

## Problem

A `summary` chunk is a routing device. SPEC 9.7 says it never reaches the
caller. But `search()` asked the index for exactly the caller's pool size
(`poolK`), and only then dropped the summary candidates. Each dropped summary
took one result slot from a real fine chunk, and no code refilled it.

A summary is dropped in four cases:

1. hierarchical retrieval is off (the default),
2. the store cannot expand,
3. the expansion call fails,
4. the query filters remove every child.

This also broke the SPEC 9.7 rule that a disabled feature behaves **exactly** as
flat retrieval. A corpus that still holds summary vectors under-returned after
an operator turned the feature off.

## Measured loss

The new tests index one summary and three eligible fine chunks, then ask for
`k=2`. Each drop path was measured on `main` before any code change:

| Case | Before (main) | After | Slots lost |
|---|---|---|---|
| hierarchical disabled | 1 hit `[11]` | 2 hits `[11, 12]` | 1 of 2 |
| expansion returns an error | 1 hit `[11]` | 2 hits `[11, 12]` | 1 of 2 |
| every child filtered out | 1 hit `[11]` | 2 hits `[11, 12]` | 1 of 2 |
| disabled vs a corpus with no summary vector | not equal | equal | 1 of 2 |

The loss is **1 of the 2 requested slots, or 50 percent of the budget**, for
each dropped summary. In general the loss is one result per dropped summary in
the pool.

## Fix

`searchWithSummaryRefill` wraps retrieval plus the 9.7 expand step. When
dropped summaries left the pool short of the budget, it re-retrieves a wider
pool. The widened size is the caller's pool plus the number of summary
candidates the previous round saw. This mirrors the tombstone oversampling of
SPEC 6.6: ask for more, remove the ineligible candidates, keep the first `k`.

The refill runs **only** when the pool both holds a summary and came up short.
A corpus with no summary vectors makes exactly one retrieval call, so the flat
path is unchanged.

`needSummaryRefill` holds the stop rule, so the loop always terminates. It
stops when the budget is met, when no summary took a slot, when the index
returned fewer candidates than it was asked for, when the widened pool would
not grow, and at a round cap of 3.

## Cost

A refill round re-runs retrieval only. The HyDE hypothesis and the
cross-lingual query variants come from the expansion cache, so no extra
generation is paid. The round does re-embed the query text, so a corpus with
stale summary vectors pays up to two extra query embeddings on an affected
query. The extra margin a refill round adds is capped at 200 candidates, so a
corpus made almost entirely of summaries cannot turn one search into an index
scan. The cap sits on the margin and never on the pool, so a widened round
always asks for at least as many candidates as the round it repairs.

## SPEC

I read `dirstral-spec/docs/SPEC.md` sections 6.6, 9.1 and 9.7 before the change.
**No spec change is needed.** The SPEC does not pin the candidate pool size. It
pins the opposite of the current code:

- 9.7: "When disabled ... retrieval behaves **exactly** as flat retrieval"
- 6.6: retrieval already oversamples for tombstones, then returns "the first
  `k` remaining"

The fix restores behavior the SPEC already requires. It does not change ranking
semantics: order, dedup, rerank and the citation-faithfulness invariant are all
untouched.

## Tests

- `tests/retrieval/hierarchical_refill_686_test.go`: 5 black-box tests. Four of
  them **fail on `main`** with the loss numbers above. The fifth pins
  termination when the corpus holds fewer fine chunks than `k`.
- `internal/retrieval/hierarchical_refill_686_test.go`: a table test over every
  stop condition of `needSummaryRefill`, plus `countSummaryHits`.

`make check` passes, including `gocyclo -over 15`.

## Not done

- No refactor of the surrounding retrieval code. The change is scoped to #686.
- No per-search query-embedding cache. That would remove the extra embed call
  of a refill round, but it is a separate change to the embedding path.
- No `dir2mcp_stats` field for refill rounds. Summary coverage reporting is
  already deferred by SPEC 9.7.

### DIFF
--- a/internal/retrieval/hierarchical.go
+++ b/internal/retrieval/hierarchical.go
@@ -45,9 +45,12 @@ type summaryExpander interface {
 // only stops a pathological corpus from re-querying the index for ever.
 const summaryRefillMaxRounds = 3
 
-// summaryRefillMaxPool caps the widened pool, so a corpus made almost entirely
-// of summary vectors cannot turn one search into an unbounded index scan.
-const summaryRefillMaxPool = 200
+// summaryRefillMaxMargin caps how many EXTRA candidates a refill round may ask
+// for on top of the caller's own pool, so a corpus made almost entirely of
+// summary vectors cannot turn one search into an unbounded index scan. The cap
+// is on the margin, never on the pool, so a widened round always asks for at
+// least as many candidates as the round it repairs.
+const summaryRefillMaxMargin = 200
 
 // searchWithSummaryRefill retrieves the candidate pool and applies the §9.7
 // expand step. It re-retrieves a WIDER pool when dropped `summary` hits left the
@@ -78,10 +81,7 @@ func (s *Service) searchWithSummaryRefill(ctx context.Context, query model.Searc
 		if !needSummaryRefill(len(hits), poolK, summaries, len(raw), fetchK, round) {
 			return hits, nil
 		}
-		fetchK = poolK + summaries
-		if fetchK > summaryRefillMaxPool {
-			fetchK = summaryRefillMaxPool
-		}
+		fetchK = poolK + summaryRefillMargin(summaries)
 	}
 }
 
@@ -102,7 +102,18 @@ func needSummaryRefill(got, poolK, summaries, rawLen, fetchK, round int) bool {
 	if round+1 >= summaryRefillMaxRounds {
 		return false
 	}
-	return poolK+summaries > fetchK && fetchK < summaryRefillMaxPool
+	// Only widen when the next request would really be larger than this one.
+	return poolK+summaryRefillMargin(summaries) > fetchK
+}
+
+// summaryRefillMargin is the number of extra candidates a refill round adds to
+// the caller's pool: one per summary that took a slot, bounded by
+// summaryRefillMaxMargin.
+func summaryRefillMargin(summaries int) int {
+	if summaries > summaryRefillMaxMargin {
+		return summaryRefillMaxMargin
+	}
+	return summaries
 }
 
 // countSummaryHits reports how many `summary` candidates the pool holds. Every
--- a/internal/retrieval/hierarchical.go
+++ b/internal/retrieval/hierarchical.go
@@ -38,6 +38,86 @@ type summaryExpander interface {
 	ExpandSummaryChunk(ctx context.Context, chunkID uint64) ([]model.ChunkMetadata, error)
 }
 
+// summaryRefillMaxRounds bounds how many times search re-retrieves a widened
+// candidate pool to replace dropped `summary` routing nodes (#686). Each round
+// asks for the caller's pool PLUS the summaries the previous round saw, so two
+// extra rounds already cover a pool whose summary count keeps growing. The cap
+// only stops a pathological corpus from re-querying the index for ever.
+const summaryRefillMaxRounds = 3
+
+// summaryRefillMaxPool caps the widened pool, so a corpus made almost entirely
+// of summary vectors cannot turn one search into an unbounded index scan.
+const summaryRefillMaxPool = 200
+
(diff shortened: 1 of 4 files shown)


yidakra:
Review note on `searchWithSummaryRefill`: the loop returns the LAST round, and it discards the rounds before it.

```go
hits := s.expandHierarchical(ctx, query, raw, fetchK)
summaries := countSummaryHits(raw)
if !needSummaryRefill(...) {
    return hits, nil   // the last round, even when an earlier round found more
}
```

A wider pool usually holds more fine chunks. It does not have to. Vector search is approximate, so a query with a larger k explores a different part of the graph and its result set is not guaranteed to contain the smaller query's result set. A wider round whose extra candidates are summaries that expand to nothing then returns fewer hits than the round it was meant to repair, and the refill loses results instead of recovering them.

The escalation loop next door already takes the other choice. `collectFilteredCandidates` (internal/retrieval/service.go) widens its fanout in the same way, but it ACCUMULATES into `filtered` across rounds and never throws an earlier round away:

```go
for n := initialSearchFanout(k, overfetchMultiplier); len(filtered) < k; {
    hits, err := idx.Search(ctx, vector, n, backendFilter)
    s.collectFilteredHits(hits, indexName, filters, k, seen, &filtered)
    ...
}
```

So the two widening loops in the same package now disagree about what a round is worth.

Two possible answers, smallest first:

1. Keep the best round: track the longest result and return that when the loop gives up. One comparison.
2. Accumulate across rounds, deduped by chunk id, which matches `collectFilteredCandidates` exactly.

I tried to build a failing case through the public `Search` path and did not get one. The internal fanout escalation re-queries the index inside a single `searchExpanded` call, so a stub index that models an approximate non-superset also perturbs that loop, and the resulting test measures the stub rather than the pipeline. So I report this as a design inconsistency with a real mechanism behind it, not as a proven defect. Your call whether it is worth the change now or a follow-up issue.

Issue #679, opened 2026-08-03
watch: full internal job queue silently drops file changes for up to 10 minutes


yidakra:
## Summary

The watcher decouples fsnotify draining from slow ingestion with a 256-entry internal job channel, but when that channel fills, debounced jobs are silently discarded. Unlike a kernel fsnotify overflow, this path neither increments `watch_overflows` nor requests the already-available immediate rescan; correctness waits for the fixed 10-minute ticker.

This is distinct from, and a follow-up to, #409 item 5: that fix handles `fsnotify.ErrEventOverflow`, not overflow of dir2mcp's own queue.

## Evidence

- `internal/ingest/watch.go:83-92` creates `fire: make(chan watchJob, 256)`.
- A single worker consumes jobs and performs potentially slow extraction/embedding (`watch.go:175-191`).
- Each debounce callback uses a non-blocking send; the `default` branch only comments that the safety rescan will reconcile (`watch.go:257-273`).
- Immediate coalesced rescan machinery exists in `run`, but `arm` cannot invoke it.
- `watch_overflows` is incremented only for kernel `fsnotify.ErrEventOverflow` (`watch.go:157-168`).

A burst of more than 256 paths after the debounce window, while one large document is processing, fills the queue and drops arbitrary changes.

## Impact

Files can remain absent or stale in retrieval for up to ten minutes with no warning or stats signal, even though the kernel watcher itself did not overflow.

## Acceptance criteria

- Internal queue saturation triggers an immediate coalesced rescan (or uses a lossless bounded scheduling design).
- Surface the loss through logs and an honest overflow/drop counter.
- Add a deterministic test with a saturated job queue and assert prompt reconciliation.

PR #822, opened 2026-08-11
fix(watch): stop serving stale chunks after a directory or eligibility change (#678, #680)

### LINKED ISSUE #678: watch: removing or moving out a directory leaves descendant documents searchable until safety rescan
## Summary

The fsnotify path handles a directory remove/rename as a single-path delete. `processDelete` tombstones only the exact directory path, while indexed documents are stored under descendant file paths. Moving or deleting a populated subtree can therefore leave every descendant searchable until the 10-minute safety rescan.

This is a follow-up to #409: individual file deletion is reconciled, but subtree removal is not.

## Evidence

- `internal/ingest/watch.go:195-202` arms the raw event path as a delete for `Remove|Rename`.
- When that path is gone, `process` calls `processDelete(ctx, rel)` (`watch.go:288-293`).
- `processDelete` calls `MarkDocumentDeleted` for exactly one `relPath` and evicts only that path (`watch.go:522-548`).
- The only broad reconciliation is `safetyRescanInterval = 10 * time.Minute` (`watch.go:20-25`).

A filesystem watcher is not guaranteed to emit per-child remove events when a watched directory is renamed out of the tree or removed. The top-level directory event is therefore insufficient to identify stored descendants.

## Impact

Search and ask can return documents from a directory that no longer exists, with stale citations, for up to ten minutes. Large subtree moves make the stale set correspondingly large.

## Acceptance criteria

- A removed/renamed directory immediately tombstones and evicts every active document beneath its relative path, or requests an immediate coalesced rescan.
- Preserve the existing atomic-save delete/recreate handling for files.
- Add watcher tests for deleting and moving out a populated nested directory.

### LINKED ISSUE #680: watch: files that become oversized or newly ignored retain their old searchable chunks
## Summary

On a normal write event, if a previously indexed path now fails discovery filters, the watcher simply drops the job. It does not tombstone/evict the old document. This affects files that grow beyond the size cap, become newly gitignored, turn into an unfollowed symlink, or otherwise become non-indexable while still existing.

## Evidence

- `internal/ingest/watch.go:304-314` documents the transition but calls `processDelete` only when `job.deleted` is true; an unindexable write returns without reconciliation.
- `indexableFile` rejects oversize files at `watch.go:346-347` and gitignored files at `watch.go:349-355`.
- Editing `.gitignore` itself also does not trigger a rescan of paths whose eligibility changed.
- The periodic safety rescan is fixed at ten minutes.

## Reproduction examples

- Index `notes.txt`, then append until it exceeds `ingest.max_file_mb`.
- Index `private/notes.txt`, then add `private/` to `.gitignore`.

In both cases the old chunks remain live after the write event, despite the path no longer being eligible under current ingest policy.

## Impact

Stale content remains searchable and citable. In the gitignore case this can be a policy/security surprise: an operator excludes a path but it remains retrievable until a later full reconciliation.

## Acceptance criteria

- Reconcile an existing document that transitions from indexable to non-indexable immediately: persist the appropriate skipped state/reason and evict its chunks.
- Changes to eligibility rules such as `.gitignore` trigger reconciliation of affected paths.
- Add watcher tests for size-cap and gitignore transitions.

yidakra:
## The symptom a user sees

Content that must leave the index stays searchable for up to **10 minutes**.

- A user deletes a directory, or moves it out of the corpus. `search` and `ask`
  keep returning the files that were inside it. The citations point at paths
  that no longer exist.
- A user adds a path to `.gitignore`. The excluded content still answers
  queries. This is a policy surprise: the operator excluded the path and it is
  still retrievable.
- A file grows past `ingest.max_file_mb`. Its old chunks stay live.

Today only one thing closes that window: the periodic safety rescan in the
watch loop, `safetyRescanInterval = 10 * time.Minute`. Nothing else reconciles
these three cases, so the stale window is up to a full rescan interval. On a
large subtree move the stale set is the whole subtree.

Both issues are the same defect class in the watch layer: a document that must
stop being searchable keeps its old chunks.

## Commit 1: removed or renamed directory (#678)

A filesystem reports the removal or the rename of a directory as **one** event,
for the directory path. The store holds no document at that path. It holds one
document per descendant file. `processDelete` tombstoned only the exact path,
so every descendant survived.

The watcher now runs a subtree pass when a path is proven gone from disk. It
lists the active documents below that path and tombstones each one.

To remove chunks is destructive, so the pass is deliberately narrow:

- The caller must first prove the parent path is gone. A document at
  `dir/child` is only reachable through a directory at `dir`.
- The store prefix filter is a coarse pre-filter only. `ListFiles` normalizes
  the prefix, which drops the trailing slash, and SQLite `LIKE` compares ASCII
  letters case-insensitively. The prefix `docs` therefore also matches
  `docsets/a.md` and `Docs/a.md`. A byte-exact test on `docs/` selects the real
  set.
- Each remaining candidate is checked with `Lstat`. A path that still exists
  keeps its chunks.

The pass also retires archive members. Their `rel_path` is synthetic
(`bundle.zip/inner.txt`), so no filesystem event ever names them.

## Commit 2: a file that stops being eligible (#680)

A write event on a path that fails the discovery filters was dropped. The
document kept its chunks.

The watcher now reconciles the transition at once. Each outcome mirrors what a
full scan does with the same path, so watch and rescan agree:

| transition | store outcome |
| --- | --- |
| past `ingest.max_file_mb` | visible row, `status=skipped`, `skip_reason=size_cap`, chunks evicted |
| symlink with following off | visible row, `skip_reason=symlink_ignored`, chunks evicted |
| newly gitignored | tombstoned |
| now a directory, or no longer a regular file | tombstoned |

A repeated write does not repeat the work. A row that already carries the same
skipped state is left alone, because the pass that set it evicted the chunks.

An edit to a `.gitignore` file also asks for a coalesced rescan. One rule can
flip the eligibility of many paths, and no per-path event describes that.

## SPEC

- **§6.6 (tombstone invariant) is preserved.** Every eviction runs through
  `MarkDocumentDeleted`, which sets `deleted=1` on the document, its
  representations and its chunks. Nothing is hard-deleted. The size-cap case
  writes the document row back with `deleted=0` after the eviction, so the
  operator keeps a visible skipped row while the representations and the chunks
  stay tombstoned. That is exactly how the initial scan represents a file it
  refused.
- **§7.1 (discovery ignore list) is unchanged.** The same gates decide
  eligibility. This PR only changes **when** the store learns the answer.

No spec change is needed, so no `dirstral-spec` PR is required.

## Proof the tests fail on `main`

Each set was run against `main` with the fix reverted and the tests in place.

`tests/ingest/watch_subtree_delete_678_test.go`:

```
--- FAIL: TestWatch_TombstonesDescendantsOfMovedOutDir (3.17s)
    timed out waiting for "docs/top.md" present=false; got paths:
    map[docs/a/b/deep.md:true docs/bundle.zip:true docs/bundle.zip/inner.txt:true
        docs/top.md:true docsets/keep.md:true keep.txt:true]
--- FAIL: TestWatch_TombstonesDescendantsOfRenamedDir (3.17s)
    timed out waiting for "docs/top.md" present=false; got paths:
    map[archive/a/b/deep.md:true ... docs/top.md:true ...]
--- FAIL: TestWatch_TombstonesDescendantsOfRemovedDir (3.20s)
    timed out waiting for "docs/bundle.zip/inner.txt" present=false; got paths:
    map[docs/bundle.zip/inner.txt:true docsets/keep.md:true keep.txt:true]
```

`tests/ingest/watch_ineligible_transition_680_test.go`:

```
--- FAIL: TestWatch_EvictsChunksWhenFileGrowsPastSizeCap (3.17s)
    timed out waiting for "notes.txt" to lose its chunks; still active: map[raw_text:true]
--- FAIL: TestWatch_EvictsChunksWhenPathBecomesGitignored (3.14s)
    timed out waiting for "private/notes.txt" present=false; got paths:
    map[.gitignore:true keep.txt:true private/notes.txt:true]
```

All five pass with the change.

## Deletion safety

Every test asserts that a document which is still eligible keeps its chunks:

- `docsets/keep.md` is the prefix trap. It shares the `LIKE` prefix of the
  removed `docs` directory and it must survive.
- `keep.txt` is an unrelated control in every test.
- The size-cap test also runs the transition in reverse. The file comes back
  under the cap and must become searchable again, which proves the
  reconciliation does not wedge the document.

## Checks

- `make check` passes. `gocyclo -over 15` is clean; `process` was split into
  `relPath`, `rescanOnIgnoreRuleChange`, `processStatError` and
  `reconcileIneligibleDocument`.
- `coderabbit review` was run. The one finding (assert the skipped row is not
  tombstoned) is fixed.
- `README.md` records both behaviors in the watcher section.

Closes #678
Closes #680


yidakra:
@coderabbitai review

yidakra:
Review note on `activeDocumentsUnder`, not blocking.

The subtree pass runs one paged `ListFiles` per deleted path, and `process` reaches it for every path that Lstat reports gone, including plain files. `ListFiles` filters with `LIKE 'prefix%'`, and SQLite only converts `LIKE` into an index range when the pattern is case-sensitive: the default `case_sensitive_like` is OFF and `rel_path` carries the default BINARY collation, so the optimizer cannot use the `rel_path` index. Each call is a table scan.

That is invisible for a single delete and it grows with the shape of the event burst. Removing a directory of 10k files on macOS emits a remove event per child, so a 100k-document corpus pays about 10k scans of 100k rows for one `rm -rf`. The watch queue is also the thing that #679 says silently drops events when it fills, so extra per-event latency here has a second-order cost.

Two ways out, if this shows up in practice:

1. A store method with a range predicate instead of `LIKE`, `rel_path >= 'docs/' AND rel_path < 'docs0'`, which the existing index does serve. `'0'` is the byte after `'/'`.
2. Coalesce the burst: run one subtree pass per deleted DIRECTORY rather than one per path.

I am not asking for either in this PR. The correctness guards are the right ones (byte-exact `HasPrefix` after the coarse prefix, plus a per-path Lstat re-check), the survivor assertions cover the `docsets` trap, and a store method is a wider change than this fix needs. Recording it so the cost is a known trade rather than a surprise.

Issue #690, opened 2026-08-03
mcp: open_file buffers the entire source before enforcing max_chars


yidakra:
## Summary

`dir2mcp_open_file` advertises a bounded response (`max_chars`, clamped to 50,000), but the raw-source path first reads the complete local file or S3 object into memory, converts the complete byte slice to a string, scans it, slices it, and only then truncates the result.

A source that grew after discovery/indexing—or a large remotely stored text/data file—can therefore consume unbounded memory and network bandwidth from a request whose output is strictly bounded.

## Evidence

- `internal/retrieval/service.go:1725-1732` clamps `max_chars` to 50,000.
- `internal/retrieval/service.go:1892-1922` documents and performs a full read:
  - object storage uses `io.ReadAll(rc)`;
  - local storage calls `readFileBounded(resolvedAbs, 0)`.
- `internal/retrieval/service.go:4102-4112` treats `maxBytes <= 0` as unbounded `io.ReadAll`.
- `internal/retrieval/service.go:2161-2178` converts the full `[]byte` to a full `string`, scans the entire string for secrets, applies span selection, and only then calls `truncateRunesWithFlag`.

The byte slice plus string conversion can transiently hold roughly two full copies, before rune conversion during truncation adds another allocation.

## Impact

- A single authenticated request can OOM the daemon with a large or concurrently growing file.
- S3 `open_file` downloads the full object even when the caller asks for a small excerpt.
- Line-range requests still buffer unrelated content before selecting the requested span.
- The risk bypasses ingest-time size policy because the source can change after indexing and `open_file` reads the live source.

This is a follow-up/regression around #407: the helper is named `readFileBounded`, but this call explicitly disables its bound. #682 covers ingestion reads; this path is retrieval/MCP request handling.

## Constraints

A safe fix cannot simply prefix-read `max_chars`: requested line spans may occur later in the file, UTF-8/rune boundaries must be preserved, and the existing secret policy scans the source before returning content. The implementation needs bounded/streaming selection and streaming secret detection (or a separately bounded policy with explicit semantics).

## Acceptance criteria

- `open_file` memory use is bounded independently of total source size for local and `CorpusFS` sources.
- Response semantics remain rune-based and never exceed normalized `max_chars`.
- Line-span selection can seek/stream to later lines without retaining the full prefix.
- Secret-exclusion behavior is preserved without retaining the full object in memory.
- Context cancellation interrupts long local and remote reads.
[...]

Issue #398, opened 2026-06-25
ingest: silent no-op drops — RepresentationStore seam, unhandled archive subtypes, sidecar-less video, binary parquet


yidakra:
**Part of #395.** Several ingest paths drop content with no error, no warning, no error-counter — the silent-failure class #395 targets. Grouped because the fix is the same shape: detect the unsupported/unhandled case and surface a clear diagnostic instead of a silent skip.

## 1 [HIGH] RepresentationStore assertion is a silent #364-style seam
`internal/ingest/service.go:327`: `if rs, ok := store.(model.RepresentationStore); ok { svc.repGen = NewRepresentationGenerator(rs) }` — **no else / no warn**. If the assertion fails, `svc.repGen` stays nil and `generateRepresentations` returns nil immediately for every document (`service.go:1938`). Result: zero raw-text/OCR/media/transcript representations → zero chunks → zero embeddings, corpus-wide, silently. This is the #364 failure mode one layer above the ChunkSource seam that #364 fixed — but without the loud-warn guard #364 added (`up.go:1067`). **Fix:** add the same loud warning + NDJSON event on the negative branch (it should be effectively impossible, so warn loudly if it ever happens).

## 2 [HIGH] Archive subtypes classified but never extracted
`classifyByExtension` (`internal/ingest/classify.go:45`) maps `.gz .tgz .bz2 .xz .7z .rar` (plus `.tar .zip`) to `archive`. But `archiveFormat` (`internal/ingest/archive.go`) only recognizes tar / tar.gz / tgz / tar.bz2 / zip — everything else returns `""`, `extractArchiveMembers` returns `nil,nil`, and the caller (`service.go:1423-1427`) treats nil error as success with no log. So a bare `.gz`/`.bz2`/`.xz`/`.7z`/`.rar` is silently ingested as an empty skipped document — known but unsearchable, zero diagnostics. **Fix:** support the common ones (bare gzip/bzip2/xz; 7z/rar if feasible) or emit a clear "unsupported archive format" outcome.

## 3 [HIGH] Sidecar-less video produces zero representations
[...]

PR #823, opened 2026-08-11
fix(ingest): scan the whole text and every derived text for secrets (#681)

### LINKED ISSUE #681: SECURITY: ingest secret scanning samples only 64 KiB and never scans OCR/STT-derived text
## Summary

The ingest exclusion policy scans only the first 64 KiB of source bytes and is applied only before representation generation. Secrets later in a text file, or secrets that become textual only through OCR/transcription/annotation/translation, can be embedded and exposed through search/ask.

This is a follow-up to #443, which improved the default regex vocabulary but did not make the scanning coverage complete.

## Evidence

- `internal/ingest/exclude.go:11` fixes `secretScanSampleBytes` at `64 * 1024`.
- `internal/ingest/service.go:2842-2847` returns only the leading sample for larger content.
- The only ingest calls to `hasSecretMatch` are the archive-member and top-level raw-byte checks (`service.go:2697`, `service.go:2791`).
- `generateRepresentations` and its OCR/STT/annotation/translation outputs are not scanned before representation/chunk persistence.

## Reproduction

1. Put a matching credential after byte 65,536 in an otherwise indexable text file; the document remains `ok` and the secret-bearing tail is chunked.
2. Put a visibly printed credential in a scanned PDF/image, or speak one in audio; the raw binary sample does not contain the extracted plaintext token, but OCR/STT output does and is indexed.

## Impact

Configured secret exclusions provide a false security boundary. Credentials can leak through MCP retrieval and generated answers even when their shape matches a configured/default pattern.

## Acceptance criteria

- Scan the complete indexable textual content with bounded-memory streaming/chunk overlap, not only the leading 64 KiB.
- Scan every derived textual representation before it is persisted/chunked/embedded.
- If a secret is found in any derived representation, prevent searchable output and persist an honest redacted skip/error state without leaking the match.
- Add tests for a match beyond 64 KiB and for OCR/STT-derived plaintext.

yidakra:
Closes #681.

## The exposure, plainly

Two kinds of credential are in the index today, on `main`, on a default configuration:

1. **A secret past byte 65,536 of a text file.** Ingest sampled the first 64 KiB of the source bytes and nothing else. Everything after that offset was chunked, embedded, and is returned by `search` and `ask`. A `.env` fragment pasted at the end of a long runbook, a key in the tail of a CSV export, an `AKIA...` line 300 KiB into a log: all indexed, all searchable.
2. **A secret that only exists in derived text.** Ingest scanned the SOURCE BYTES only. A credential printed on a scanned contract is pixels in the source and plain text only after OCR. A credential read out on a recorded call is audio in the source and plain text only after transcription. A credential inside a `.docx` is compressed in the source and plain text only after extraction. None of these was ever tested against `security.secret_patterns`. The same held for a translated transcript, a recognition statement, a subtitle sidecar's cues, and a stored annotation.

Configured secret exclusions were therefore a false boundary. `tests/ingest/secret_scan_coverage_681_test.go` demonstrates both cases against `main` (results below).

## The fix

**Scan what becomes searchable.** `secretScanBytes(docType, content)` splits on the classified doc_type:

* a raw-text doc_type (`text`, `code`, `md`, `data`, `html`) is chunked and embedded from these very bytes, so **every byte is scanned**. There is no offset past which a credential stops being indexed, so there must be none past which it stops being scanned.
* every other doc_type reaches the index only through a derived text. Its raw bytes are a container, so the 64 KiB head sample stays as a cheap early-out and the derived text carries the full scan.
* one case sits between the two: a binary payload that classified into a text-oriented doc_type, such as a `.parquet` reaching `data`. #398 already refuses to put such content on the raw-text path, so it is never chunked and never indexed. It takes the head sample too, because a full scan of it would be the whole cost of this change with none of the benefit. The binary check reads only the first 8000 bytes, so the decision itself is free.

**Screen every derived text before it is persisted.** `Service.screenDerivedSecrets` runs at each producer, before the title heuristic, before the quality gate, before the representation upsert:

| producer | where |
| --- | --- |
| OCR / image / PDF extraction | `generateOCRMarkdownRepresentation` |
| structured (docling) extraction, incl. the html route | `persistStructuredRepresentation` |
| pandoc extraction | `generatePandocMarkdownRepresentation` |
| STT transcript, per track | `transcribeAndPersistTrack` |
| translated transcript, per language | `translateOneTranscript` |
| recognition annotations | `recognize.go` |
| subtitle sidecar cues | `sidecar.go` |
| stored annotation | `StoreAnnotationRepresentations` |

The verdict is **document-wide**, not representation-wide. A document is one unit of disclosure: leaving its clean representations searchable next to a withheld one would still serve the text around the credential, and would report a partially indexed document as healthy. So the first derived match records the document `secret_excluded` and retires every representation this run already wrote for it, with its chunks (SPEC 6.6: the SQLite tombstone is what removes a vector from retrieval, now and after a restart).

**The verdict is sticky.** A raw-byte verdict is reproducible, because the next scan re-reads the same bytes. A derived verdict is not: the incremental gate is designed NOT to re-run OCR or STT for an unchanged file. Without a carry, the next scan would rebuild the document as `ok`, write that over the withheld row, and early-return on the unchanged-content gate, leaving a corpus that reports an indexed document whose representations are gone. `carrySecretExclusion` keeps the withheld status while the stored `content_hash` is settled and equal to the freshly computed one. An edited file, or a `--force` reindex (which clears the stored hash), re-derives and decides again, so removing the credential lets the document back in.

## SPEC: no dirstral-spec PR needed

I checked before writing code.

* SPEC 7.2 applies `security.secret_patterns` "to file **contents**". It sets no sample bound, so scanning more of the content is inside the contract.
* SPEC 15.2 closes the `skip_reasons` enum and already defines `secret_excluded`: "withheld because it matched secret-detection". A derived match is still secret-detection on the document's content, so it reuses that reason. No new reason, no new status, no schema change.
* SPEC 14.2 `FORBIDDEN` ("path/content blocked by policy") already exists for the one tool surface that needed a new answer (below).

The submodule pointer is untouched.

## The measured cost

Full scanning is a real change, so I measured it before committing to it. Benchmark: the shipped default `security.secret_patterns` (7 regexes), ordinary prose that matches none of them, Apple M4, Go 1.26.

| document size | head sample (`main`) | full scan (this PR) |
| --- | --- | --- |
| 64 KiB | 20 ms | 20 ms (identical: the file IS the sample) |
| 256 KiB | 22 ms | 89 ms |
| 1 MiB | 23 ms | 359 ms |
| 10 MiB (the default `ingest.max_file_mb` cap) | 32 ms | 3.63 s |

The scanner runs at about **2.9 MB/s**. That is not this change: it is Go's regexp engine on the default patterns, and `main` already pays 20 ms per document for its 64 KiB sample. I measured the alternatives before accepting it. Combining the 7 patterns into one alternation is not faster (2.66 MB/s: the combined automaton costs what the 7 passes cost). Per pattern, the two case-insensitive alternations (the AWS-secret heuristic and the JWT rule) account for 244 ms of the 359 ms; the patterns with a literal prefix run at 430-460 MB/s, because the engine can skip on the prefix.

**What I chose, and why it is proportionate.**

* The cost is bounded by the ingest file cap, which defaults to 10 MiB. The worst case is about 3.6 seconds for one document at the cap.
* It is paid **once per content version**, on the same path that already chunks and embeds those exact bytes. A 10 MiB text file produces roughly 2,500 chunks, each an embedding call. Minutes of provider work, seconds of scanning. An unchanged file on the next scan pays nothing: the content gate skips it.
* The doc_type split keeps it off the files where it would be pure waste. A 10 MiB video or a 10 MiB PDF never has its container bytes run through the regexes, which no pattern could usefully match anyway. Their derived text is scanned in full instead, and that text is much smaller than the media.

**What I chose NOT to do.** A literal prefilter (extract each pattern's required substrings with `regexp/syntax`, then skip the pattern when the input holds none of them) would reach about the 450 MB/s the prefixed patterns already show, a speedup of roughly 150x. I did not build it. A bug in the direction that matters there is a **silent false negative on a security control**, which is exactly the class of defect this issue is about. Correctness first. It is a good follow-up, gated on a differential fuzz test that asserts "if the pattern matches, the prefilter admits it".

## Agreement with PR #813 (`open_file`)

#813 relaxed the `open_file` secret check to "every byte the request reads is scanned: the window, everything before it, and a 64 KiB margin". Its justification named ingest's own 64 KiB sample: a document whose secret sits past 64 KiB "is already indexed, and its text is already reachable through `search`".

**That sentence is no longer true**, and I updated it rather than leave a stale argument in the code (`internal/retrieval/service.go` and `internal/retrieval/open_file_stream.go`, comments only, no behavior change). What the two halves still agree on is the property SPEC 15.4 actually requires: `open_file` never returns the secret. Every byte it returns was scanned and found clean, so the most a caller can obtain from a withheld document is a clean early window, never the credential and never the text behind it. Closing that last gap would need the whole-source read that #690 removed, so it is a deliberate trade, now stated as one.

The derived-text half of `open_file` has no gap at all: `openFileFromOCRCache` reads its whole cache file and scans all of it, so a withheld document's OCR or transcript text is already answered `FORBIDDEN`.

## `dir2mcp_annotate` is the one path that does NOT withhold

`StoreAnnotationRepresentations` screens the annotation and its flattened text, but it refuses the ANNOTATION without touching the source document. The annotation is model output ABOUT an already-scanned document, so a match says the model wrote a credential shape, not that the corpus file holds one. To retire a healthy document's representations on that evidence would be wrong.

The refusal needed an honest answer, because the tool otherwise returns `"stored": true` and echoes the annotation back to the caller. It now returns the canonical `FORBIDDEN` (SPEC 14.2) carrying `ingest.ErrSecretExcluded`, whose message contains no part of the matched text.

## No secret is ever logged

Every new log line names the document path, the producer label, and the fact that a configured pattern matched. No offset, no pattern, no payload. The withheld row clears `title` (a title lifted from a withheld document is derived text from it) and `error_message` (a 15.6 diagnostic surface).

## The withheld document must not also be an error

A media document carries a second verdict after its transcript: a video with no sidecar, no transcript, no media chunks, and no recognition is recorded `status="error"`, because it is known but unsearchable (#398/#495).

A withheld document reached that verdict by the same route, because its text was never persisted. It was then stamped `status="error"` on top, which overwrote the `secret_excluded` row and counted one document as both a skip and an error. A withheld document is unsearchable by design, not by failure, so it is neither. Every stage now stops as soon as the document is withheld: after extraction, after the sidecar, after the transcript, and after recognition. `tests/ingest/secret_scan_coverage_681_test.go:TestSecretScan_WithheldVideoIsNotAlsoBrandedAnError` pins it, and it fails on `main`.

The sidecar path was restructured for a related reason. Screening per language inside the persist loop let a clean English transcript commit, and count a representation, before a German sidecar withheld the document and retired that same transcript. The cues of every language are now chunked first and screened together, so the persist loop runs only for a document already known clean.

## Accounting

A withheld document counts as exactly one skip and never also as indexed, so `scanned = indexed + skipped + errors` still holds, and it raises exactly one `file_skip` event carrying `secret_excluded` (SPEC 3.2). Two cases are deliberately not counted, because both would push a run's totals past its scanned count: an archive member (members are never credited; the container accounts for the archive, matching `persistArchiveMemberSizeCapSkips`) and the two-phase derivation pass (`scanPass` counts the corpus on the first pass only). Both still write their own `secret_excluded` row, so the honest-coverage aggregate reports them.

## Tests

New: `tests/ingest/secret_scan_coverage_681_test.go`, 13 cases against a real SQLite store and complete scans.

Verified against `main` with the copy-aside method (`git stash` is banned in this repo, because `refs/stash` is shared across worktrees). Files were copied to a uniquely named scratch directory, reverted with `git checkout HEAD --`, the tests run, then restored. Results on `main`:

| test | on `main` |
| --- | --- |
| `PastTheHeadSample_IsWithheld` | **FAIL**: `status = "ok"`, 1 live `raw_text` representation |
| `OnlyInExtractedText_WithholdsTheDocument` | **FAIL**: `status = "ok"`, 1 live `extracted_markdown` |
| `OnlyInTranscript_WithholdsTheDocument` | **FAIL**: `status = "ok"`, 1 live `transcript` |
| `OnlyInTranslation_RetiresTheCleanTranscript` | **FAIL**: `status = "ok"`, 2 live representations |
| `OnlyInSubtitleSidecar_WithholdsTheMedia` | **FAIL**: `status = "ok"`, 1 live `transcript` |
| `DerivedVerdictSurvivesTheNextScan` | **FAIL** |
| `RemovingTheSecretLetsTheDocumentBackIn` | **FAIL** |
| `DerivedExclusionCountsAsSkipNotIndexed` | **FAIL**: counters `scanned:1 indexed:1 skipped:0`, want `1/0/1/0` |
| `DerivedExclusionEmitsOneFileSkip` | **FAIL**: no `file_skip` raised |
| `WithheldVideoIsNotAlsoBrandedAnError` | **FAIL**: `status = "ok"`, indexed 1, 1 live `transcript` |
| `WithinTheHeadSample_StaysWithheld` | **PASS** (preservation, as it must) |
| `CleanLargeFile_StaysIndexed` | **PASS** (false-positive guard on the wider scan, as it must) |
| `BinaryPayloadInATextDocType_IsNotIndexed` | **PASS** (pins the #398 outcome that the head-sample carve-out relies on) |

`AnnotationIsRefusedWithoutTouchingTheDocument` does not compile on `main`, because it names the new sentinel, so it was left out of that run.

Every test uses the shipped default patterns rather than one invented for the test, and a synthetic `AKIA...` key that is not a real credential.

## Not in scope

* **Multimodal media chunks** (`rep_type=media`). They embed source bytes with no text stage, so there is nothing to run a text regex over. A credential visible only in a video frame under `embed.multimodal=replace` is covered by no text policy.
* **Document summaries.** A summary is derived from chunks that were already screened, so it inherits a clean input. To screen it again would only catch a model that invents a credential shape.
* **The literal prefilter** described under "The measured cost".
* **The clean-early-window gap in `open_file`** described under "Agreement with PR #813".

## Checklist

- [x] `make check` passes locally, including `gocyclo -over 15`. Four functions went over the gate and were fixed by extraction, not by working around it: `settleProcessedDocument`, `buildArchiveMemberDocument`, `shiftSegmentsForLeadingSilence`, and `screenAnnotationForSecrets` plus `annotationPreview`.
- [x] new behavior has test coverage that fails against `main`
- [x] no secret, matched payload, or pattern is logged or persisted
- [x] `README.md` and `docs/` remain truthful (neither documents the scan coverage)
- [x] `coderabbit review --committed --base main`: 0 findings
- [x] no unrelated files changed, submodule pointer untouched

## CI note

The first CI run failed the `race` job on `tests/cli` `TestReindex_AfterCrash_RepeatedCrashesDoNotRotatePartialState`, with no data race reported. That test does not touch ingest or the secret policy. It passes locally on this branch under `-race`, alone and as a whole suite, and the job passed on the next run. `tests/cli` produced a second, different flake locally (`TestUpPublicAuthNoneAllowedWithForceInsecure`), which also passed alone and in a full re-run. Both look like load-sensitive flakes in that suite rather than a regression here.

### DIFF
--- a/internal/ingest/exclude.go
+++ b/internal/ingest/exclude.go
@@ -39,8 +39,14 @@ const secretScanSampleBytes int64 = 64 * 1024
 // The doc_type split is what keeps the cost proportionate: it never runs the
 // regex set over the bytes of a 10 MiB video, which no pattern could usefully
 // match anyway.
+//
+// One case sits between the two: a binary payload that classified into a
+// text-oriented doc_type, such as a .parquet reaching "data". #398 already
+// refuses to put such content on the raw-text path, so it is never chunked and
+// never indexed. It therefore takes the head sample too: a full scan of it would
+// be the whole cost of the change with none of the benefit.
 func secretScanBytes(docType string, content []byte) []byte {
-	if ShouldGenerateRawText(docType) {
+	if ShouldGenerateRawText(docType) && !looksLikeBinaryContent(content) {
 		return content
 	}
 	return contentSample(content)
--- a/internal/ingest/exclude.go
+++ b/internal/ingest/exclude.go
@@ -8,7 +8,43 @@ import (
 	"strings"
 )
 
+// secretScanSampleBytes is the head-sample budget for a source whose searchable
+// text is NOT its raw bytes: a PDF, an image, an office document, audio, video.
+// Scanning a container's compressed or encoded bytes past the head buys nothing,
+// because a credential inside such a file only becomes readable text after
+// extraction, OCR, or transcription. Those DERIVED texts are scanned in full
+// before they are persisted (see Service.screenDerivedSecrets), so the head
+// sample is a cheap early-out, not the security boundary.
+//
+// A source whose raw bytes ARE the searchable text (text, code, markdown, data,
+// html) is scanned in FULL instead; see secretScanBytes. #681: the head sample
+// used to be the only scan for every source, so a credential past 64 KiB was
+// indexed and reachable through `search`.
 const secretScanSampleBytes int64 = 64 * 1024
+
+// secretScanBytes returns the bytes of a source document that the ingest secret
+// policy scans, given the document's classified doc_type.
+//
+// The rule is "scan what becomes searchable":
+//
+//   - a raw-text doc_type (text/code/md/data/html) is chunked and embedded from
+//     these very bytes, so every byte is scanned. There is no offset past which
+//     a credential stops being indexed, so there must be none past which it
+//     stops being scanned.
+//   - every other doc_type reaches the index only through a derived text
+//     (extraction, OCR, transcription, annotation, recognition, a subtitle
+//     sidecar). Its raw bytes are a container, so only the head sample is
+//     scanned here and the derived text carries the full scan.
+//
+// The doc_type split is what keeps the cost proportionate: it never runs the
+// regex set over the bytes of a 10 MiB video, which no pattern could usefully
+// match anyway.
+func secretScanBytes(docType string, content []byte) []byte {
+	if ShouldGenerateRawText(docType) {
+		return content
+	}
+	return contentSample(content)
+}
 
 func compileSecretPatterns(patterns []string) ([]*regexp.Regexp, error) {
 	return CompileSecretPatterns(patterns)
--- a/internal/ingest/recognize.go
+++ b/internal/ingest/recognize.go
@@ -402,6 +402,15 @@ func (s *Service) generateRecognitionRepresentation(ctx context.Context, doc mod
 		return false, nil
 	}
 
+	// #681: recognition statements are text this daemon derives from the media and
+	// then indexes, so they are screened before persistence. The scan runs over the
+	// segment texts exactly as they will be chunked, not over hashInput, which
+	// interleaves length prefixes for a different purpose. On a match the document
+	// is already withheld, so this reports "nothing recognized".
+	if s.screenDerivedSecrets(ctx, doc, derivedKindRecognition, joinSegmentTexts(segments)) {
+		return false, nil
+	}
+
 	meta, err := json.Marshal(recognitionMeta{
 		Source:       recognizeSource,
 		Provider:     s.cfg.RecognizeProvider,
--- a/internal/ingest/secret_derived.go
+++ b/internal/ingest/secret_derived.go
@@ -1,0 +1,264 @@
+package ingest
+
+import (
+	"context"
+	"errors"
+	"os"
+	"regexp"
+
+	"github.com/dirstral/dir2mcp/internal/model"
+)
+
+// Derived-text secret screening (dir2mcp #681).
+//
+// SPEC §7.2 applies security.secret_patterns to file CONTENTS, and §15.2 records
+// the outcome as the closed skip_reason `secret_excluded`. Ingest honored that
+// for one kind of content only: the raw bytes of the source file. A credential
+// that reaches the index through a DERIVED text was never tested against the
+// patterns at all, so it was chunked, embedded, and served by `search` and `ask`.
+//
+// Three ordinary corpora hit this:
+//
+//   - a scanned contract or a screenshot: the key is pixels in the source and
+//     plain text only after OCR;
+//   - a recorded standup or a support call: the key is audio in the source and
+//     plain text only after transcription;
+//   - a .docx or .pdf: the key is inside a compressed container in the source and
+//     plain text only after extraction.
+//
+// The same applies to a translated transcript, a recognition statement, a
+// subtitle sidecar's cues, and a stored annotation. Every one of them is text
+// this daemon creates and then makes searchable, so every one of them is scanned
+// here BEFORE it is persisted.
+//
+// The verdict is document-wide, not representation-wide. A document is a single
+// unit of disclosure: leaving its clean representations searchable next to a
+// withheld one would still let a caller pull the surrounding context of a
+// credential, and would report a partially indexed document as healthy. So the
+// first derived match withholds the whole document as `secret_excluded` and
+// retires anything this run already wrote for it.
+//
+// Nothing here logs, persists, or returns the matched text, the matching pattern,
+// or an offset. The operator learns WHICH document was withheld and that a
+// configured pattern matched; the payload never leaves the process.
+
+// Derived-text kinds, used only for the operator-facing log line and to name the
+// producer in review. They are diagnostic labels, not a persisted vocabulary:
+// SPEC §15.2 closes the skip_reasons enum, and every one of these withholds the
+// document under the SAME `secret_excluded` reason the raw-byte path uses.
+const (
+	derivedKindExtraction  = "extraction"
+	derivedKindTranscript  = "transcript"
+	derivedKindTranslation = "translation"
+	derivedKindRecognition = "recognition"
+	derivedKindSidecar     = "subtitle sidecar"
+)
+
+// ErrSecretExcluded reports that derived text was withheld because it matched a
+// configured security.secret_patterns entry. It carries no part of the text, the
+// pattern, or the offset, so it is safe to surface verbatim on a tool or
+// diagnostic surface. `dir2mcp_annotate` maps it to the canonical §14.2
+// FORBIDDEN ("path/content blocked by policy").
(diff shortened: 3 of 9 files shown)


yidakra:
@coderabbitai review

yidakra:
@coderabbitai review

Issue #487, opened 2026-07-02
s3: remaining discovery-side gaps (gitignore no-op, nil Content-Length streaming, region/endpoint validation)


yidakra:
Follow-up to #432. PR #479 fixed part 1 (open_file retrieval read path over CorpusFS for S3). Remaining discovery/ingest-side:
- Part 2 — .gitignore is a no-op on S3 discovery.
- Part 3 — nil Content-Length object streaming.
- Part 4 — source.kind: s3 region/endpoint validation (CONFIG_INVALID).

Issue #682, opened 2026-08-03
ingest: size cap is TOCTOU-only; content reads and S3 localization are unbounded


yidakra:
## Summary

`ingest.max_file_mb` is enforced during discovery, but the actual content reads do not enforce the discovered size. If a local file or remote object changes between discovery and read, dir2mcp reads the new object without a byte limit. S3 media/archive localization is likewise an unbounded download. The configured cap therefore does not provide a hard memory/disk resource bound.

This is a follow-up to #408 and #497: fan-out and visible discovery skips were fixed, but the byte cap is still time-of-check/time-of-use only.

## Evidence

- `buildDocumentWithContent` copies `f.SizeBytes` into metadata, then calls unbounded `io.ReadAll(rc)` (`internal/ingest/service.go:2744-2767`). It neither limits to the configured cap nor verifies the number of bytes against the discovery result.
- Other derivation reads use unbounded `io.ReadAll` (`service.go:2450-2460`).
- `S3FS.Localize` streams the complete GET via unbounded `io.Copy` into a temp file (`internal/corpusfs/s3.go:374-408`).
- Discovery/list/HEAD and later Open/GetObject are separate operations; a mutable source can change size or identity between them.

## Impact

A file that was under the cap when discovered can grow arbitrarily before processing, causing high memory use or OOM. A changed S3 object can consume unbounded cache disk during media/archive localization. Persisted size/ETag/content metadata may also describe a different snapshot from the bytes actually processed.

## Acceptance criteria

- Enforce the configured maximum at every source-byte read/localization boundary using a `limit+1` check.
- Reject/persist `size_cap` honestly if the actual read exceeds the cap.
- Detect source identity changes between discovery and read and retry/reconcile instead of combining stale metadata with new bytes.
- Clean up partial localized files on every over-limit/error path.
- Add local growth-race and S3 changed-object tests.

PR #832, opened 2026-08-12
fix(corpusfs): refuse an S3 range response that serves the wrong bytes (#673)

### LINKED ISSUE #673: corpusfs: known-size S3 reader reopens after Close and trusts ignored or truncated range responses
## Summary

The known-size S3 range reader has weaker lifecycle and integrity checks than the unknown-size stream reader. It can reopen after `Close`, and it accepts a server that ignores or truncates a range request, returning wrong bytes as if they were valid.

## Evidence

`internal/corpusfs/s3reader.go`:

- `s3RangeReader` has no `closed` state. `Close` sets `body=nil`, so the next `Read` issues a new GET.
- `openFrom` sends `Range: bytes=N-` but never validates `Content-Range`/response offset.
- `Read` returns EOF from the body without checking whether `offset` reached the HEAD-reported `size`.

The unknown-size `s3StreamReader` already rejects read/seek after close, and its regression test covers that behavior. The known-size test stub always honors ranges and omits response metadata, so ignored/truncated ranges are untested.

## Impact

With an S3-compatible endpoint/proxy that ignores Range, seeking to offset N returns bytes from offset 0. A prematurely truncated body can also be accepted as a complete object. Consumers may hash, parse, or cite corrupted bytes.

## Expected

- Closed readers stay closed and return `fs.ErrClosed`.
- Ranged responses are validated against the requested offset.
- EOF before the advertised object size is surfaced as an integrity/read error.

## Acceptance criteria

- Tests cover use-after-close for both reader variants.
- A fake endpoint that ignores `Range` is rejected rather than returning wrong bytes.
- Premature EOF relative to known size is reported.
- Correct ranged GET behavior remains lazy and seekable.

yidakra:
Closes #673.

## What a user saw

`search` returned a passage that is not in the document at that place. A citation pointed at a file whose bytes do not contain the quoted text. `open_file` on the same document showed different content from the hit that cited it.

Nothing reported an error. The scan said `indexed`, `errors=0`, and the coverage report was clean. The document was stored with a `content_hash` over bytes the corpus does not actually have, so the next scan agreed with itself and never re-read the file.

Two ordinary object stores produce that result today.

1. **A proxy or gateway that drops the Range header.** The reader seeks to byte N, asks for `bytes=N-`, and receives the WHOLE object with a 200. The bytes at byte 0 are then read as the bytes at byte N.
2. **A body that ends early.** The bytes that arrived are a PREFIX of the document. They arrived as `io.EOF`, which every caller reads as the end of the file.

Both are silent. The wrong bytes are hashed, chunked, embedded and cited, and no log line names the file.

## The mechanism

`s3RangeReader` asked for a range and then trusted the answer. It never looked at the response, and it never compared what it received against the size `HeadObject` reported.

### 1. A ranged response must state the offset it served

`openFrom` now checks the response before it reads a byte of it, and closes the body when the check fails.

The only accepted evidence is a well formed `Content-Range` that states the requested first byte position. RFC 9110 §14.4 requires a store to send it on a 206, so the rule costs a conforming store nothing. A store that drops the Range header answers with a 200 and states no range at all, which is exactly the case that must fail.

The whole header value is validated, not only the part that is read: the `bytes` unit, a numeric first and last position, a last position not before the first, and a complete length that is numeric or `*`. A store that sends `bytes 32-x/64` has not stated where its bytes start, and reading its leading digits would trust the one response that already proves the store unreliable. A stated numeric complete length must also equal the size `HeadObject` reported, because a different total describes a different object.

**A response length is not accepted as a substitute.** `Content-Length` states how many bytes arrive, never where they start. An object replaced between the HEAD and the GET can serve a full body whose length equals the requested tail, so a length check passes while the store serves byte 0. The one response without a `Content-Range` that is still accepted is a read at offset 0, where a whole object body IS the requested range. That keeps every existing whole file read working.

### 2. A short body is not `io.EOF`

`Read` now ends through `endOfBody`. The end of a GET body is `io.EOF` only when the reader delivered every byte the store said the object has. A body that stops earlier delivered a prefix, and it fails with `ErrObjectTruncated`.

This follows the precedent #829 set for the neighbouring case rather than a new policy: over cap delivery ends with `ErrObjectTooLarge`, never with `io.EOF`, because a silent short read looks to every caller like a complete file, which is the #487 truncation bug. The cap is checked FIRST in `endOfBody`, so an object past the configured cap keeps the single answer #829 gave it and this PR does not weaken that bound.

### 3. A closed reader stays closed

`Close` now marks the reader closed. `Read` and `Seek` after `Close` return `fs.ErrClosed`, and no new GET is issued. Double `Close` is a no op.

`Close` used to only drop the body, so the next `Read` opened a SECOND ranged GET on a reader the caller had already released. That request runs on a context the caller may have cancelled, and a store that ignores its Range answers it from byte 0 while the caller believes it continues at its saved offset. `s3StreamReader` already behaved this way. Both variants now agree, and the test file pins the rule on both.

### Two sentinels, because callers need different answers

* `ErrRangeNotHonored` is a property of the SERVER. The same object read without a seek is fine, so an operator learns the backend is unusable for range reads, not that the document is broken.
* `ErrObjectTruncated` is a property of this READ. A retry may deliver the whole object.

Neither message carries a byte of the object. They name the bucket, the key, the offsets and the sizes, so they are safe to log. `TestS3RangeReader_RefusalNamesTheObjectNotItsBytes` pins that.

## SPEC: no dirstral-spec PR needed

I checked before I wrote code.

SPEC §7.10 defines the `open / range-read` capability: a CorpusFS must "support random-access range reads (read *N* bytes at offset *O*)", and "on `s3` a range read maps to a ranged `GET`". It also states that `content_hash` is computed over "the bytes returned by open, identically across backends, so document identity is backend-independent".

A range read that returns the bytes at offset 0 does not satisfy that contract, and a hash over a truncated body breaks the identity invariant §7.8 depends on. This PR enforces what the SPEC already requires. It adds no new capability, no new config surface, no new error code, and no new skip reason. A refused read surfaces through the read error path that a corrupt or unreadable object already used. The submodule pointer is untouched.

## Verified against `main`

`git stash` is banned in this repository, because `refs/stash` is shared across worktrees. The files were copied to a uniquely named scratch directory, the tracked files were reverted with `git checkout HEAD --`, the tests were run, and everything was restored afterwards.

The new sentinels do not exist on `main`, so the file cannot compile there. The proof run used the same tests with each `errors.Is(err, corpusfs.ErrX)` assertion replaced by `err == nil`. That is a WEAKER assertion, so a pass would have been possible, and `main` still fails all four.

| test | on `main` |
| --- | --- |
| `S3RangeReader_RefusesAServerThatIgnoresTheRange` | **FAIL**: `read error = <nil>`, 32 bytes delivered from offset 0 for a request that asked for offset 32 |
| `S3RangeReader_RefusesARangeServedFromTheWrongOffset` | **FAIL**: `read error = <nil>` |
| `S3RangeReader_ShortBodyIsNotEOF` (both subtests) | **FAIL**: `read error = <nil>` after 100 of 4096 bytes, at offset 0 and after a seek |
| `S3RangeReader_StaysClosed` | **FAIL**: `Read after Close = <nil>`, `Seek after Close = <nil>`, 1 GetObject after Close |
| `S3RangeReader_RefusalNamesTheObjectNotItsBytes` | **FAIL**: `read succeeded, want a refusal` |
| `S3StreamReader_StaysClosed` | **PASS** (the streaming reader was already correct, as it must be) |
| `S3RangeReader_HonestServerRoundTripsARangedRead` | **PASS** (false positive guard, as it must be) |

The `RefusesAMalformedContentRange` and `AcceptsAnUnknownCompleteLength` cases came out of the review round. They pin the strict parser, so they were written against the FIRST commit on this branch, where the length only fallback accepted the malformed forms.

## Tests

`tests/corpusfs/s3_range_integrity_673_test.go` (17 cases). The stub is network free and takes no credentials and no endpoint. Its behaviour is selectable per test: honor the range, drop it, state the wrong offset, state a literal header, or serve a short body. Every body byte names its own offset, so a test proves WHICH bytes it received, not only how many.

`tests/corpusfs/s3_test.go`: the existing `fakeS3` honored ranges but stated no response metadata, which is why an ignored range was untested (#673 names this). It now sets `Content-Range` and `Content-Length` on a ranged answer, exactly as a real store does on a 206.

## Review round

`coderabbit review --committed --base main` returned 3 findings. All 3 were valid and all 3 are fixed in the second commit: the length only acceptance is removed, the parser validates the complete syntax, and a stated complete length is compared with the size HEAD reported. Each fix has a test.

## Not in scope

* **`Localize`** issues an unranged GET and never calls HEAD, so it has no reported size to compare a short download against. A truncation check there needs a HEAD call per download, which is a separate cost decision.
* **A `Content-Range` total that disagrees with HEAD** is refused as `ErrRangeNotHonored`. The cause may be an object that changed rather than a store that ignores ranges. A distinct "object changed under the reader" sentinel would be more precise, and it belongs with the ETag re-read work in §7.8, not here.
* **`s3StreamReader`** gets no truncation check, because a HEAD that omits `Content-Length` gives it no size to compare against. That is why it exists (#487).
* **No new MCP error code.** On the ingest path a refused read becomes a `status="error"` row with the message, which is the visible and correct answer. On `open_file` it joins the class every mid-read failure already falls into (`mapReadDocumentError` default). SPEC §14.4 names no code for "the source served the wrong bytes", and this repository is spec-first, so a new code needs a dirstral-spec PR first. An honest error under a general code is still a large improvement on silently wrong content.

## Checklist

- [x] `coderabbit review --committed --base main`: 3 findings, all fixed
- [x] `make check` passes locally, including `gocyclo -over 15`
- [x] new behaviour has test coverage that fails against `main`
- [x] no object content is logged or persisted
- [x] `README.md` and `docs/` remain truthful
- [x] no unrelated files changed, submodule pointer untouched

### DIFF
--- a/internal/corpusfs/s3reader.go
+++ b/internal/corpusfs/s3reader.go
@@ -6,10 +6,37 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"strconv"
+	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 )
+
+// ErrRangeNotHonored is returned when an object store answers a ranged GET with
+// bytes that do not start at the requested offset (#673).
+//
+// The condition is not exotic. An S3-compatible proxy or gateway that drops the
+// Range header answers with the WHOLE object, so a reader that seeked to offset N
+// receives the bytes at offset 0 and cannot tell. Those bytes are then hashed,
+// chunked and cited as if they came from N, so the corpus indexes content the
+// source document does not have at that position.
+//
+// It is a distinct sentinel because it is a property of the SERVER, not of the
+// object: the same object read without a seek is fine, so a caller may report the
+// backend as unusable for range reads instead of the document as broken.
+var ErrRangeNotHonored = errors.New("corpusfs: object store did not honor the requested byte range")
+
+// ErrObjectTruncated is returned when an object body ends before the size the
+// store itself reported (#673).
+//
+// The bytes that did arrive are a PREFIX of the document. Returned as io.EOF they
+// are indistinguishable from a complete read, so the pipeline stores a short
+// document, a content hash over the wrong bytes, and citations that do not match
+// the source. That silent short read is the #487 truncation failure, and #682
+// established the rule this follows: a read that did not deliver the object fails
+// with its own error and never with io.EOF.
+var ErrObjectTruncated = errors.New("corpusfs: object body ended before the reported size")
 
 // s3RangeReader is an io.ReadSeekCloser over an S3 object that satisfies reads
 // via ranged GETs. It tracks a logical offset and lazily opens a GET body on the
@@ -29,6 +56,7 @@ type s3RangeReader struct {
 
 	offset int64
 	body   io.ReadCloser
+	closed bool
 }
 
 // Read fulfills io.Reader, opening a ranged GET from the current offset on
@@ -44,7 +72,16 @@ type s3RangeReader struct {
 // repository uses. The reader delivers one byte more than the cap and then fails,
 // because that extra byte is what separates an object of exactly the cap (which is
 // inside the policy and must read cleanly to io.EOF) from an object past it.
+//
+// A read after Close returns fs.ErrClosed. Close leaves the logical offset where
+// it was, so a reopen would issue a fresh ranged GET on a reader the caller has
+// already released: the context may be cancelled, and if the store ignores the
+// Range the reopened body serves offset 0 while the caller believes it is at
+// r.offset (#673).
 func (r *s3RangeReader) Read(p []byte) (int, error) {
+	if r.closed {
+		return 0, fs.ErrClosed
+	}
 	if r.offset > r.maxBytes {
 		return 0, fmt.Errorf("corpusfs: read s3://%s/%s: %w (cap %d bytes)", r.bucket, r.key, ErrObjectTooLarge, r.maxBytes)
 	}
@@ -64,10 +101,32 @@ func (r *s3RangeReader) Read(p []byte) (int, error) {
 	}
 	n, err := r.body.Read(p)
 	r.offset += int64(n)
+	if errors.Is(err, io.EOF) {
+		return n, r.endOfBody()
+	}
 	return n, err
 }
 
-// openFrom starts a ranged GET beginning at off through the end of the object.
+// endOfBody decides what the end of a GET body means at the current offset.
+//
+// It is io.EOF only when the reader delivered every byte the store said the
+// object has. A body that stops earlier delivered a PREFIX, and reporting that
+// prefix as io.EOF is the silent truncation #487 fixed once already, so it fails
+// with ErrObjectTruncated instead. The cap is checked first so an object past the
+// configured limit keeps the single answer #682 gave it.
+func (r *s3RangeReader) endOfBody() error {
+	if r.offset > r.maxBytes {
+		return fmt.Errorf("corpusfs: read s3://%s/%s: %w (cap %d bytes)", r.bucket, r.key, ErrObjectTooLarge, r.maxBytes)
+	}
+	if r.offset >= r.size {
+		return io.EOF
+	}
+	return fmt.Errorf("corpusfs: read s3://%s/%s: %w (got %d of %d bytes)",
+		r.bucket, r.key, ErrObjectTruncated, r.offset, r.size)
+}
+
+// openFrom starts a ranged GET beginning at off through the end of the object,
+// and refuses a response that does not start there (#673).
 func (r *s3RangeReader) openFrom(off int64) error {
 	if off >= r.size {
 		return io.EOF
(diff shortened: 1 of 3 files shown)


yidakra:
Line-reviewed this myself rather than wait for the CodeRabbit window, and independently reproduced the defect on main.

## Verified on main

I checked out `origin/main` in a separate worktree, copied in the test file alone, and weakened only the sentinel assertions to `err == nil` (main has no sentinels to compare against). Main returns NO error in every case:

```
RefusesAServerThatIgnoresTheRange: read error = <nil> (got 32 bytes)
RefusesARangeServedFromTheWrongOffset: read error = <nil>
ShortBodyIsNotEOF/from_the_start: read error = <nil>, after 100 of 4096 bytes
ShortBodyIsNotEOF/after_a_seek: read error = <nil>, after 100 of 4096 bytes
StaysClosed: Read after Close = <nil>; 1 GetObject call after Close
RefusesARangeStatedOnlyByItsLength: read error = <nil>
```

The first line is the one that matters: a seek to byte 32 returned the 32 bytes at offset 0, silently. Those bytes get hashed, chunked and cited as if they came from 32.

## The design calls I checked

**Content-Length is rejected as evidence.** Right, and it is the finding your first review round caught. A length states how many bytes arrive, never where they start, so an object replaced between the HEAD and the GET can produce a full body whose length happens to equal the requested tail.

**An absent Content-Range is accepted only at offset 0.** Correct: a whole-object body IS the requested range there, so refusing it would break every non-seeking read against a store that omits the header.

**A malformed header is an error, never an absent one.** This is the subtle one and it is right. Treating `bytes 32-x/64` as absent would fall through to the `off == 0` acceptance, so the most broken response would get the most trusting path.

**`end` is parsed but not compared with the object size.** Correct per RFC 9110: a store may legally answer an open-ended `bytes=N-` with a shorter range, and a body that then ends early is caught by `endOfBody` instead. Good separation.

**`endOfBody` checks the cap first.** This preserves #682's single answer for an over-cap object rather than reclassifying it as truncated. Worth keeping when this file changes again.

**`refuseRange` names offsets and sizes only.** No object bytes in an error, so the message is safe to log.

## One question, not a blocker

`Seek` after `Close` returns `fs.ErrClosed`. `io.Seeker` does not require that, and `os.File` does the same, so I am not asking for a change. But it is a slightly stronger contract than `Read`-after-`Close` needs, and the comment already explains it. Fine as is.

Still worth the CodeRabbit pass before merge, since round one found three real problems. I am not blocking on it.

yidakra:
@coderabbitai review

yidakra:
CodeRabbit pass is done.

`coderabbit review --committed --base main` on the final commit: **No new findings**. The 3 findings from round one stay closed, and each carries a test:

1. the length-only acceptance is removed, so `Content-Length` is never evidence of a start;
2. the parser validates the complete `Content-Range` syntax, so `bytes 32-x/64` is a refusal and not a trusted prefix;
3. a stated numeric complete length must equal the size `HeadObject` reported.

The GitHub bot could not run: the account review quota is still spent, and it answers `@coderabbitai review` with "Review rate limited". It posted no inline comments, so there are no open threads on this PR.

Thank you for the independent reproduction on `main`. Your run and mine agree line for line.

On the `Seek`-after-`Close` question: I kept `fs.ErrClosed` for it. `io.Seeker` does not require the error, as you say. The reason to keep it is that a seek on this reader only sets up the next `Read`, and that `Read` must fail. A silent seek would let a caller believe it moved a reader that cannot move, which is the same class of quiet wrongness the PR removes. `os.File` answers the same way, so the contract is not novel.

Issue #711, opened 2026-08-03
elevenlabs bridge: upstream MCP responses are buffered without a size limit


yidakra:
## Summary

The `elevenlabs-bridge` MCP client buffers every upstream initialize and `tools/call` response with an unbounded `io.ReadAll`. A compromised, buggy, or misconfigured `MCP_URL` can exhaust the bridge process memory before JSON decoding or HTTP error handling runs.

This is a separate implementation from the legacy CLI remote shims tracked in #704.

## Evidence

`internal/elevenlabsbridge/client.go` `doRPC`:

1. sends the request with a 45-second `http.Client` timeout;
2. defers response body close;
3. calls `io.ReadAll(resp.Body)` without checking `Content-Length` or using `io.LimitReader`;
4. returns the full bytes for every 2xx and non-2xx status.

The path is shared by session initialization and all bridge routes (`/ask`, `/search`, `/list_files`, `/stats`). Non-2xx bodies are stored in `HTTPError` and can also be copied verbatim to the downstream response.

The bridge limits inbound request bodies to 1 MiB, but has no corresponding upstream response bound.

## Impact

Any reachable endpoint configured as `MCP_URL` can drive unbounded allocation. A normal but unexpectedly large `list_files` or search response can also consume excessive memory and be duplicated across the raw body, JSON envelope, decoded structured result, and downstream encoding.

## Expected behavior

Use a shared bounded response reader for initialize and tool calls, including error bodies. Reject responses beyond a documented cap with a concise `502`-class error that does not echo upstream content.

## Acceptance criteria

- Oversized 2xx initialize and tool responses fail before full buffering/decoding.
- Oversized non-2xx bodies are bounded and never proxied in full.
- Within-limit responses retain current behavior.
- Tests cover declared and streaming/chunked oversized bodies.

Issue #704, opened 2026-08-03
cli: remote MCP shims buffer initialize and tool responses without a size limit


yidakra:
## Summary

The legacy `ask`, `search`, `open-file`, and `list-files` remote shims read every MCP initialize and tool-call response with an unbounded `io.ReadAll`. A compromised, buggy, or simply misconfigured endpoint from `connection.json` can exhaust the CLI process's memory before JSON decoding or error handling runs.

## Evidence

`internal/cli/remote_commands.go:130-140`:

1. sends the request with a 45-second HTTP client;
2. defers body close;
3. calls `io.ReadAll(resp.Body)` for every status and content type.

There is no `Content-Length` check or `io.LimitReader`. The same path handles both `initialize` and `tools/call`, including non-2xx error bodies. Even `open-file --max-chars` does not protect the client because that bound is only a request argument which an untrusted server can ignore.

This is distinct from #670, which covers provider adapters rather than the MCP client.

## Expected

Apply a finite response-body bound before buffering, with a clear error when exceeded. The cap should accommodate legitimate bounded tool results while preventing an endpoint from driving unbounded allocation.

## Acceptance criteria

- Both initialize and tools/call responses use a shared bounded reader.
- Oversized 2xx and non-2xx response tests fail cleanly without decoding the full body.
- Within-limit responses preserve current behavior.
- The returned error identifies the response-size limit without including response content or credentials.
- Redirect/auth behavior is reviewed while centralizing the HTTP path so connection headers cannot create a separate credential-leak regression.

PR #833, opened 2026-08-12
fix(providers): route every adapter through one hardened HTTP path (#670)

### LINKED ISSUE #670: providers: newer adapters regress redirect-secret and successful-response size hardening from #416
## Summary

Follow-up to #416: provider clients added or left outside that fix still decode successful responses without a size bound, and custom API-key headers can follow redirects to a different origin.

## Evidence

Successful JSON is decoded directly from `resp.Body` with `json.NewDecoder` in:

- `internal/anthropic/client.go`
- `internal/cohere/client.go` (embed, rerank, and chat)
- `internal/colbertrerank/client.go`
- `internal/whisperapi/client.go`

These paths have no equivalent of the bounded-success-body helpers used by hardened clients, so a compromised/misconfigured endpoint can return an unbounded 2xx body and exhaust memory.

Separately:

- Anthropic sends `x-api-key`.
- ElevenLabs sends `xi-api-key`.
- Their default/custom `http.Client` paths do not install a `CheckRedirect` policy.

Go's redirect handling strips `Authorization` and cookies across untrusted origins, but custom key headers are not protected by that rule. A redirect can therefore forward these credentials to another host. Anthropic's timeout-cloning helper also preserves no explicit redirect guard.

## Expected

All provider adapters should share the same bounded-read and redirect policy:

- cap successful response bodies before JSON decode;
- reject cross-origin redirects (or explicitly strip every credential-bearing custom header);
- preserve custom transports/timeouts without dropping the policy.

## Acceptance criteria

- Oversized 2xx response tests cover Anthropic, Cohere, ColBERT, and Whisper API.
- Cross-origin redirect tests prove `x-api-key` and `xi-api-key` never reach the redirect target.
- The fix reuses a common hardened client/body helper where practical.
- Existing provider-specific error-body caps remain intact.

yidakra:
## Summary

Issue #416 hardened the provider HTTP clients. It refused redirects that carry a
custom API-key header, and it capped a successful response body before the JSON
decode. That fix was copied into each adapter package. Adapters written later
were written without it, so the hardening regressed (#670).

This PR removes the duplication. `internal/providerhttp` now holds the one
hardened path. All nine adapters use it.

## The exposure

**1. A redirect carries the API key to another host.**

Go strips `Authorization`, `Www-Authenticate`, `Cookie` and `Cookie2` only when
a redirect crosses to another host. It copies every other request header,
including a custom credential header. So a 3xx from a compromised endpoint, a
misconfigured `base_url` or an upstream open redirect hands the key to the
redirect target. The affected headers are `x-api-key` (Anthropic) and
`xi-api-key` (ElevenLabs). Bearer adapters are safer, but not safe: Go compares
the hostname only, so a redirect that keeps the host and changes the port or the
path still carries `Authorization`.

The new test measures this on `main`. The redirect target receives the
credential:

```
anthropic.Generate    target hits=1 leaked headers=[X-Api-Key]
elevenlabs.Transcribe target hits=1 leaked headers=[Xi-Api-Key]
openai.Embed          target hits=1 leaked headers=[Authorization]
gemini.Generate       target hits=0 leaked headers=[]
```

**2. An unbounded successful response exhausts memory.**

A 2xx body was decoded straight from `resp.Body` with `json.NewDecoder`. The
clients send `Accept: application/json` but no `Accept-Encoding`, so Go adds
gzip and inflates the body for free. A hostile or broken endpoint can therefore
answer `200 OK` with an endless or gzip-bombed body, and the daemon buffers all
of it until it dies. Error bodies were already capped at 4096 bytes. Only the
success path was open.

## Which adapters were affected

Missing the body cap: `anthropic`, `cohere` (embed, rerank and chat),
`colbertrerank`, `whisperapi`.

Missing the redirect policy: every adapter except `gemini`. `gemini` got the
policy in #416 and is the only one that already passed the new redirect test on
`main`.

Not affected: the MCP client shims and the ElevenLabs bridge. They were fixed by
#704 and #711 and keep their own hardened path in `internal/protocol`.

## The fix: one shared path

`internal/providerhttp` holds:

- `RefuseRedirect`, the redirect policy;
- `NewClient`, `WithTimeout` and `ClientOrDefault`, the three client
  constructors, each of which installs the policy;
- `MaxJSONResponseBytes` (64 MiB) and `MaxAudioResponseBytes` (256 MiB);
- `ReadLimitedBody` and `ReadLimitedJSONBody`, the capped body reader.

Six duplicate copies of `readLimitedBody` and four copies of
`clientWithTimeout` are gone. The net change is 285 lines removed from the
adapters.

The helpers never log. An error names the byte limit and the HTTP status only.
It never holds a credential, a request payload or a response payload.

## What makes the next adapter safe by construction

A shared package alone does not stop the next author from writing a fresh
client. `tests/security/provider_http_boundary_670_test.go` parses every
provider adapter package and fails the build on three rules:

1. an `http.Client` literal must set `CheckRedirect`;
2. a response body must not go into `json.NewDecoder` or `io.ReadAll` without
   the cap;
3. an HTTP round trip must use a client that came from `providerhttp`.

The test discovers adapter packages structurally: a package under `internal/`
that imports `net/http` and uses `model.ProviderError`. There is no allowlist to
go stale, so a new adapter is covered on the day someone adds it. Each failure
message names the file, the line and the helper to call.

The rules were verified against a seeded violation. All three fire.

## Proof that the tests fail on main

The two adapter tests were copied onto `main` and run there. They fail:

```
--- FAIL: TestAdaptersRefuseARedirect
    --- FAIL: anthropic.Generate            the redirect target got 1 requests, want 0
    --- FAIL: cohere.Rerank                 the redirect target got 1 requests, want 0
    --- FAIL: cohere.Embed                  the redirect target got 1 requests, want 0
    --- FAIL: cohere.Generate               the redirect target got 1 requests, want 0
    --- FAIL: colbertrerank.Rerank          the redirect target got 1 requests, want 0
    --- FAIL: whisperapi.TranscribeStructured
    --- FAIL: elevenlabs.Transcribe
    --- FAIL: elevenlabs.Synthesize
    --- PASS: gemini.Generate               (already hardened by #416)
    --- FAIL: mistral.Embed
    --- FAIL: omniembed.Embed
    --- FAIL: openai.Embed
--- FAIL: TestAdaptersRejectAnOversizedSuccessBody
    --- FAIL: anthropic.Generate            an oversized success body must fail
    --- FAIL: cohere.Rerank
    --- FAIL: cohere.Embed
    --- FAIL: cohere.Generate
    --- FAIL: colbertrerank.Rerank
    --- FAIL: whisperapi.TranscribeStructured
```

The boundary test also fails on `main`, on the adapters named in the issue.

Every one of these passes on this branch.

## Tests

- `tests/providerhttp/adapter_redirect_670_test.go`: twelve adapter calls
  against an origin that answers 307. Each case asserts that the call reaches
  the origin, that the call fails, that the redirect target gets zero requests,
  and that no header carries the test credential.
- `tests/providerhttp/adapter_response_limit_670_test.go`: six adapter calls
  against a 200 response that is over the cap. The body is valid for the
  adapter, so a client with no cap returns success. That is why the test fails
  without the fix.
- `tests/providerhttp/shared_client_670_test.go`: the shared helpers, the two
  caps, the exact-limit boundary, and the promise that a caller's own client is
  never mutated.
- `tests/security/provider_http_boundary_670_test.go`: the three structural
  rules.
- The per-package `limit_test.go` files under `internal/` are removed. They
  tested the six copies that this PR deletes.

## Spec

`dirstral-spec/docs/` pins nothing about provider redirect handling or provider
response size. No spec change is needed, so no gated submodule re-pin is needed.

## Behavior change to note

Redirect refusal now applies to every adapter, not only `gemini`. A provider
endpoint that answers 3xx now returns a clear non-2xx provider error instead of
following the redirect. A direct provider API call never needs a redirect, and
the error names the status, so an operator can see what happened.

## Not done

- The other items of #416 stay open: `Retry-After` handling, backoff jitter and
  the plaintext-`http://` credential warning. They are separate defects and are
  out of scope here.
- The non-adapter HTTP clients (`internal/ingest` docling and recognize,
  `internal/x402`) keep their own clients. They are not provider adapters and
  the boundary test does not cover them.

Closes #670

## Summary by CodeRabbit

- **Security**
  - Provider requests now reject redirects, preventing credentials from being sent to unintended destinations.
  - Successful JSON responses are limited to 64 MiB, while audio responses are limited to 256 MiB.
  - Oversized responses are rejected safely without exposing response content in error messages.

- **Documentation**
  - Security Defaults documentation now explains response-size limits and redirect protections.


yidakra:
@coderabbitai review

Issue #664, opened 2026-08-03
ci: conformance tests exercise the legacy Handler instead of the production SDK transport


yidakra:
## Summary

The conformance suite's `newServer` helper uses:

```go
httptest.NewServer(mcp.NewServer(cfg, nil, opts...).Handler())
```

But `dir2mcp up` serves MCP-path POST/DELETE traffic through `SDKTransport`, bypassing that handler chain. The legacy handler is now materially different from production.

## Confirmed false-greens

The full `go test -race ./...` suite passes while the real binary:

- rejects an explicitly allowed cross-origin browser request in the SDK guard,
- omits CORS headers on production responses,
- accepts `tools/list` before `notifications/initialized`,
- accepts subsequent traffic without `MCP-Protocol-Version`,
- uses SDK-generated session identifiers and SDK content conversion paths the conformance suite never sees.

The same architecture gap was noted in closed #404, but the test helper remains on the dead path.

## Impact

Protocol, lifecycle, CORS, session, cancellation, and content-conversion regressions can merge under green “conformance” checks.

## Acceptance criteria

- Provide a test helper that runs `SDKTransport` on a real listener with the same options as `dir2mcp up`.
- Move wire-conformance tests to that helper.
- Keep `Server.Handler()` tests only for intentionally supported direct-handler behavior, clearly labeled as unit tests.
- Assert parity for initialize, notifications, tools/list, tools/call, x402, CORS, DELETE, and binary content.
- Fail CI if production and direct-handler tool descriptors or error contracts diverge.

Issue #663, opened 2026-08-03
mcp: SDK transport converts video clip content into an empty text item


yidakra:
## Summary

`dir2mcp_open_media_clip` emits a `toolContentItem` whose type is `video` for video sources. The production SDK adapter only maps `text` and `audio`; every other type falls through to `TextContent{Text: item.Text}`.

Video items carry bytes in `Data`, not `Text`, so the real response contains an empty text content item and drops the playable video content.

## Evidence

- Handler: video document -> `contentType = "video"`, base64 data stored in `Data`.
- `convertToolCallResult`: `case "text"`, `case "audio"`, then default empty text.
- MCP Go SDK v1.4.1 has text/image/audio/resource content types, but no native `VideoContent`.
- Direct `Server.Handler()` tests serialize the repository's custom `video` object and therefore do not expose the production conversion.

## Impact

The structured payload still carries base64 data, but the canonical `content[]` media delivery is lost for video clients; strict MCP clients receive an apparently successful call with empty content.

## Acceptance criteria

- Reconcile the canonical contract with content types supported by the pinned MCP protocol.
- Encode video using a valid interoperable content form (for example an embedded resource/blob or resource link) rather than an invented/empty type.
- Never silently map unknown binary content to empty text.
- Add production-SDK tests for both audio and video clip responses.

yidakra:
Canonical MCP-valid video representation is tracked in dirstral/dirstral-spec#60.

PR #838, opened 2026-08-12
fix(mcp): serve a configured cross-origin request on the production transport (#652)

### LINKED ISSUE #652: mcp: production SDK transport breaks configured cross-origin access
## Summary

The production `SDKTransport` used by `dir2mcp up` bypasses the repository's CORS response middleware and does not configure the MCP SDK's own `CrossOriginProtection` from `allowed_origins`.

Consequently, configured browser integrations are rejected even though direct `Server.Handler()` tests pass.

## Reproduction

Start with `--allowed-origins https://elevenlabs.io`, then initialize with:

```
Origin: https://elevenlabs.io
Sec-Fetch-Site: cross-site
```

The real binary returns:

```
HTTP/1.1 403 Forbidden
cross-origin request detected from Sec-Fetch-Site header
```

A same-host allowed-origin request reaches 200, but the actual POST response has no `Access-Control-Allow-Origin` header.

## Root cause

- `SDKTransport.Serve` sends MCP-path POST/DELETE requests directly to `serveHTTPRequest`, bypassing `Server.Handler()` and its `corsMiddleware`.
- `sdkmcp.NewStreamableHTTPHandler` is created without `StreamableHTTPOptions.CrossOriginProtection`, so SDK v1.4.1 applies its default same-origin policy instead of the configured allowlist.
- The preflight middleware advertises only `POST, OPTIONS`, although production supports DELETE session termination.
- Existing CORS tests instantiate `NewServer(...).Handler()`, not the production transport.

## Impact

Browser-based MCP clients such as the documented ElevenLabs integration cannot use an explicitly allowed cross-origin endpoint. A preflight can appear valid while the real request is denied or lacks the required response header.

## Acceptance criteria

- Configure the SDK cross-origin guard from the validated `AllowedOrigins` policy without weakening DNS-rebinding protection.
- Add CORS headers to successful/error responses on the production POST and DELETE paths.
- Advertise DELETE when session termination is supported.
- Add end-to-end tests through `SDKTransport`/a real listener for allowed and denied cross-origin POST, OPTIONS, and DELETE requests.

yidakra:
## Summary

An operator adds an origin to `allowed_origins`. A browser client still cannot
use the server. The production transport refuses the request that the operator
just allowed, and it omits the CORS header the browser needs to read the answer.

Closes #652.

## 1. What the user sees

**The browser client.** The preflight succeeds, so the client believes the
endpoint is open. The real POST then gets:

```
HTTP/1.1 403 Forbidden
Content-Type: text/plain; charset=utf-8

cross-origin request detected from Sec-Fetch-Site header
```

The client reports a connection or transport failure, not a configuration
problem. With an older browser that sends no `Sec-Fetch-Site`, the text is
different and the outcome is the same:

```
cross-origin request detected, and/or browser is out of date: Sec-Fetch-Site is missing, and Origin does not match Host
```

The documented ElevenLabs integration is exactly this shape.

**The operator.** Nothing. The origin is in the config. `dir2mcp status` is
healthy. A `curl` from the shell works, because `curl` sends no `Origin` header.
The 403 body is plain text from inside the SDK, so it does not carry the
server's canonical `FORBIDDEN_ORIGIN` code, and it does not look like this
server's own refusal. The operator has one allowlist and no signal that a second
policy overrode it.

There is a second failure hidden behind the first. Even when a request did pass,
the MCP-path POST response carried no `Access-Control-Allow-Origin` header, so
the browser discarded the body it received.

**A third, quieter change of behaviour.** The preflight advertised
`Access-Control-Allow-Methods: POST, OPTIONS`, while the transport also serves
DELETE, the MCP session-termination verb. A browser reads that list and never
sends the DELETE, so a browser client could not end its own session: the session
stayed on the server until it timed out. The list is now
`POST, DELETE, OPTIONS`, and a cross-origin DELETE from an allowed origin is
served. This is a separate user-visible fix inside the same issue, not a side
effect of the origin work.

## 2. Why the order produces this outcome

Two guards run on an MCP-path POST or DELETE:

1. `Server.allowOrigin` enforces the validated `allowed_origins` allowlist and
   answers `403 FORBIDDEN_ORIGIN` when the origin is not named (SPEC §10.5,
   §17).
2. the SDK's `http.CrossOriginProtection`, inside `StreamableHTTPHandler`.

`internal/cli/up.go` gives MCP-path POST and DELETE to `SDKTransport`.
`SDKTransport` built the SDK handler with `JSONResponse: true` and nothing else,
so `NewStreamableHTTPHandler` installed a zero-value `CrossOriginProtection`:
no trusted origins at all.

Guard 1 accepts the configured origin and returns true. Guard 2 runs next, sees
`Sec-Fetch-Site: cross-site` with an empty trusted set, and answers 403. The
later guard therefore decides every case, and the earlier, configured guard can
only ever narrow a policy it never gets to widen. `allowed_origins` became a
list with no effect on the wire.

The preflight was unaffected, which is what hid the bug. `OPTIONS` is a safe
method, so guard 2 always allows it, and `SDKTransport` hands OPTIONS back to
`Server.Handler()` where `corsMiddleware` answers 204 with the CORS headers. A
green preflight and a 403 POST is the exact signature of this bug.

## 3. The fix: make the two guards agree

The SDK guard stays on. It is given the policy it was missing.

**a. Build the SDK guard from `allowed_origins`.** `newSDKOriginGuard` registers
every allowlist entry that is a complete origin (`scheme://host[:port]`) as a
trusted origin, normalized the same way `isOriginAllowed` normalizes an origin
(lowercase scheme and host). The SDK's trusted set is therefore always a subset
of the configured policy: it can never accept an origin the allowlist refuses.
This alone fixes the documented case, `--allowed-origins https://elevenlabs.io`.

**b. Reconcile the two allowlist forms the SDK cannot express.** The SDK matches
a trusted origin by exact string. This server's matcher is looser on purpose:

* a scheme+host entry matches any port, so `http://localhost` allows
  `http://localhost:5173`. This is the shipped default, and it is how a local
  browser app reaches the server.
* a bare host entry matches any scheme and any port.

Neither form can be enumerated up front. Learning each concrete origin into the
SDK's trusted set is not an option either: that set only grows, it has no
removal, and both the port and the scheme are caller-controlled, so an allowed
host would let a caller grow the set without bound.

So for those entries `adjudicate` records the decision the allowlist already
made in the field the SDK guard reads, and only when the two guards disagree
(`Check` is called first, and a request the SDK already accepts is left exactly
as it arrived). This runs only after `allowOrigin` returned true. A refused
origin never reaches it: `allowOrigin` writes 403 and stops.

**c. CORS headers on the production POST and DELETE responses.**
`corsMiddleware` lives in `Server.Handler()`, and `SDKTransport` routes MCP-path
POST and DELETE past it. It now also wraps that branch, so a browser can read
the body it received.

**d. Advertise DELETE where it is served.** The preflight advertised
`POST, OPTIONS` while the transport serves DELETE session termination. A browser
that reads the advertised list never sends the DELETE, so the session stays alive
until it expires. The list is now `POST, DELETE, OPTIONS`, but only when a
transport that serves DELETE is in front of the handler. The handler chain on its
own answers 405 to a DELETE, so advertising it unconditionally would be the
mirror error.

**e. Own the refusal contract.** When the SDK guard still refuses a request, the
transport answers with the canonical `FORBIDDEN_ORIGIN` JSON-RPC envelope. It
calls the SDK's own `Check` to decide, so the set of refused requests does not
change; only the body does. Left to the SDK, that refusal arrives as a bare
`text/plain` line with no code, and the SDK ignores a deny handler. An opaque 403
that names nothing is what made this issue hard to diagnose in the first place.

### What I did not do

I did not disable `CrossOriginProtection`, and I did not add an insecure bypass
pattern for the MCP path. Both would have been shorter. Both would trade the
configuration bug for a security hole, because the SDK guard also refuses a case
this server's allowlist cannot judge: a request that declares itself cross-site
and sends no `Origin` header at all. The allowlist has no origin to test there
and returns true. Two tests below pin that the guard survives.

The MCPGODEBUG escape hatch (`disablecrossoriginprotection`) is likewise not
used.

I also left one adjacent bug alone, and filed it instead: #841. The SDK compares
`Content-Type` for exact equality with `application/json`, while this server
accepts a prefix, so a POST that names a charset gets a plain-text 415 from
inside the SDK. It is the same guard family, added in the same SDK release, but a
different guard with a different symptom, so it does not belong in this PR.

## 4. SPEC check

No SPEC change is needed, so nothing here is blocked on `dirstral-spec`.

* SPEC §10.5 and §17: "if `Origin` header is present, enforce allowlist",
  otherwise 403. That is what this PR restores. The allowlist becomes the
  authority again, which is what the SPEC already requires.
* The MCP `protocolVersion` target stays `2025-11-25`. Nothing in the negotiated
  version, the lifecycle or the tool contracts moves.
* The SPEC does not name CORS response headers. `corsMiddleware` and its header
  set are pre-existing repository behaviour; this PR only puts them on the path
  production actually serves. The DNS-rebinding mitigation the SDK applies to a
  loopback listener is untouched.

## 5. Proof the tests fail on main

The tests are in `tests/conformance`, so they run on the production transport
(`newServer` starts `SDKTransport` on a real listener, per #664/#827).

The preflight-only assertion is gone. `TestProduction_CORSPreflightIsServed`
passed while no browser client could use the server, so it was replaced by
`TestProduction_CORSAllowedOriginPreflightThenPostAreServed`, which drives the
whole browser sequence: preflight, then the POST it authorized, then a tool call,
and asserts `Access-Control-Allow-Origin` on the POST responses.

Same test files, product code from `main`:

```
$ git checkout main -- internal/mcp && rm internal/mcp/origin_guard.go
$ go test ./tests/conformance -run 'TestProduction_CORS|TestProduction_DisallowedOrigin|TestProduction_CrossSite|TestProduction_CrossOrigin|TestProduction_DNSRebinding' -v

--- FAIL: TestProduction_CORSAllowedOriginDeleteTerminatesTheSession
--- FAIL: TestProduction_CORSAllowedOriginPreflightThenPostAreServed
--- FAIL: TestProduction_CORSAllowedOriginWithoutSecFetchSiteIsServed
--- FAIL: TestProduction_CORSAllowlistEntryCaseIsIgnored
--- FAIL: TestProduction_CORSBareHostAllowlistEntryStaysNarrow
--- FAIL: TestProduction_CORSHeaderIsOnAnErrorResponseToo
--- FAIL: TestProduction_CORSPortlessAllowlistEntryMatchesAnyPort
--- FAIL: TestProduction_CrossOriginStillNeedsTheToken
--- FAIL: TestProduction_CrossSiteMarkerWithoutOriginIsRefused
--- PASS: TestProduction_DisallowedOriginDeleteIsRefused
--- PASS: TestProduction_DisallowedOriginIsRefused
--- PASS: TestProduction_DisallowedOriginPreflightGetsNoHeaders
--- PASS: TestProduction_DNSRebindingProtectionIsKept
FAIL
```

The messages an operator would see:

```
    cross_origin_test.go:97: preflight Access-Control-Allow-Methods="POST, OPTIONS" must contain DELETE
    cross_origin_test.go:103: initialize from https://allowed.example.com: status=403 want=200 body=cross-origin request detected from Sec-Fetch-Site header
    cross_origin_test.go:129: initialize from https://allowed.example.com: status=403 want=200 body=cross-origin request detected, and/or browser is out of date: Sec-Fetch-Site is missing, and Origin does not match Host
    cross_origin_test.go:145: initialize from http://localhost:5173: status=403 want=200 body=cross-origin request detected from Sec-Fetch-Site header
    cross_origin_test.go:411: error response Access-Control-Allow-Origin="" want="https://allowed.example.com" body={"jsonrpc":"2.0","id":65,"error":{"code":-32001,...}}
    cross_origin_test.go:443: cross-site marker without an Origin: body=cross-origin request detected from Sec-Fetch-Site header
```

Nine failures cover the allowed cross-origin POST, the POST with no marker, the
default portless localhost entry, a bare host entry, a capitalized entry, the
cross-origin DELETE, the error response with no CORS header, the non-canonical
refusal body, the token gate, and the missing DELETE advertisement. The four
refusal and guard tests pass on main, as they must.

With the fix:

```
$ go test ./tests/conformance
ok  	github.com/dirstral/dir2mcp/tests/conformance	0.34s
```

## 6. Proof a disallowed origin is still refused

`TestProduction_DisallowedOriginIsRefused` runs four cases through the
production transport. Each asserts 403, no `Access-Control-Allow-Origin` header,
and the canonical `FORBIDDEN_ORIGIN` body, so the refusal is this server's own
and not an accident of the SDK's default:

| case | origin | allowlist |
| --- | --- | --- |
| unrelated origin, cross-site marker | `https://evil.example.com` | not named |
| unrelated origin, no marker | `https://evil.example.com` | not named |
| host that only looks allowed | `http://localhost.evil.example.com` | `http://localhost` is not a suffix wildcard |
| allowed host on the wrong scheme | `http://allowed.example.com` | entry names `https` |
| disallowed origin that claims to be same-origin | `https://evil.example.com` | not named, and a caller can set any `Sec-Fetch-Site` value |

`TestProduction_CORSBareHostAllowlistEntryStaysNarrow` asserts the property the
design rests on, from the outside: a bare host entry cannot become an exact
trusted origin, so the SDK's trusted set is never wider than the configured
policy. The entry serves a port the config never named, and it refuses every
other host.

Two further tests cover the refused verbs the issue names:
`TestProduction_DisallowedOriginPreflightGetsNoHeaders` (a preflight from an
unnamed origin answers 204 and advertises nothing) and
`TestProduction_DisallowedOriginDeleteIsRefused` (a DELETE from an unnamed origin
gets 403 `FORBIDDEN_ORIGIN`, and the session it named keeps serving).

Two more tests pin the guard rather than the bug. Both pass before and after the
fix, so they exist to fail on a later change that switches the SDK guard off:

* `TestProduction_CrossSiteMarkerWithoutOriginIsRefused`: a POST with
  `Sec-Fetch-Site: cross-site` and no `Origin` still gets 403, and the test
  asserts the body is **not** the allowlist contract, so the refusal provably
  comes from the SDK guard.
* `TestProduction_DNSRebindingProtectionIsKept`: a loopback listener still
  refuses a POST whose `Host` header names another host.

## 7. Review findings addressed

A `/code-review` pass at high effort raised four findings. All four are handled in
`5b23ced`.

1. **Forging `Sec-Fetch-Site` drops the SDK CSRF guard for any allowed origin,
   and the shipped allowlist names the local host on any port (medium).** True,
   and it is what the configuration asks for: the default `allowed_origins` names
   `http://localhost` and `http://127.0.0.1`, and SPEC §16 pins those defaults, so
   narrowing them is a spec decision and not this PR. Two bounds are now written
   down in the code and pinned by a test: authorization runs BEFORE the origin
   decision, and auth is on by default, so a local page with no bearer token gets
   401 and reaches no tool; and an operator who wants less narrows
   `allowed_origins`. The exposure needs `--auth none`, an explicit opt-out of
   authentication. `TestProduction_CrossOriginStillNeedsTheToken` pins the order
   of the two gates, and fails on main.
2. **`Access-Control-Allow-Methods` widened in the shared middleware while only
   the SDK transport serves DELETE (low).** Fixed as described in 3d: the
   advertisement is now conditional, so the preflight and the real response agree
   on both paths.
3. **The charset `Content-Type` gap has no conformance case, so the suite stays
   green over it (low).** Added `TestProduction_CharsetQualifiedContentTypeIsServed`,
   committed and skipped, with #841 named in the skip reason. That follows the
   pattern PR #827 used for #663. The fix itself stays out of this PR.
4. **The SDK's refusal is a non-canonical plain-text 403, and a test pinned it
   (low).** Fixed as described in 3e, and the test now requires the canonical
   envelope instead of recording the plain-text body. It fails on main.

## 8. Checks

* `make check`: pass (fmt, vet, golangci-lint, gocyclo `-over 15`, ineffassign,
  misspell, the full suite, build). No new function needs a cyclomatic split.
* `go test ./tests/conformance ./tests/mcp`: pass.
* `coderabbit review`: run, findings addressed.

## 9. Note for reviewers

`Server.servesSessionTermination` is set by `NewSDKTransport`, once, before the
listener accepts anything, and is read only after that. It is what lets the
shared `corsMiddleware` advertise DELETE in production and not advertise it for a
bare `Server.Handler()`, which answers 405 to a DELETE.

### DIFF
--- a/internal/mcp/origin_guard.go
+++ b/internal/mcp/origin_guard.go
@@ -1,0 +1,163 @@
+package mcp
+
+import (
+	"log"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// sdkOriginGuard keeps the MCP SDK's cross-origin protection and this server's
+// `allowed_origins` policy in agreement (issue #652).
+//
+// Two guards run on an MCP-path POST or DELETE, in this order:
+//
+//  1. Server.allowOrigin, which enforces the validated `allowed_origins`
+//     allowlist and answers 403 FORBIDDEN_ORIGIN when the origin is not named
+//     (SPEC §10.5, §17).
+//  2. the SDK's http.CrossOriginProtection, inside StreamableHTTPHandler.
+//
+// Guard 2 knew nothing about the allowlist, so it refused every cross-origin
+// browser request that guard 1 had just accepted. The operator configured an
+// origin and the browser client still failed.
+//
+// The fix keeps both guards and gives them one policy:
+//
+//   - newSDKOriginGuard builds the SDK guard from the same allowlist, so every
+//     allowlist entry that is a complete origin becomes a trusted origin.
+//   - adjudicate covers the two allowlist forms the SDK's exact-string
+//     trusted-origin API cannot express.
+//
+// Disabling the SDK guard is not an option: it also refuses a cross-site request
+// that carries no Origin header at all, a case the allowlist cannot judge.
+type sdkOriginGuard struct {
+	crossOrigin *http.CrossOriginProtection
+}
+
+// secFetchSiteAdjudicated is the Sec-Fetch-Site value that makes the SDK guard
+// accept a request whose origin this server already accepted. The SDK reads this
+// header first, so it is the one field that carries the decision.
+const secFetchSiteAdjudicated = "same-origin"
+
+// newSDKOriginGuard builds the guard from the validated allowlist.
+//
+// An allowlist entry becomes a trusted origin when it is a complete origin
+// (scheme://host[:port]). An entry in any other form stays out of the trusted
+// set, and adjudicate handles it per request.
+func newSDKOriginGuard(allowedOrigins []string) *sdkOriginGuard {
+	guard := &sdkOriginGuard{crossOrigin: http.NewCrossOriginProtection()}
+	for _, allowed := range allowedOrigins {
+		trusted := trustedOriginForm(allowed)
+		if trusted == "" {
+			continue
+		}
+		if err := guard.crossOrigin.AddTrustedOrigin(trusted); err != nil {
+			// trustedOriginForm only returns a complete origin, so this branch is
+			// defensive. The allowlist check stays the authority, so a rejected
+			// form is reported and is not fatal.
+			log.Printf("warning: the MCP SDK cross-origin guard rejected allowed origin %q; the allowed_origins check stays the authority for it: %v", trusted, err)
+		}
+	}
+	return guard
+}
+
+// trustedOriginForm returns the exact Origin string to trust for an allowlist
+// entry, or "" when the entry is not a complete origin.
+//
+// The SDK guard compares the Origin header by exact string, so the entry is
+// normalized the same way isOriginAllowed normalizes an origin: a lowercase
+// scheme and a lowercase host, with the port kept as written. The result is
+// therefore always an origin the allowlist accepts as well, so the SDK's trusted
+// set can never be wider than the configured policy.
+func trustedOriginForm(allowed string) string {
+	allowed = strings.TrimSpace(allowed)
+	if allowed == "" || !strings.Contains(allowed, "://") {
+		// A bare host entry matches any scheme and any port. It is not a
+		// complete origin, so adjudicate handles it per request.
+		return ""
+	}
+	parsed, err := url.Parse(allowed)
+	if err != nil || parsed.Scheme == "" || parsed.Host == "" {
+		return ""
+	}
+	if parsed.Path != "" || parsed.RawQuery != "" || parsed.Fragment != "" {
+		return ""
+	}
+	return strings.ToLower(parsed.Scheme) + "://" + strings.ToLower(parsed.Host)
+}
+
+// protection returns the guard to hand to the SDK handler.
+func (g *sdkOriginGuard) protection() *http.CrossOriginProtection {
+	if g == nil {
+		return nil
+	}
+	return g.crossOrigin
+}
+
+// check runs the SDK's own cross-origin check and returns its error.
+//
+// The transport calls this after adjudicate, so it can answer a refusal with this
+// server's canonical contract. Left to the SDK, the same refusal arrives as a
+// bare text/plain body from inside the streamable handler, with no JSON-RPC
+// envelope and no canonical code, and the SDK ignores a deny handler. An opaque
+// 403 that names no code is what made issue #652 hard to diagnose.
+//
+// This is the SDK's own function on the SDK's own guard, so the set of refused
+// requests does not change. Only the response the client reads changes.
+func (g *sdkOriginGuard) check(req *http.Request) error {
+	if g == nil || g.crossOrigin == nil || req == nil {
+		return nil
+	}
+	return g.crossOrigin.Check(req)
+}
+
+// adjudicate records, on the request, that the `allowed_origins` allowlist has
+// already accepted this origin.
+//
+// Call it only after Server.allowOrigin returned true, and only on the path that
+// forwards the request to the SDK handler. A refused origin never reaches this
+// function: allowOrigin writes 403 and stops.
+//
+// Why it is needed. The SDK guard matches trusted origins by exact string, so it
+// cannot express the two looser allowlist forms this server supports:
+//
+//   - a scheme+host entry matches any port ("http://localhost" allows
+//     "http://localhost:5173"), which is the shipped default;
+//   - a bare host entry matches any scheme and any port.
+//
+// Learning each concrete origin into the SDK's trusted set is not safe: that set
+// only grows and has no removal, so an allowed host with a free port or scheme
+// would let a caller grow it without bound. Instead, the decision the allowlist
+// already made is written into the field the SDK guard reads, and only when the
+// two guards actually disagree.
+//
+// A request with no Origin header is left untouched. The allowlist has no origin
+// to judge there, so the SDK guard keeps full authority and still refuses a
+// request that declares itself cross-site without naming an origin.
+//
+// Consequence to keep in mind. This makes `allowed_origins` effective, so the
+// policy in the configuration is now the policy on the wire, for better and for
+// worse. The shipped default names "http://localhost" and "http://127.0.0.1", and
+// a scheme+host entry matches any port, so by default any page served from the
+// local machine is an allowed origin. Two things bound that:
+//
+//   - authorize runs BEFORE the origin decision on this path, and auth is on by
+//     default, so a page that holds no bearer token gets 401 and reaches no tool;
+//   - an operator who wants a narrower policy narrows `allowed_origins`.
+//
+// An operator who runs with `--auth none` therefore serves every local origin,
+// which is what that combination of settings asks for. The SDK's default guard
+// used to hide this by refusing the configured policy outright.
+func (g *sdkOriginGuard) adjudicate(req *http.Request) {
+	if g == nil || g.crossOrigin == nil || req == nil {
+		return
+	}
+	if strings.TrimSpace(req.Header.Get("Origin")) == "" {
+		return
+	}
+	if g.crossOrigin.Check(req) == nil {
+		// The SDK guard already agrees. Leave the request exactly as it came.
+		return
+	}
+	req.Header.Set("Sec-Fetch-Site", secFetchSiteAdjudicated)
+}
--- a/internal/mcp/server.go
+++ b/internal/mcp/server.go
@@ -118,6 +118,12 @@ type Server struct {
 
 	rateLimiter *ipRateLimiter
 
+	// servesSessionTermination reports whether a transport in front of this
+	// handler serves DELETE on the MCP path. It decides whether the CORS
+	// preflight advertises DELETE (issue #652). Written once during transport
+	// construction, before serving starts, then read only.
+	servesSessionTermination bool
+
 	x402Client      x402.FacilitatorClient
 	x402Requirement x402.Requirement
 	x402Enabled     bool
@@ -353,6 +359,31 @@ func (s *Server) Handler() http.Handler {
 	return s.corsMiddleware(mux)
 }
 
+// corsAllowedMethods returns the methods to advertise for the MCP endpoint.
+//
+// DELETE, the MCP session-termination verb, is advertised only when a transport
+// that serves it sits in front of this handler. A browser reads the advertised
+// list and never sends a method the list omits, so a browser client could not end
+// its own session and the session stayed on the server until it timed out (issue
+// #652). Advertising DELETE unconditionally would be the mirror error: the
+// handler chain on its own answers 405 to a DELETE, so the preflight would
+// promise a verb that then fails.
+func (s *Server) corsAllowedMethods() string {
+	if s.servesSessionTermination {
+		return "POST, DELETE, OPTIONS"
+	}
+	return "POST, OPTIONS"
+}
+
+// markSessionTerminationServed records that a transport in front of this handler
+// serves DELETE session termination. It is called during transport construction,
+// before the listener accepts anything.
+func (s *Server) markSessionTerminationServed() {
+	if s != nil {
+		s.servesSessionTermination = true
+	}
+}
+
 // corsMiddleware wraps the handler to support CORS preflight (OPTIONS) and
 // response headers for the MCP endpoint. Required for browser-based MCP
 // clients such as ElevenLabs Conversational AI.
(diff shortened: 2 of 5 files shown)


yidakra:
Line-reviewed, and verified against main independently.

## The defect reproduces exactly as #827 measured it

Main, with the new tests:

```
--- FAIL: TestProduction_CORSAllowedOriginPreflightThenPostAreServed
--- FAIL: TestProduction_CORSAllowedOriginWithoutSecFetchSiteIsServed
--- FAIL: TestProduction_CORSAllowedOriginDeleteTerminatesTheSession
```

with the real message an operator gets:

```
initialize from https://allowed.example.com: status=403
  cross-origin request detected, and/or browser is out of date:
  Sec-Fetch-Site is missing, and Origin does not match Host
```

That is a configured, allowed origin being refused, and the message blames the browser.

The four refusal tests PASS on main, as they must, so the fix cannot be read as loosening the policy.

## The design call I checked hardest

`adjudicate` writes `Sec-Fetch-Site: same-origin` onto the inbound request. Rewriting a browser-supplied security header deserves scrutiny, so I checked three things:

1. **Nothing else reads it.** `grep -rn "Sec-Fetch-Site" internal/ tests/` returns only this guard. So no other decision, log line or metric sees the rewritten value.
2. **It runs only after `allowOrigin` returned true**, and only when the two guards actually disagree. A refused origin never reaches it: `allowOrigin` writes 403 and stops.
3. **A request with no Origin header is left untouched**, so the SDK guard keeps full authority over a request that declares itself cross-site without naming an origin. `TestProduction_CrossSiteMarkerWithoutOriginIsRefused` passes on main and on the branch, which pins exactly that.

The rejected alternative is the right one to reject, and the reason given is the sharp one: `AddTrustedOrigin` only grows and has no removal, so learning each concrete origin would let a caller with an allowed bare host grow that set without bound.

## Two notes, neither blocking

**The trusted-set narrowing is worth keeping visible.** `trustedOriginForm` returns "" for a bare-host entry, so those entries never enter the SDK's trusted set and are handled per request instead. That means the SDK's set is always a SUBSET of the configured policy, never wider. That property is what makes the whole approach safe, and it is currently stated in a comment on a helper. If this file grows, it deserves a test asserting it directly rather than by construction.

**Preflight now advertises DELETE.** The main run shows `Access-Control-Allow-Methods="POST, OPTIONS"` failing an assertion that requires DELETE. That is a real gap being closed (the DELETE path terminates a session and the transport handles it), and it is worth naming in the PR body as a second user-visible change rather than leaving it inside the CORS fix.

No changes requested.

yidakra:
@coderabbitai full review

yidakra:
Both notes are addressed in `8081a6e`.

**1. The subset property now has its own test.** It was stated in a comment on
`trustedOriginForm` and held only by construction.
`TestProduction_CORSBareHostAllowlistEntryStaysNarrow` asserts it from the
outside, on the production transport, with a config that names a bare host only:

```go
cfg.AllowedOrigins = []string{"localhost"}
```

* `http://localhost:5173` is served, so the entry works on a port the config
  never named. That is the per-request path, because a bare host entry cannot
  become an exact trusted origin.
* `https://localhost.evil.example.com` gets 403 `FORBIDDEN_ORIGIN` with no CORS
  header, so the looser form is still not a suffix wildcard.

It fails on main for the first reason:

```
    cross_origin_test.go:164: initialize from http://localhost:5173: status=403 want=200 body=cross-origin request detected from Sec-Fetch-Site header
--- FAIL: TestProduction_CORSBareHostAllowlistEntryStaysNarrow
```

**2. The DELETE advertisement is now named as its own user-visible change.** The
PR body §1 states it separately: a browser reads `POST, OPTIONS`, never sends the
DELETE, and so a browser client cannot end its own session. The session stays on
the server until it times out. It is a fix in its own right, not a side effect of
the origin work.

`make check` is green. The conformance suite is green in CI on `8081a6e`.

yidakra:
@coderabbitai full review

yidakra:
A `/code-review` pass at high effort raised four findings. All four are handled in
`5b23ced`, and PR body §7 records each one. Summary:

1. **Forging `Sec-Fetch-Site` drops the SDK CSRF guard for every allowed origin,
   and the default allowlist names the local host on any port (medium).** Correct,
   and it is what the configuration asks for. SPEC §16 pins those defaults, so
   narrowing them is a spec decision, not this PR. The two real bounds are now
   written down where the forge happens, and one of them is pinned by a test:
   authorization runs before the origin decision and auth is on by default, so a
   local page with no bearer token gets 401 and reaches no tool. The exposure
   needs `--auth none`, an explicit opt-out of authentication.
   `TestProduction_CrossOriginStillNeedsTheToken` pins the gate order and fails on
   main.
2. **DELETE advertised in the shared middleware while only the transport serves it
   (low).** The advertisement is now conditional on a transport that serves DELETE
   being in front (`Server.servesSessionTermination`, set once in
   `NewSDKTransport`). The preflight and the real response now agree on both
   paths.
3. **No conformance case for the charset `Content-Type` gap (low).** Added
   `TestProduction_CharsetQualifiedContentTypeIsServed`, committed and skipped,
   with #841 named in the skip reason, following the pattern PR #827 used for
   \#663. The suite no longer stays silently green over it.
4. **The SDK's refusal is a non-canonical plain-text 403, and the test pinned that
   body (low).** The transport now calls the SDK's own `Check` and answers a
   refusal with the canonical `FORBIDDEN_ORIGIN` envelope. The same requests are
   refused; only the body changes. The test now requires the canonical envelope and
   fails on main.

Nine test functions now fail on main and pass here. The four refusal and guard
tests pass on both. `make check` is green, and all eight CI checks pass on
`5b23ced`.

Issue #655, opened 2026-08-03
ingest: pdf/images/audio/archives mode configuration is inert and unvalidated


yidakra:
## Summary

The canonical `ingest.pdf.mode`, `ingest.images.mode`, `ingest.audio.mode`, and `ingest.archives.mode` fields are parsed, persisted, and shown with meaningful defaults, but the ingestion runtime never reads them.

## Evidence

The corresponding `Config.IngestPDFMode`, `IngestImagesMode`, `IngestAudioMode`, and `IngestArchivesMode` identifiers occur only inside config loading/saving and config round-trip tests. No ingestion, routing, or CLI runtime path consumes them. `Config.Validate()` also does not validate their advertised enums.

For example, `ingest.archives.mode: shallow`, `off`, or an arbitrary typo all load, while archive extraction behavior remains unchanged.

## Impact

Operators cannot disable or select the advertised per-format behavior, and invalid values silently succeed. This is especially dangerous for media because provider-cost and privacy expectations can be wrong.

## Acceptance criteria

- Define and implement every advertised mode in the routing/ingestion policy, or remove/deprecate the unsupported knobs under the spec migration policy.
- Validate the closed enums at config load and after CLI overrides.
- Add behavior tests proving each mode changes processing, not merely that YAML round-trips.
- Emit a migration warning for legacy fields if another configuration surface supersedes them.

yidakra:
Fixed by merged PR #842 (stop accepting two inert config surfaces in silence); the PR named this issue in the title but carried no Closes keyword, so it stayed open. Recon before new work caught it.

Issue #654, opened 2026-08-03
retrieval: rag.k_default and rag.generate_answer are accepted but have no runtime effect


yidakra:
## Summary

Two canonical `rag` configuration knobs are loaded and persisted but never wired into either retrieval-service construction path.

## Evidence

- `Config.RAGKDefault` is only referenced in config loading/saving/tests.
- MCP parsing and `retrieval.Service.search` hardcode a fallback of 15.
- The Go `config.Default()` value is 10, while the canonical spec template says 15, so even introspection and actual behavior disagree.
- `Config.RAGGenerateAnswer` is likewise only loaded/persisted. No service setter or Ask-path branch consumes it; Ask generation is controlled solely by whether a generator exists.

## Impact

Changing `rag.k_default` or `rag.generate_answer` gives no warning and does not change requests that omit `k` or Ask behavior. Operators can believe generation is disabled when the server still sends retrieved context to a chat provider.

## Acceptance criteria

- Define the normative false behavior for `rag.generate_answer` in the spec.
- Wire one effective default-k value into MCP, local CLI, and retrieval service paths.
- Make `config.Default()`, generated config, runtime behavior, and spec agree.
- Honor `generate_answer: false` without provider calls, or reject/deprecate the knob explicitly.
- Add behavioral tests rather than config round-trip-only tests.

yidakra:
Canonical semantics and precedence are tracked in dirstral/dirstral-spec#61.

Issue #628, opened 2026-07-30
config: unrecognized keys are silently ignored, and Config.Warnings is never surfaced to the operator


yidakra:
Follow-up to #624 / #626, covering the operator-facing half of that failure mode.

## Part 1 — unrecognized config keys are silently dropped

The flat parser funnels every scalar through `setFileScalarValue` (`internal/config/config.go`), whose setter chain ends in a `default:` no-op. A key nobody claims is discarded without a word.

That is how #624 hurt: `.dir2mcp.yaml` contained `recognize_provider: serve` (written there by `dir2mcp config init`), the loader did not recognise that spelling, and recognition stayed off with no error and no warning. The operator has no way to tell a working config from an ignored one.

The same silence covers ordinary mistakes — a typo (`recognise.provider`), a stale key from an older release, wrong nesting depth, `stt.provdier`. All are accepted in complete silence.

## Part 2 — `Config.Warnings` is write-only

`Config.Warnings` (`config.go:235-239`) is documented as "non-fatal parsing messages… Callers can inspect and log these as desired". **No caller does.** Across the repo there are four append sites and zero production readers (only tests read it):

- `config.go:4201` — `DIR2MCP_SESSION_TIMEOUT is deprecated…`
- `config.go:4223` — `invalid duration for %s`
- `config.go:4273` — `invalid boolean for DIR2MCP_X402_TOOLS_CALL_ENABLED`
- `cli/app.go:1384` — `write config snapshot` failure

So those warnings are already being swallowed: an operator with a deprecated env var or an unparseable duration is told nothing. (`support_bundle.go` prints its own local `warnings` slice — unrelated.)

This matters for Part 1 too: emitting an unknown-key warning into `Config.Warnings` would be equally invisible. Both halves are needed for either to be useful.

## Proposal

**Part 1:** collect unrecognized scalar keys during parse and record one warning naming them. Detection can be done by effect rather than by enumerating every key: `fileConfig` fields are pointers, so applying a key to a fresh `fileConfig` and finding it unchanged proves nothing claimed it (the technique used by the #626 audit). Keys belonging to the separately-parsed `providers:` / `model:` / `carbon:` / `cost:` subtrees and the nested `media.translate.glossary` / `media.stt.language_providers` maps must be excluded, or valid configs would warn — false positives here would be worse than the current silence.

**Part 2:** print `cfg.Warnings` to stderr after load (`loadConfigWithGlobalOptions`), following the existing `warning: %v` convention.

[...]

PR #842, opened 2026-08-12
fix(config): stop accepting two inert config surfaces in silence (#661, #655)

### LINKED ISSUE #661: config: chunking.strategy is accepted and persisted but ignored and undocumented
## Summary

`chunking.strategy` is a recognized config key and `Config.ChunkingStrategy` is loaded, snapshotted, and saved, but no runtime path reads it. The canonical spec does not define the key or its accepted values, and validation accepts any string.

## Impact

A configuration such as `chunking.strategy: semantic` succeeds silently while chunking behavior remains unchanged. Generated/effective configuration can preserve a misleading knob indefinitely.

## Acceptance criteria

Choose and document one path:

- implement a versioned strategy enum and wire it into chunk construction/identity/reindex checks, or
- deprecate the field with a migration warning and remove it from generated snapshots/config, or
- reject it as unrecognized if it was never public.

Add a behavioral or deprecation test; a round-trip test alone is insufficient.

yidakra:
## User-visible symptom

An operator sets a config key. The server accepts it, saves it, and reports it
back in the effective config. The behavior does not change. Nothing warns.

- `chunking.strategy` was a recognized key. `chunking.strategy: semantic` loaded,
  was written back into `.dir2mcp.yaml` and into the effective snapshot, and no
  code ever read it. It was supposed to select a chunking strategy. Chunking in
  fact selects itself per document type, and the canonical spec defines no
  strategy selector at all.
- `ingest.pdf.mode`, `ingest.images.mode`, `ingest.audio.mode` and
  `ingest.archives.mode` are advertised as closed sets (`off|ocr|auto`,
  `off|ocr_auto|ocr_on`, `off|auto|on`, `off|shallow|deep`). Nothing validated
  them, so `ingest.archives.mode: shalow` loaded exactly as happily as `deep`.
  They were supposed to select per-format handling. The ingestion runtime never
  reads any of them, so no value changes anything, and a typo was accepted in
  silence. That is worst for media and archives, where the operator's belief
  about provider cost and privacy can be wrong.

## Decision per key

Two commits, one decision each, so each one reads on its own.

### 1. `chunking.strategy`: REJECTED as unrecognized (commit 1, closes #661)

The spec never mentions the key. It is absent from the SPEC §16.2 config template
and from `bs-011`, and §7.5 selects chunking by document type: characters for
text, line windows for code, time windows for a transcript. The chunker has no
strategy concept, so there is nothing to wire the value into and no specified
meaning to wire it to.

The key therefore leaves the recognized config surface. It is now an unrecognized
key like any typo: config load names it in a warning (#628) and `up` prints that
warning, and no generated config or snapshot carries it forward.

An effective snapshot written by an older release still loads unchanged. The
snapshot loader ignores the dead line without a warning, which is the right split:
a snapshot is machine-written and carries no operator intent, while a hand-edited
`.dir2mcp.yaml` does and gets the warning.

A hard startup error was rejected on purpose. #628 fixed this exact class of
problem with a warning, on the stated ground that an unknown key must not break an
existing deployment. An error would stop a running deployment over a key that
never did anything, while the warning tells the operator the same fact.

### 2. The four `ingest.*.mode` keys: VALIDATED now, behavior is spec-blocked (commit 2, refs #655)

Two halves, and they have different owners.

**Validation is ours, and it lands here.** The sets are advertised as closed, so a
value outside a set can be refused without deciding what any member means.
`Validate()` now rejects it and normalizes the accepted value. The gate lives in
`Validate()`, not in the `up` command, for the reason `validateMCPPath` gives in
#837: `Validate()` is the one gate every entry point runs, so a persisted config,
the CLI flag overlay and the save paths are all covered, and the check also runs
after the CLI overrides merge, which is what #655 asks for.

**Behavior is the spec's, and it stops here.** The canonical spec names the members
of each set and defines the behavior of none of them. Writing what `pdf.mode: auto`
or `images.mode: ocr_on` does would author normative semantics this repository does
not own; this repository is spec-first. One concrete example of the cost of
guessing: the template's default is `archives.mode: deep`, and the shipped
behavior expands only the top level of an archive, recording a nested archive as
`skip_reason=archive`. So the default value already names a behavior the
implementation does not have, and "deep" cannot be implemented as a side effect of
a validation PR. `dirstral-spec#68` is open on the adjacent question of
representation reconciliation across pipeline config changes.

So #655 stays open for the behavior half, and this PR says so instead of closing it.
Meanwhile the README stops implying that the modes work, and names the keys that
really control extraction, transcription and archive depth today.

### 3. `rag.k_default` and `rag.generate_answer`: NOT in this PR (#654 stays open)

`dirstral-spec#61` is open and owns this decision. It says in its own summary that
the ambiguity "has produced inert implementation fields and conflicting defaults in
dir2mcp#654", and it asks which surfaces `k_default` applies to, what the precedence
is among a request `k`, the per-tool schema default and the server config, and
whether `k_default` must satisfy the same 1..50 bound. Threading an effective
default `k` through `parseKArg` and the five tool schemas answers those questions,
which is spec work. The spec PR merges first, then the code PR re-pins the submodule.

For the record, the spec support for these two keys is thinner than it looks:
`k_default` and `generate_answer` each appear on exactly ONE line of the pinned
spec, inside the §16.2 "Minimal config template" block. A value in a template is
not a specified default.

## Tests

New behavioral tests, both files under `tests/`:

- `tests/config/chunking_strategy_removed_661_test.go`
- `tests/config/ingest_format_modes_validation_655_test.go`

### Proof they fail on main

Run against the unmodified `internal/config` from `main` (`d7b6354`):

```
--- FAIL: TestChunkingStrategy_FlatSpellingIsReportedUnrecognized
--- FAIL: TestChunkingStrategy_NestedSpellingIsReportedUnrecognized
--- FAIL: TestChunkingStrategy_IsNotWrittenBack
--- FAIL: TestChunkingStrategy_SnapshotDoesNotCarryIt

--- FAIL: TestIngestFormatModes_TypoIsRejected
--- FAIL: TestIngestFormatModes_EmptyKeepsTheDefault
--- FAIL: TestIngestFormatModes_ValueIsNormalized
--- FAIL: TestIngestFormatModes_LoadFailsOnATypo
--- FAIL: TestIngestFormatModes_LoadFailsOnANestedTypo
--- FAIL: TestIngestFormatModes_SaveRefusesATypo
```

Both suites pass on this branch. Two tests pass on main by design and are kept as
guards: `TestChunkingStrategy_LoadStillSucceeds` (an existing deployment carrying
the dead key must still start) and
`TestIngestFormatModes_AdvertisedValuesAreAccepted` (the false-positive guard: no
documented value may be refused).

`make check` passes locally, gocyclo `-over 15` included.

## Docs

`README.md` gains two pieces, because behavior and docs must agree:

- a chunking-keys block that names `chunking.max_tokens` / `chunking.overlap_tokens`
  and states that `chunking.strategy` is not a setting;
- a per-format mode block that gives each enum and its default, says plainly that
  no accepted value changes behavior yet, and gives the working alternative for each
  goal (`ingest.extractor: off`, `stt.provider: off`, `ingest.on_unsupported: strict`).

`docs/*.md` are pointer stubs to the `dirstral-spec` submodule, so they need no edit.

Closes #661
Refs #655
Refs #654

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,25 @@ Use it for persistent, non‑sensitive settings such as connector definitions, d
 you might want to check into source control. Values defined here may be overridden at runtime by
 environment variables.
 
+A key that no setting claims does not fail the load. `dir2mcp up` prints one warning that
+names every unrecognized key, then starts, so a typo or a stale key from an older release
+is reported instead of being dropped in silence.
+
+#### Chunking keys
+
+```yaml
+chunking:
+  max_tokens: 0        # 0 = the chunker default
+  overlap_tokens: 0    # must be smaller than max_tokens when max_tokens is set
+```
+
+`chunking.strategy` is **not** a setting. Releases before this one accepted the key, saved
+it, and never read it: chunking is selected per document type (characters for text, line
+windows for code, time windows for a transcript), and the canonical spec defines no
+strategy selector. The key is now unrecognized, so a config that still carries it loads
+with the warning above and no longer publishes the key back into the saved config or the
+effective snapshot.
+
 ### Environment variables (overrides / secrets)
 
 Sensitive keys and temporary runtime overrides are supplied via environment variables. They take
--- a/README.md
+++ b/README.md
@@ -430,6 +430,43 @@ An extractor counts as *available* only when it can actually **run**, not merely
 - *Expected docling but the banner shows `mistral-ocr (fallback ...)`* — docling either isn't on `PATH` or is present-but-broken (it failed the `docling --version` functional check, e.g. an ABI-incompatible venv). Under `auto` a non-functional docling is skipped and the cascade continues to docling-serve/Mistral; fix the install (or use the `-full` track), or pin `extractor: docling` to turn the broken install into a loud error instead of a silent fallback.
 - *Wrong host's docling chosen / pointing at a custom binary* — set `ingest.docling.command` (`DIR2MCP_DOCLING_COMMAND`) to the command template; the resolved path is redacted from diagnostics. Confirm via the `extractor` row of `dir2mcp doctor` or `routing.json` in a support bundle.
 
+#### Per-format mode keys: validated, no runtime effect yet
+
+The config template also carries one mode key per format:
+
+```yaml
+ingest:
+  pdf:
+    mode: ocr          # off|ocr|auto
+  images:
+    mode: ocr_auto     # off|ocr_auto|ocr_on
+  audio:
+    mode: auto         # off|auto|on
+  archives:
+    mode: deep         # off|shallow|deep
+```
+
+Each of the four is a closed set. A value outside its set is rejected at startup
+with `CONFIG_INVALID`, so `ingest.archives.mode: shalow` fails instead of loading as
+if it had been understood. The value is also case-normalized, and an absent key keeps
+the default above.
+
+**No accepted value changes behavior yet.** The canonical spec lists the members of
+each set without defining what any member does, so dir2mcp validates them and waits
+for that definition rather than inventing one (see `dirstral-spec`). Until then, use
+the keys that do work:
+
+| Goal | Setting that works today |
+|---|---|
+| Turn PDF/image text extraction off | `ingest.extractor: off` |
+| Choose the PDF/image engine | `ingest.extractor` (table above) |
+| Turn audio/video transcription off | `stt.provider: off` |
+| Report a format nothing can read | `ingest.on_unsupported: strict` |
+
+Archive handling today expands the top level of an archive. An archive nested inside
+an archive is not expanded; it is recorded as a skipped document with
+`skip_reason=archive`, so the gap appears in the coverage report.
+
 ### docling extraction over HTTP (docling-serve)
 
 Instead of spawning the docling CLI per document, dir2mcp can call a long-running [**docling-serve**](https://github.com/docling-project/docling-serve) HTTP container. This is the same docling engine and produces **byte-identical** output (the same structured `DoclingDocument` → Markdown + region citations) — only the transport differs (spec 0.10.0 §7.4.B).
--- a/internal/cli/support_bundle_redact.go
+++ b/internal/cli/support_bundle_redact.go
@@ -149,7 +149,6 @@ var snapshotAllowedKeys = map[string]bool{
 
 	// Chunking and ingest routing — the settings an extraction/OCR bug report
 	// is actually triaged against.
-	"chunking_strategy":       true,
 	"chunking_max_tokens":     true,
 	"chunking_overlap_tokens": true,
 	"ingest_gitignore":        true,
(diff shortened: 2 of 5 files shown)


yidakra:
@coderabbitai review

yidakra:
@coderabbitai review

yidakra:
CodeRabbit is account-level rate limited (four refusals on this PR), so I reviewed it myself. Verified against main independently.

## Verified on main

Ten tests fail, and the two guards pass, which is the split that matters:

```
--- FAIL: TestChunkingStrategy_FlatSpellingIsReportedUnrecognized
--- FAIL: TestChunkingStrategy_NestedSpellingIsReportedUnrecognized
--- FAIL: TestChunkingStrategy_IsNotWrittenBack
--- FAIL: TestChunkingStrategy_SnapshotDoesNotCarryIt
--- FAIL: TestIngestFormatModes_TypoIsRejected
--- FAIL: TestIngestFormatModes_EmptyKeepsTheDefault
--- FAIL: TestIngestFormatModes_ValueIsNormalized
--- FAIL: TestIngestFormatModes_LoadFailsOnATypo
--- FAIL: TestIngestFormatModes_LoadFailsOnANestedTypo
--- FAIL: TestIngestFormatModes_SaveRefusesATypo

--- PASS: TestChunkingStrategy_LoadStillSucceeds
--- PASS: TestIngestFormatModes_AdvertisedValuesAreAccepted
```

The second guard is the important one: no documented value may be refused. A validation PR whose real risk is over-refusal needs exactly that assertion, and it has it.

## The scope decision was the right one

`Closes #661`, `Refs #655`, `Refs #654`. Validating a closed enum needs no spec change; implementing what each member DOES is a spec decision, and stopping at that line is correct.

The `archives.mode: deep` finding is the best evidence for why: the canonical template ships `deep` as the default, and the code expands only the top level, recording a nested archive as `skip_reason=archive` with no recursion, no depth counter and no byte budget. **The default value names behavior the implementation does not have.** That cannot be fixed as a side effect of a validation PR, and it is strong input for the spec.

## One thing I checked closely: the on_unsupported refactor

`validateIngestOnUnsupported` was a hand-written switch; it now routes through `normalizeClosedEnum`. I diffed the semantics against main:

- main: lowercase, trim, empty falls back to the default, accept `lenient`/`strict`, else error naming the key and the offending value, then write the normalized value back.
- the helper: identical, including the write-back and the error shape.

So that is a behavior-preserving refactor, not a smuggled change. Worth stating explicitly because it touches a key that was already working, and a reviewer skimming the diff would see an existing validator replaced.

## Two notes, neither blocking

**The warning-versus-error choice for `chunking.strategy` is right and worth keeping visible.** #628 settled that class with a warning on the stated ground that an unknown key must not break an existing deployment. An error would stop a running deployment over a key that never did anything. The split you describe is also right: an operator-written config gets the warning, a machine-written snapshot loads silently.

**`README.md` now states which keys work today** (`ingest.extractor: off`, `stt.provider: off`, `ingest.on_unsupported: strict`) next to the ones that only validate. That is the part that keeps this honest rather than merely quieter, since a reader of the mode block would otherwise assume a validated key is a working key.

No changes requested. This PR is reviewed and I am counting it as review-complete despite the missing bot pass.

Issue #861, opened 2026-08-13
ingest: a recognizer's sources are parsed and then dropped before storage


yidakra:
`internal/ingest/recognize.go` reads a recognizer's `sources` off the wire and
then drops it before storage. The value never reaches the `spans` table, so no
client can learn WHICH recognizer produced an annotation.

## Where it is dropped

- `recognizeWireResponse` decodes `Sources []string` at `recognize.go:69`.
- `Recognize` assigns it to `model.RecognizedAnnotation.Sources` at `:108`.
- `recognitionSegments` (`:278`) builds its internal `validAnnotation` struct
  from `startMS, endMS, text, entities, event` only. `sources` stops there.

So the field is parsed, carried one function deep, and then silently discarded.
`entities` and `event` used to have the same defect. #856 fixed those.

## Why PR #862 did not fix it

PR #862 serves `entities` and `event` on a hit's span, which dirstral-spec
0.51.0 permits. It does NOT carry `sources`, on purpose:

- Every `Span` variant in the canonical `common.json` is
  `additionalProperties: false`. The `time` branch declares exactly
  `kind, start_ms, end_ms, speaker, speaker_label, entities, event, derivation`.
- `sources` is not in that list. A span that carried it would be non-conforming,
  and a strict MCP client would reject the whole tool result. That is the #397
  "Failed to call tool" class.

So the wire half is spec-blocked, not forgotten.

## What this issue needs

1. A dirstral-spec change first, per the spec-first governance loop. Decide
   where recognizer attribution belongs: a `sources` field on the `time` span,
   or a separate provenance object on the `Hit`. A span field is the cheaper
   option but it widens a closed type.
2. Then the code: carry `sources` through `recognitionSegments` into the span's
   `extra_json`, rebuild it on the store read next to `entities`/`event`, and
   serve it.

## Why it matters

`derivation` (spec 0.51.0) tells a caller whether an annotation records or
describes. `sources` tells the caller WHO said it. A corpus that mixes a
play-by-play feed with a vision model needs both to be auditable.

Issue #860, opened 2026-08-13
recognition: nothing describes what a frame SHOWS, so no fan, crowd or celebration query is answerable (scoping proposal, measured)


yidakra:
Part of #741.

The pilot user asks for "best fan moments", "show me the crowd reaction" and "when did a fan catch a ball". None of these work today. The reason is structural, not a tuning problem. All four recognizers resolve an IDENTITY or read a TEXT OVERLAY. None of them describes what a frame shows.

| recognizer | what it answers | source of truth |
|---|---|---|
| playbyplay | who pitched, who batted | MLB GUMBO feed |
| scorebug | who is on screen now | burned-in overlay |
| jersey | which number is visible | jersey crop |
| face | which rostered person | roster image bank |

A crowd shot has no feed row, no overlay, no jersey and no rostered face. It exists only in the pixels. This issue proposes a fifth recognizer that describes a frame, and it reports the measurements that make the cost and the accuracy real.

Everything below is measured on the pilot GPU host (2x NVIDIA A2, 15 GiB each) against `giants_823215_proxy.mp4` (1920x1080, 60 fps, 12236 s) and `feed_823215.json` (84 plays). Scratch work lives in `~/giants-pilot/capscope-a08a/` on that host.

---

## 1. Where it plugs in

New module `annotator/dirstral_annotator/recognizers/caption.py`, class `SceneCaptionRecognizer`, `name = "caption"`.

It follows the `faces.py` contract exactly:

* The heavy backend is an injected callable, `CaptionFn = Callable[[list[Path]], list[tuple[str, float]]]` (a batch of frame JPEGs in, a `(text, confidence)` pair per frame out). The default adapter builds a vision-language model. The package still imports with no model present, as it does today for `EmbedFn`.
* Construction raises `RecognizerUnavailable` when the backend is absent: no `torch`, no `transformers`, no CUDA device when the operator asked for CUDA, or weights that do not resolve offline. `Pipeline._shared` then records `_Unavailable` and the cascade continues. A metadata-only deployment is unaffected.
* Frames come from `base.iter_frames`, so the caption pass shares the one frame extraction the cascade already pays for.

[...]

PR #862, opened 2026-08-13
fix(mcp): show the entity and event a recognition hit matched on (#856)

### LINKED ISSUE #856: retrieval: the entities and events filters are silently ignored, and hit spans carry no attribution
Found while building the SF Giants pilot. This blocks every aggregate question the recognition capability is supposed to answer.

## Reproduction

A filter value that CANNOT match anything still returns hits:

```
dir2mcp_search {"query":"pitch","events":["no_such_event_xyz"],"k":20}
  -> 19 hits returned

dir2mcp_search {"query":"pitch","entities":["player:nobody-at-all"],"k":20}
  -> 19 hits returned
```

Verified on a real corpus of 394 recognition chunks, each with exactly one span
carrying `{"entities":[...],"event":"..."}` in the `spans` table. Reproduced with
`retrieval_hybrid_enabled: false` as well, so it is not the lexical path leaking
past the filter.

## Two separate defects

**1. A returned hit carries no attribution.** The span on a hit is only the time
range:

```json
{"end_ms": 3354398, "kind": "time", "start_ms": 3346398}
```

No `entities`, no `event`, although the `spans` row for that chunk has both. So
`model.Filter.MatchesAnnotation(hit.Span)` has nothing to match on. If the filter
DID run on these hits it would reject every one of them, because a non-empty
filter against an empty span is a non-match by design.

**2. The filter is not applied on the path in use.** Since every hit would fail
the match, an applied filter would return zero hits. It returns 19. So
`matchFilters` is not reached for these candidates. Its only call sites are
`collectFilteredHits` (`internal/retrieval/service.go:3424`) and
`internal/retrieval/hierarchical.go:215`.

Either defect alone would break the feature. Together they cancel out into "the
filter does nothing", which is why it looks like it works until you test a value
that must not match.

## Why it matters

This is the difference between a correct answer and a confident wrong one. The
pilot corpus has exactly 5 home runs, each with `event: home_run` in its span.
Asked "Who hit home runs? List every player", the server cannot use that
structure, falls back to top-k semantic search over prose, and answers from a
partial sample: it named 3 of 5 and missed a real one. Raising k to 50 did not
change the answer, and a larger chat model gave the identical wrong list, so the
evidence selection is the problem rather than the generation.

With a working filter the same question is a complete enumeration with citations.

## Design intent, already written down

The code says what it should do (`internal/retrieval/service.go:4275`):

> Optional recognition entity/event filter (dirstral-spec design 0004 §7):
> restrict to annotation hits referencing any requested entity id, and/or
> carrying any requested event. OR within each field, AND across them [...] A hit
> whose span carries no attribution never matches a non-empty filter.

So the contract is specified and the implementation does not deliver it.

## Acceptance criteria

- A hit's span carries the `entities` and `event` recorded for its chunk, so a
  client can see why a hit matched.
- `events: ["no_such_event"]` returns ZERO hits. A test must assert that,
  because it is the assertion that would have caught this.
- `events: ["home_run"]` on a corpus with 5 such spans returns exactly those 5,
  whatever the query text says.
- The filter applies on every retrieval path, vector and lexical and fused, so
  the answer does not depend on which one ran.
- Both tools that advertise the parameters are covered: `dir2mcp_search` and
  `dir2mcp_ask`.

## Related

- #849 made five tool schemas agree with their handlers. This is the same class
  one layer deeper: the parameter is parsed, threaded into the query, and then
  ignored.
- #855 added the structured outcome events that make this filter worth fixing.

yidakra:
Closes #856

PR #857 fixed the SELECTION half of #856: the entity and event filters now apply
on every retrieval path, and a value that can match nothing returns zero hits.

This PR fixes the VISIBILITY half, which was spec-blocked at the time.

## The defect

A served hit's span carried only `kind`, `start_ms` and `end_ms`:

```json
{"end_ms": 3354398, "kind": "time", "start_ms": 3346398}
```

So a caller who asked for `events: ["home_run"]` and got five hits could not
confirm that all five carry that event. The caller could not tell a filtered
result from an unfiltered one either. That is #856's first acceptance criterion,
and it is why the original defect went unnoticed for so long: no client could
look.

## What unblocked it

dirstral-spec 0.51.0 (`e5319c9`) adds three OPTIONAL fields to the `time` branch
of the published `Span`: `entities`, `event` and `derivation`. df-005 0.2.0 is
the source of truth. This PR re-pins the submodule to that commit, which is the
gated re-pin the governance loop requires.

## The change

`serializeTimeSpan` now emits `entities` and `event` when the span carries them:

```json
{"kind": "time", "start_ms": 3346398, "end_ms": 3354398,
 "entities": ["player:heliot-ramos", "team:san-francisco-giants"],
 "event": "home_run"}
```

Both fields are omitted when absent, so a plain transcript span is
byte-identical to the one served before. df-005 makes the fields additive, and a
chunk that is not a recognition annotation carries none of them.

**Every tool, not just search.** All 7 handlers that serve a Span go through one
function, `buildOpenFileSpan`: `search`, `related`, `ask`, `ask_audio`,
`transcribe_and_ask`, `open_file` and `open_media_clip`. An `ask` citation
therefore gains the fields as well. The 5 served output schemas share one
`Span` definition through `sharedDefinitions()`, so the advertised contract moves
with the payload. #849 exists because five tools disagreed with their own
schemas; one shared definition is what stops that repeating here.

The served schema MUST gain the fields. Every Span branch is
`additionalProperties: false`, so a field the server emits but does not advertise
makes a strict MCP client reject the whole tool result: the #397 "Failed to call
tool" class.

## Cost: none

The attribution is already in memory at every serialization point. `model.Span`
carries `Entities []string` and `Event string`, and the store rebuilds both from
the time span's `extra_json` on every chunk read (PR #857 pinned that
behaviour). So this is a pure serialization change. It adds no query, no per-hit
lookup, and nothing to batch.

## `derivation`: declared, never emitted

The served schema declares `derivation`, but the server never sets it. Nothing
in this tree produces one: the captioning recognizer that needs it is scoped in
#860. df-005 makes absent mean `observed`, which is what every current producer
is, so an omitted field is already correct.

Declaring it without emitting it is deliberate. A schema states what MAY appear.
Declaring it makes the served schema and the canonical `common.json` accept the
same set of payloads, so a client that validates against either one accepts a
conforming response from any implementation. Leaving it out would make the
served schema reject a conforming response from a producer that does emit it.

The normative rule df-005 attaches to it is a CONSUMER rule: a consumer MUST NOT
present a `generated` annotation as the source's own account of what happened.
Nothing here presents one, because nothing here produces one.

## `sources`: stays dropped, on purpose

`internal/ingest/recognize.go` decodes `Sources` at line 69 and assigns it at
line 108, but `recognitionSegments` (line 278) builds its internal struct from
`startMS, endMS, text, entities, event` only. So `sources` is parsed and then
silently dropped before storage.

This PR does NOT carry it through. The canonical `time` branch declares exactly
`kind, start_ms, end_ms, speaker, speaker_label, entities, event, derivation`.
`sources` is not in that list, and the branch is `additionalProperties: false`,
so a span carrying it would be non-conforming and a strict client would reject
the whole result. Carrying it to storage without a wire surface would also be
dead weight this PR cannot test end to end.

It needs a spec change first, per the spec-first governance loop. Filed as #861
with the full trace, so the knowledge is not lost.

## Tests

`tests/conformance/span_attribution_856_test.go`, on the production SDK
transport, against a real SQLite store, a real vector index and the real
retrieval service:

1. `TestSearch_FilteredHitShowsTheEventItMatched` is the assertion that matters:
   a hit returned under `events: ["home_run"]` must SHOW `event: "home_run"`.
2. `TestAsk_CitationSpanShowsTheAttribution` covers `ask` hits AND `ask`
   citations, so the fix is not search-only.
3. `TestSpan_NonRecognitionSpanCarriesNoAttribution` is the compatibility guard:
   a prose chunk's span gains none of the three fields.
4. `TestSpan_ServedSpansValidateAgainstCanonicalSchema` validates every served
   span against `definitions.Span` read from `common.json` INSIDE the pinned
   submodule. PR #852 did this for `stats` and found a real break.
5. `TestSpan_CanonicalValidatorRejectsAnUndeclaredField` is the negative
   control. It feeds the validator a span carrying `sources` and requires a
   rejection, so check 4 cannot pass vacuously. It is also the machine-readable
   reason `sources` stays off the wire.
6. `TestSpan_AdvertisedTimeBranchMatchesCanonical` pins the other direction: the
   `tools/list` time branch must declare the same properties as the canonical
   file.

### Proof they fail without the fix

With `internal/mcp/tools.go` reverted to `main` and the submodule at the new pin:

```
--- FAIL: TestSearch_FilteredHitShowsTheEventItMatched
    the dir2mcp_search hit was selected by events:["home_run"] but its span
    does not show that event, so a caller cannot tell a filtered result from an
    unfiltered one:
        {"end_ms": 3354398, "kind": "time", "start_ms": 3346398}
--- FAIL: TestAsk_CitationSpanShowsTheAttribution
    (same span, on the ask path)
--- FAIL: TestSpan_AdvertisedTimeBranchMatchesCanonical
    canonical declares "entities", "event" and "derivation", which the served
    schema does not:
    served=[end_ms kind speaker speaker_label start_ms]
```

On the FULL `main` tree (old pin plus old code) the two payload tests fail the
same way. The schema-parity test passes there only because the old pinned
`common.json` declares no such fields either, which is the drift the re-pin
closes.

## Gates

- `make check`: green.
- `go test -race ./tests/conformance ./tests/mcp ./internal/mcp ./internal/model
  ./internal/retrieval`: green.
- `gocyclo -over 15`: clean. The `time` branch moved out of `buildOpenFileSpan`
  into `serializeTimeSpan` to keep it that way.

`assertSameStrings` in `stats_canonical_schema_850_test.go` gained a `subject`
parameter so a Span drift is not reported as a stats drift. Behaviour is
unchanged.

## The red govulncheck job is not this PR

`govulncheck` reports 7 Go standard library advisories against `go1.25.12`, the
toolchain `go.mod` pins. All 7 are fixed in `go1.25.13`. The job reads the
vulnerability database fresh on every run, and the advisories landed between two
runs of this same branch: run 31749393887 passed it, run 31749641447 failed it,
and the two differ only by a comment and two test files.

So no diff here can turn it green. It needs a `go.mod` bump on `main`, filed as
#863. Every other job on this PR is green.

## Review

The CodeRabbit GitHub bot reported "Review rate limited" on this PR. The
CodeRabbit CLI ran instead and raised one minor finding: `speaker_label` is
suppressed when `speaker` is empty. That nesting is pre-existing code this PR
moved verbatim, and the case cannot occur here: the subtitle assigner returns no
label without an id, the diarizer skips a segment whose id is empty, and
`timeSpanExtraJSON` persists nothing when `speaker` is empty. Changing it would
emit a field the store cannot persist. The invariant is now recorded in a
comment so the nesting does not read as an accident.

https://claude.ai/code/session_01KMUGXpNtkZH63icDRKLCZ9

### DIFF
--- a/internal/mcp/tools.go
+++ b/internal/mcp/tools.go
@@ -3494,6 +3494,12 @@ func serializeTimeSpan(span model.Span) map[string]interface{} {
 		"start_ms": span.StartMS,
 		"end_ms":   span.EndMS,
 	}
+	// speaker_label is nested under speaker on purpose: a label always comes with
+	// a stable id in this tree. The subtitle assigner returns ("", "") for a cue
+	// with no voice name, the diarizer skips a segment whose id is empty, and
+	// timeSpanExtraJSON persists nothing at all when speaker is empty. So a
+	// label-only span cannot be produced or stored, and emitting one would
+	// advertise a speaker the caller cannot then filter on.
 	if speaker := strings.TrimSpace(span.Speaker); speaker != "" {
 		out["speaker"] = speaker
 		if label := strings.TrimSpace(span.SpeakerLabel); label != "" {
--- a/internal/mcp/tools.go
+++ b/internal/mcp/tools.go
@@ -3456,22 +3456,7 @@ func buildOpenFileSpan(span model.Span) map[string]interface{} {
 			"page": span.Page,
 		}
 	case "time":
-		out := map[string]interface{}{
-			"kind":     "time",
-			"start_ms": span.StartMS,
-			"end_ms":   span.EndMS,
-		}
-		// Additive (SPEC §8.6.8/§9.2): a diarized transcript's time span surfaces
-		// the stable speaker id and optional label. Omitted when absent, so a
-		// non-diarized span is byte-identical to before and consumers degrade to
-		// a flat citation.
-		if speaker := strings.TrimSpace(span.Speaker); speaker != "" {
-			out["speaker"] = speaker
-			if label := strings.TrimSpace(span.SpeakerLabel); label != "" {
-				out["speaker_label"] = label
-			}
-		}
-		return out
+		return serializeTimeSpan(span)
 	case "region":
 		return buildRegionSpan(span)
 	default:
@@ -3484,6 +3469,47 @@ func buildOpenFileSpan(span model.Span) map[string]interface{} {
 			"kind": "document",
 		}
 	}
+}
+
+// serializeTimeSpan renders a "time" span per df-005 0.2.0. Every optional field
+// is omitted when absent, so a plain transcript span stays byte-identical to the
+// one served before this function existed.
+//
+// speaker/speaker_label come from diarization (td-003). entities/event come from
+// the recognition capability (bs-007, design 0004 §7) and are the values the
+// `entities` and `events` filters match on. Serving them is what lets a caller
+// SEE why a hit matched: before this, a caller who asked for
+// events: ["home_run"] and received five hits could not confirm that all five
+// carry that event, nor tell a filtered result from an unfiltered one (#856).
+//
+// The values are already on the in-memory span (the store rebuilds them from the
+// span's extra_json on every chunk read), so this costs no extra query.
+//
+// `derivation` is declared by df-005 but is NOT emitted here: no producer in this
+// tree sets it yet (the captioning recognizer that needs it is issue #860), and
+// df-005 makes absent mean `observed`, which is what every current producer is.
+func serializeTimeSpan(span model.Span) map[string]interface{} {
+	out := map[string]interface{}{
+		"kind":     "time",
+		"start_ms": span.StartMS,
+		"end_ms":   span.EndMS,
+	}
+	if speaker := strings.TrimSpace(span.Speaker); speaker != "" {
+		out["speaker"] = speaker
+		if label := strings.TrimSpace(span.SpeakerLabel); label != "" {
+			out["speaker_label"] = label
+		}
+	}
+	// NormalizeEntityIDs trims, drops blanks and dedupes, which is exactly the
+	// set model.Filter matches against. So a served hit shows the matchable
+	// values, never a blank id the filter could not have selected on.
+	if entities := model.NormalizeEntityIDs(span.Entities); len(entities) > 0 {
+		out["entities"] = entities
+	}
+	if event := strings.TrimSpace(span.Event); event != "" {
+		out["event"] = event
+	}
+	return out
 }
 
 // buildRegionSpan renders a region span per spec §15.1.1: page range plus the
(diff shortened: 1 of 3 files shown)


yidakra:
@coderabbitai review

Issue #411, opened 2026-06-25
retrieval: min_score floor is compared against incommensurable score scales (RRF ~0.033 vs cosine ~1.0) → any floor silently drops all hybrid results


yidakra:
**Part of #395. Blocks the default-abstention recommendation in #403** — adding a naive default `min_score` would nuke hybrid search.

## Verified bug
`internal/retrieval/service.go` `applyMinScoreFloor` compares `h.Score < floor`, but the "authoritative Score" is on a different scale per retrieval mode:
- pure vector → cosine similarity (~0.0–1.0)
- hybrid / HyDE-fuse / cross-lingual → **RRF score** (`hybrid.go:75` overwrites `Score = entry.score`), whose theoretical max is `2/(rrfK+1) = 2/61 ≈ 0.033` (`rrfK = 60`, `hybrid.go:13`)
- rerank enabled → the reranker's own provider-specific scale

So with hybrid (the default) or HyDE or cross-lingual enabled, **every** RRF score is ≤ ~0.033, and any `retrieval.min_score ≥ ~0.04` silently drops **all** results → empty answer/search with no error. Worse, `runHybridSearch` falls back to **raw vector hits (cosine scale)** when BM25 returns zero matches (`hybrid.go:121-123`), so the *same corpus/config* yields RRF-scaled scores for some queries and cosine-scaled for others — the floor's behavior is inconsistent query-to-query. There is no score normalization before the floor.

## Symptom
An operator setting a reasonable cosine-shaped `min_score` (0.3–0.5) silently gets zero results on hybrid. And #403's proposed default-on min_score abstention cannot be implemented safely against the current raw fused scores.

## Fix
Normalize scores to a common scale (e.g. min-max per result set, or map RRF→[0,1]) before applying the floor, and document the floor's scale. Then #403's default abstention becomes feasible.

Issue #427, opened 2026-06-25
retrieval: recency decay applied after top-k truncation and inverts ranking for negative scores; slicePage phantom page; transcript regex misses minutes >=100


yidakra:
**Part of #395.** Scoring/slicing edge bugs. Findings 1-2 are opt-in (`retrieval.recency_half_life` defaults off).

## 1 [MED, opt-in] Recency decay runs AFTER top-k truncation — can't re-rank
`internal/retrieval/service.go:900-913`: `searchExpanded` returns hits already truncated to `k` (every `searchWithHyDE`/`searchByMode`/`fuseRRFMulti` truncates), and only then does `search()` call `applyRecencyDecay`. A fresh, highly-relevant doc ranked at k+1 by pure relevance can never be surfaced no matter how aggressive the half-life; worse, if decay pushes several top-k hits below `min_score`, the response returns **fewer than k** even though k+1..k+n (never seen by this stage) would survive. **Fix:** apply decay to the wider fusion/rerank pool (~50) before the final truncation.

## 2 [LOW-MED, opt-in] Recency decay inverts ranking for negative scores
`service.go:1068-1069`: `factor := math.Exp(...) (∈(0,1]) ; out[i].Score *= factor`. Multiplicative decay only *decreases* a positive score; for a **negative** score it moves toward 0 → *increases* it, so an older doc outranks a newer one. Negative cosine similarities are reachable (in-process `cosineSimilarity` hnsw_index.go:394 / diskindex.go:733, pgvector `1-(<=>)`, qdrant cosine) and aren't filtered when `min_score=0` (default). **Fix:** clamp/standardize scores to ≥0 before multiplicative decay, or decay in rank space.

## 3 [LOW] `slicePage` phantom trailing page on form-feed
`service.go:3043-3048`: `strings.Split(content, "\f")` on content ending in `\f` yields a trailing empty segment, so a 2-page doc has `len(parts)==3` and `open_file page=3` returns `""` with `ok=true` instead of an out-of-range error. (Only the raw-text fallback path; metadata spans normally drive page reads.) **Fix:** drop a trailing empty segment / bound against the true page count.

## 4 [LOW] Transcript timestamp regex misses minutes >= 100
[...]

PR #865, opened 2026-08-14
fix(retrieval): measure the min_score floor against the best hit, not the score range (#858)

### LINKED ISSUE #858: retrieval: the min_score floor always discards the lowest-scoring hit, whatever its score
Found while making the SF Giants pilot answer "list every home run" completely.

## The defect

`applyMinScoreFloor` compares each hit against a MIN-MAX NORMALIZED score:

```go
rel := normalizedRelevance(hits)   // (score - min) / (max - min)
for i, h := range hits {
    if rel[i] < floor { continue }
    out = append(out, h)
}
```

Min-max normalization maps the lowest-scoring hit to **exactly 0.0** by
construction. `0.0 < floor` is true for any floor above zero. So the
worst-scoring hit is discarded on every search that has a positive
`retrieval.min_score`, whatever its absolute score is and however good it is.

The shipped default is `0.05`, so this is on by default.

## Why it is worse as the result set gets smaller

The proportion lost grows as the candidate count falls:

| candidates | returned | lost |
|---|---|---|
| 2 | 1 | 50% |
| 3 | 2 | 33% |
| 20 | 19 | 5% |

So the defect is hardest to notice on a broad query and most damaging on a
narrow one, which is the opposite of what an operator would guess.

## Measured on a live corpus

A corpus with exactly 2 spans carrying `event: home_run`, filtered with
`events: ["home_run"]` and `k=50`:

```
retrieval_min_score: 0.05  ->  1 hit    (Bryce Eldridge)
retrieval_min_score: 0     ->  2 hits   (Bryce Eldridge, Curtis Mead)
```

Both spans satisfy the filter. Both are inside `k`. The floor removed one for
being the lowest of two, not for being weak.

The user-visible result was an `ask` answer that named one of two home run
hitters and read as complete.

## Why this is not the intended behaviour

The doc comment on the guard states the design (`internal/retrieval/evidence.go`):

> The relevance floor (config `retrieval.min_score`, applyMinScoreFloor) is a
> RELATIVE pruning control: it compares each hit against a score MIN-MAX
> NORMALIZED over the result set, which is what makes it scale-free across the
> cosine / RRF / rerank modes (#411). [...] A relative floor can prune weaker
> hits relative to the best one; it can never say the best one is itself too
> weak.

"Prune weaker hits relative to the best one" is the intent. Always deleting the
minimum is not that: it fires on a set whose members are all strong and nearly
equal, because normalization is blind to how close they are.

## Suggested fix

The rule should be "keep a hit unless it is much worse than the best one", and
the current arithmetic cannot express that at the bottom of the range. Options:

1. Compare against `max` alone, `score / max`, so a hit 1% behind the best keeps
   0.99 rather than 0.0. This preserves the scale-free property that #411 wanted
   and removes the guaranteed-drop, because nothing normalizes to zero unless its
   score really is zero.
2. Keep min-max but use `<=` semantics deliberately reversed: never prune the
   set's minimum, so the floor can only remove a hit that is above the worst.
   Simple, and it makes the guarantee explicit rather than accidental.
3. Skip the floor when the candidate count is below some N. This masks the
   problem rather than fixing it and I do not recommend it.

I prefer option 1. It changes what the floor means for every query, so it needs
a measurement pass over a real corpus before it lands, and probably a note in
the SPEC where §9.4.3 describes the floor.

## Acceptance criteria

- A search whose candidates all have near-identical scores returns ALL of them.
  A test must assert that, because it is the assertion that would have caught
  this: any test with a wide score spread passes today.
- With 2 candidates and a positive floor, both are returned when both are strong.
- The scale-free property across cosine, RRF and rerank modes is preserved, which
  is what #411 introduced the normalization for.
- The `ask` abstention path (§9.4.3, `classifyEvidence`) is unaffected: it is an
  ABSOLUTE test and must stay separate from this relative one.

## Related

- #411 introduced the min-max normalization to make the floor scale-free.
- #856 and PR #857 fixed the entity/event filter. This defect is what stops that
  filter from delivering a complete enumeration on a default configuration.
- #785 is about `search` having no absolute relevance verdict. This is the
  opposite end of the same control: the relative floor is doing something it was
  not meant to do.

yidakra:
Closes #858

## The defect

`applyMinScoreFloor` compared each hit against a MIN-MAX NORMALIZED score. Min-max maps the lowest-scoring hit of a set to exactly `0.0` by construction, so `0.0 < floor` held for every positive floor and the worst hit was deleted on every search, whatever its absolute score was and however close to the best it was. The shipped default is `retrieval.min_score: 0.05`, so this ran in every deployment.

The loss grew as the result set shrank: 5% of a 20-hit set, 33% of a 3-hit set, 50% of a 2-hit set. It therefore hid on broad queries and bit hardest on precise ones. Measured on the pilot corpus: 2 spans carried `event: home_run`, both matched the filter and both sat inside `k`; `min_score: 0.05` returned one, `min_score: 0` returned two.

## The fix, and why this option

The issue listed three options. This PR takes option 1, `score / max`.

- **Option 1 (chosen).** A ratio to the best hit says exactly what the doc comments and SPEC 9.4.3 say the floor is for: "prune hits that are much worse than the best one". A hit one percent behind the best keeps 0.99. A hit reaches 0 only when its own score is 0, so no hit is deleted for its rank.
- **Option 2 (never prune the set minimum).** Rejected. On scores `{1.0, 0.02, 0.01}` it deletes the 0.02 and keeps the 0.01, so it keeps the worst hit and drops a better one. The guarantee is stated, but the rule underneath it stays arbitrary.
- **Option 3 (skip the floor below N candidates).** Rejected, as the issue says: it masks the defect, needs an arbitrary N, and still deletes the minimum of every larger set.

## Scale-free across the modes (#411)

A ratio is invariant under positive rescaling of the pool, and rescaling is the exact difference between the modes. One configured floor keeps one meaning in each:

- **cosine** (pure vector, ~0..1): the best hit is 1.0 and the floor prunes a hit below 5% of it. A test drives 0.98/0.97/0.96/0.95 (all kept) and 0.99/0.98/0.02 (the last dropped).
- **RRF** (hybrid / HyDE-fuse / cross-lingual, max ~0.033): the best fused hit is 1.0, so a cosine-shaped floor of 0.5 still cannot wipe the set, which is the #411 regression test. RRF ranks compress (rank 10 of 61 is 0.87 of rank 1), so the floor prunes little there: an RRF score encodes rank rather than relevance, and `k` is the rank control.
- **rerank** (provider scale): tests run the same near-identical scores at 0.91/0.90/0.89 and at 91/90/89 and assert the identical outcome, then assert 91/90/1 still drops the last one.
- **index=both**: `searchBothIndices` min-max rescaled each axis before merging, so each axis's weakest hit became an exact 0.0 that the floor deleted downstream. That rescaling is now the same ratio to the axis best, with negatives clamped to 0.

## Negative scores

Checked at each source before choosing the formula. RRF sums `1/(rrfK+rank)` and is always positive. `normalizeScores` emits [0,1]. Recency decay only scales positive scores. The two paths that can go negative are the vector index (`cosineSimilarity` is a real cosine, so an anti-correlated pair is negative) and a rerank provider (`relevance_score` is passed through verbatim, and a self-hosted cross-encoder can report logits).

`score / max` is therefore guarded:

- `max <= 0` (all-zero pool, fully anti-correlated pool) reports "no ratio defined" and the floor **fails open**, keeping the pool. A negative max would inverse the ordering and a zero max would divide by zero.
- `max == +Inf` reports the same, since every finite score would degenerate to 0.
- a NaN never wins the max comparison, so an all-NaN pool fails open too. A NaN **beside readable scores** is the opposite case and it drops: it carries no relevance reading of its own, and a plain `rel < floor` test would have admitted it, because every comparison against NaN is false. `normalizeScores` clamps the same value to 0, so no NaN reaches `SearchHit.Score` on the index=both path.
- a negative score **beside a positive best** keeps a negative ratio and still drops, which judges that hit's own score and not its rank.

## The abstention path is unchanged

SPEC 9.4.3 requires the relative pruning floor and the absolute evidence threshold to stay separate. `classifyEvidence`, `evidenceThresholds` and `abstainOnWeakEvidence` are untouched. They read `EvidenceScore` / `EvidenceScale`, which nothing here writes. A new test drives two near-identical candidates that both clear the relative floor and are both below the absolute cosine threshold: `ask` still abstains with empty citations and both candidates still show up in `hits`.

MMR keeps min-max (`normalizedRelevance`): its objective needs the full [0,1] spread to trade relevance against diversity, and it re-orders candidates rather than deleting them.

## Spec

SPEC 9.4.3 does not pin the normalization. It says the pruning floor **MAY** be relative ("normalized per result set") and **MUST** apply across one comparable scale. A ratio to the best hit is a per-result-set relative normalization onto [0,1], so the submodule pin is unchanged and no spec PR is needed. 16.2 leaves the numeric default to the server; `0.05` still ships, now meaning "drop a hit below 5% of the best hit".

## Tests

New `tests/retrieval/min_score_floor_relative_858_test.go` plus extra cases in the existing internal floor test. Against `main`'s `internal/retrieval/service.go`, 7 of the 9 new tests fail:

```
--- FAIL: TestSearch_MinScoreFloor_858_NearIdenticalScoresAllSurvive   got [1 2 3], want 4
--- FAIL: TestSearch_MinScoreFloor_858_TwoStrongCandidatesBothSurvive  got [1], want 2
--- FAIL: TestSearch_MinScoreFloor_858_BothIndexNearIdentical          got 2 hits, want 3
--- FAIL: TestSearch_MinScoreFloor_858_RerankScaleNearIdentical        got 2 hits, want 3
--- FAIL: TestSearch_MinScoreFloor_858_RerankScaleIsScaleFree          got 2 hits, want 3
--- FAIL: TestSearch_MinScoreFloor_858_FilteredHybrid                  got [1], want [1 4]
--- FAIL: TestAsk_MinScoreFloor_858_AbstentionStaysAbsolute            got 1 hit, want 2
--- FAIL: TestApplyMinScoreFloor_RRFScaleNotWipedOut                   got [1 2], want 3
--- FAIL: TestApplyMinScoreFloor_NoPositiveBestIsPassThrough           all negative: got [1], want 2
```

The two that pass on both are the guards: the floor must still drop a hit at 2% of the best, and an anti-correlated hit must still drop.

One existing assertion changed, and it is the one that pinned the defect: `TestApplyMinScoreFloor_RRFScaleNotWipedOut` asserted "the weakest normalized hit must drop above floor 0". It now asserts that near-identical RRF hits all survive the default floor, and that a floor of 0.99 still keeps only the best, so the floor is not a no-op on that scale.

## Gates

`make check` passes, plus `go test -race ./tests/retrieval/...`. `gocyclo -over 15` is clean.

Docs updated in the same PR: README relevance-floor section, the `retrieval.min_score` config comments and validation message, `evidence.go`, `related.go` and `model/types.go`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved minimum-score filtering by comparing results to the best match, reducing the risk of incorrectly removing similarly relevant results.
  - Weak or negatively scored results are filtered more consistently.
  - Preserved absolute-threshold behavior for cases where no sufficiently relevant result exists.
  - Improved score handling across vector, hybrid, dual-index, and reranked searches.

- **Documentation**
  - Updated relevance-score and filtering documentation to explain the new relative scoring behavior.


yidakra:
CodeRabbit's NaN finding is addressed in 0752632d, both places it named.

- `applyMinScoreFloor` now drops a NaN ratio explicitly (`math.IsNaN(rel[i]) || rel[i] < floor`). A plain `rel < floor` admitted it, because every comparison against NaN is false.
- `normalizeScores` clamps the same value to 0 (`!(rel[i] > 0)` catches NaN and negatives alike), so no NaN reaches `SearchHit.Score` on the index=both path.
- The set-level pass-through keeps its fail-open rule, and the code says why the two calls differ: an all-NaN set has no readable member, so pruning would be arbitrary, while a single NaN beside readable hits carries no relevance reading of its own.
- Regression test: `TestApplyMinScoreFloor_NaNScoreDropsBelowPositiveBest` covers the mixed finite/NaN pool and the all-NaN pool.

The README pass-through sentence now lists the `NaN` and `+Inf` cases too.

yidakra:
Verified on the live pilot corpus, not only in unit tests. I ran the same filtered queries against both binaries with `retrieval.min_score = 0.05`, changing nothing else.

| filtered query | main | this branch |
|---|---|---|
| `events=["batted_ball"]`, k=50 | 7 hits, lowest score `0.014706` deleted | 8 hits |
| `events=["captivating"]`, k=50 | 6 hits, lowest score `0.030077` deleted | 7 hits |

The `captivating` case is the clearest evidence. On main the deleted hit scored `0.030077` while the kept one above it scored `0.03031`. The two are 0.8% apart. The hit was deleted for being the lowest of the set, not for being weak, which is exactly what #858 describes.

Three findings from that run that change how the issue should be read.

**1. The floor runs before the final truncation to `k`.** This is deliberate (#427), so a survivor ranked at `k+1..k+n` can surface. It means the deleted hit is invisible whenever the candidate pool is larger than `k`, because truncation would have cut it anyway. Every unfiltered query I probed returned a full 20 of 20 on both binaries for that reason. The defect becomes user-visible only when the surviving pool is at or below `k`, which is the narrow or filtered query. The README said the floor is applied "last, after ... dedup/truncation", which reads as the opposite. I corrected that wording in 8fdfc110.

**2. An exact tie masks the defect.** The `events=["home_run"]` query in the issue now returns 2 of 2 on main, because after the re-recognition in #855 and #859 both home runs score exactly `0.032522`. A tie is the degenerate case that min-max maps to all-1, so both survive. The original 1 of 2 measurement was taken before that re-index. The defect is unchanged; that one repro no longer shows it.

**3. The pilot itself was not affected in production.** Its config carries `retrieval_min_score: 0`, so the floor is off there. The measurement in the issue was taken with the floor switched on by hand. The defect is real for any deployment on the shipped default, which the pilot is not.

The pilot was restored to its original binary and config after the run, and its browser suite passes.

Issue #92, opened 2026-03-01
Missing SPEC-defined exit codes (0-6)


Ali-Bilge:
## Problem

SPEC §2.4 defines specific exit codes with distinct meanings, but they may not be properly implemented or used consistently.

## Expected exit codes

| Code | Meaning |
|------|---------|
| 0 | Success |
| 1 | General error |
| 2 | Config/validation error |
| 3 | Ingestion error |
| 4 | Server startup error |
| 5 | Auth/payment error |
| 6 | Signal/interrupt |

## What needs to happen

- Define named constants for each exit code (e.g. in `internal/cli/` or a shared package)
- Audit all `os.Exit()` calls and error returns to ensure correct codes are used
- Ensure signal handlers return code 6

## Affected file

`internal/cli/app.go`

PR #866, opened 2026-08-14
fix(ingest): carry a recognizer's sources through to the served span (#861)

### LINKED ISSUE #861: ingest: a recognizer's sources are parsed and then dropped before storage

yidakra:
## The defect

`internal/ingest/recognize.go` read a recognizer's `sources` off the wire, copied
it onto `model.RecognizedAnnotation`, and then dropped it. `recognitionSegments`
rebuilt each annotation as `{startMS, endMS, text, entities, event}`, so the tags
stopped one function short of the span. A client could see WHAT an annotation was
about (`entities`, `event`) but never WHO said it: a scorebug reading and a face
match looked the same.

`entities` and `event` had the same defect. #856 fixed those. #862 left `sources`
alone on purpose, because the canonical `time` branch is
`additionalProperties: false` and did not declare it.

## The spec is in now

dirstral-spec PR #92 merged as 0.52.0 (df-005 0.3.0). It adds the optional
`sources` array to the `time` span. This PR re-pins the submodule to `5beca83`,
the merge commit on the spec's default branch.

## What changed

The tags now take the same path `entities` and `event` take:

- `internal/model/types.go`: `Span.Sources`, plus `NormalizeSources` (trim, drop
  blanks, de-duplicate, keep the producer's order).
- `internal/ingest/recognize.go`: `recognitionSegments` carries the tags onto the
  span instead of dropping them.
- `internal/store/sqlite_store.go`: `timeSpanExtraJSON` persists `sources` in the
  span's extra_json; `annotationFromExtraJSON` rebuilds it on every chunk read.
- `internal/mcp/tools.go`: `serializeTimeSpan` serves it, and the advertised time
  branch declares it.

Every handler that serves a Span goes through `serializeTimeSpan`, so `search`,
`ask`, `related`, `open_file` and `open_media_clip` gain the field together. The
five served output schemas share one Span definition, so they gain it together
too. The declaration is mandatory: every Span branch is
`additionalProperties: false`, so a field the server emits but does not advertise
makes a strict client reject the whole tool result (the #397 class).

## Provenance only

df-005 0.3.0 is explicit: an implementation MUST NOT require a client to read
`sources` to rank or filter correctly. Nothing here scores, filters, dedups or
selects citations on the field. `TestSources_AreProvenanceOnly` pins that: two
corpora that differ ONLY in the tags run the same four queries, and the served
payloads must be byte-identical once the tags are removed. A ranking rule that
started to read them fails the test even if the hit count stayed the same.

## Optional means absent

An annotation that names no recognizer omits the key entirely: not `null`, not
`[]`. `TestSpan_NoSourcesOmitsTheFieldEntirely` checks the decoded payload, which
tells all three shapes apart.

## The derivation hash

The hash covers what is STORED, and the tags are now stored, so they join the
hash input: a backend that re-attributes an annotation from the scorebug to the
face matcher must re-derive rather than keep the old provenance because the prose
matches. The tag group is APPENDED ONLY WHEN PRESENT. The entity list ends at a
known count, so both readings stay unambiguous, and an annotation with no tags
keeps the exact hash input it had before. That matters: recognition is the most
expensive derivation in the tree, and a blanket change would re-run the backend
over every video in every corpus for a field those annotations do not have.
`TestASourcelessAnnotationKeepsItsOldHashInput` pins the old string.

## Tests

New, and each fails without the matching half of the fix:

- `tests/ingest/recognize_sources_861_test.go`: the tags reach the span in order;
  a change re-derives; the length-prefixed encoding cannot collide; a blank or
  repeated tag is not a difference; a sourceless annotation keeps its old hash.
- `tests/store/annotation_sources_861_test.go`: the round trip through
  extra_json, including a span whose only optional field is `sources`.
- `tests/conformance/span_sources_861_test.go`: the served payload on the
  production SDK transport, validated against the canonical `common.json` inside
  the pinned submodule, plus the omission rule and the provenance-only rule.

`tests/conformance/span_attribution_856_test.go` changes in two places. Its
non-recognition guard now also requires `sources` to be absent. Its negative
control moves off `sources`, which the canonical schema now declares, onto
`confidence`, which the recognize wire carries and the span branch still does not
declare, so the control keeps rejecting a real mistake.

Closes #861

https://claude.ai/code/session_01KMUGXpNtkZH63icDRKLCZ9

## Summary by CodeRabbit

- **New Features**
  - Added optional recognizer provenance to time-span annotations through a normalized `sources` array.
  - Provenance is now preserved in stored annotations and surfaced in search results and ask citations.
  - Source details are omitted when unavailable and do not affect ranking, filtering, scores, ordering, or citation behavior.

- **Bug Fixes**
  - Changes to recognition sources now correctly trigger annotation re-derivation.
  - Improved handling of duplicate, blank, and malformed source values.

- **Tests**
  - Added coverage for ingestion, persistence, schema validation, search, and citation output.

### DIFF
--- a/internal/ingest/recognize.go
+++ b/internal/ingest/recognize.go
@@ -282,6 +282,7 @@ func recognitionSegments(anns []model.RecognizedAnnotation) ([]chunkSegment, str
 		text     string
 		entities []string
 		event    string
+		sources  []string
 	}
 	valid := make([]validAnnotation, 0, len(anns))
 	for _, ann := range anns {
@@ -296,6 +297,13 @@ func recognitionSegments(anns []model.RecognizedAnnotation) ([]chunkSegment, str
 			// persisting only the text made the filter unimplementable.
 			entities: model.NormalizeEntityIDs(ann.Entities),
 			event:    strings.TrimSpace(ann.Event),
+			// Carried for the same reason, but for a different question
+			// (df-005 0.3.0 `sources`): entities and event say WHAT the
+			// annotation is about, sources says WHICH recognizer said it. The
+			// backend has always sent it and this function used to drop it, so
+			// no client could tell a scorebug reading from a face match (#861).
+			// Provenance only: nothing downstream ranks or filters on it.
+			sources: model.NormalizeSources(ann.Sources),
 		})
 	}
 	sort.Slice(valid, func(i, j int) bool {
@@ -314,7 +322,7 @@ func recognitionSegments(anns []model.RecognizedAnnotation) ([]chunkSegment, str
 			Text: v.text,
 			Span: model.Span{
 				Kind: "time", StartMS: v.startMS, EndMS: v.endMS,
-				Entities: v.entities, Event: v.event,
+				Entities: v.entities, Event: v.event, Sources: v.sources,
 			},
 		})
 		// The attribution joins the derivation hash: a backend that changes
@@ -332,6 +340,26 @@ func recognitionSegments(anns []model.RecognizedAnnotation) ([]chunkSegment, str
 			v.startMS, v.endMS, len(v.text), v.text, len(v.event), v.event, len(v.entities))
 		for _, id := range v.entities {
 			fmt.Fprintf(&hashInput, "|%d:%s", len(id), id)
+		}
+		// The recognizer tags join the hash too, and for the same reason: a
+		// backend that re-attributes an annotation from the scorebug to the
+		// play-by-play feed has stored different provenance, so the
+		// representation must be re-derived rather than kept because the prose
+		// matches. Same length-prefixed encoding, because a tag is an opaque
+		// backend token that may contain any delimiter.
+		//
+		// The whole group is APPENDED ONLY WHEN PRESENT, unlike the entity
+		// count above. The entity list ends at a known count, so a line either
+		// stops there or continues with the source count: both readings are
+		// unambiguous. Emitting "|0" instead would change the hash input of
+		// every annotation that carries no sources, and that re-runs the
+		// recognition backend over a whole corpus of video to store a field
+		// those annotations do not have.
+		if len(v.sources) > 0 {
+			fmt.Fprintf(&hashInput, "|%d", len(v.sources))
+			for _, source := range v.sources {
+				fmt.Fprintf(&hashInput, "|%d:%s", len(source), source)
+			}
 		}
 		hashInput.WriteByte('\n')
 	}
--- a/internal/ingest/recognize.go
+++ b/internal/ingest/recognize.go
@@ -271,10 +271,12 @@ func RecognitionSegments(anns []model.RecognizedAnnotation) ([]ChunkSegment, str
 // core does not rely on the draft schema alone): empty or whitespace-only text
 // (a non-searchable chunk), a negative start (the wire contract is 0-based, §5),
 // or a reversed span (end before start). The survivors are sorted by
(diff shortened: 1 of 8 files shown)


yidakra:
On the stdio-coverage finding in `tests/conformance/span_sources_861_test.go`: I checked it against the code and the premise does not hold, so I am not making the change. Here is the evidence.

**The test does not use `newDirectHandlerServer`.** It builds its server with `newServerWithRetriever` (line 130). That helper starts the production transport:

```go
transport := mcp.NewSDKTransport(server, ln, "", "")
go func() { srv.done <- transport.Serve(ctx, server.Handler()) }()
```

It listens on a real TCP socket and wires the server the way `internal/cli/up.go` wires it. So these assertions already read the contract a real client gets, which is what the finding asks for. `newDirectHandlerServer` appears nowhere in the file.

**The conformance suite carries no stdio transport, by design.** The package doc states the suite drives the server over HTTP and asserts only on HTTP responses. A grep for stdio across `tests/conformance/` returns nothing. There is no "existing stdio test-server/client helper" to use.

**stdio is covered, in the right place.** dir2mcp speaks MCP over Streamable HTTP. The stdio path a desktop client uses is the separate `mcp-remote` bridge, which is a process boundary rather than a Go call. It is exercised by the release gate `make release-smoke TRANSPORT=stdio`, which drives the same `bunx mcp-remote` bridge Claude Desktop uses. Recreating that inside a Go test would mean spawning the bridge from the test process, which is why the gate exists as a separate manual step.

So the coverage the finding wants exists, and one of the two transports it names is already what this test runs. Resolving.

